PR #1, opened 2020-11-18
[WIP] design specification

scarrazza:
Here we can discuss about the design specification for the apfelcomb replacement.
I am including a very basic setup description, with particular emphasis on what I dislike in apfelcomb


scarlehoff:
Why not, [Pineco](https://wiki.pokemoncentral.it/Pineco#/media/File:Artwork204.png)?

design/README.md
```diff
@@ -0,0 +1,73 @@
...
+The software suite should provide tools for the preparation of:
+- EKO evolution kernels for NNPDF theories.
+- PineAPPL grids predictions for all datasets in NNPDF.
+
+Both objects must be stored in a NNPDF repository (server).
+
+### Procedure
+
+Starting from the previous two input values the algorithm should:
+- download the precomputed EKO evolution kernel for the specific theory identification number from the NNPDF repositories. Preferably store the kernels in a format which does not require EKO at all.
```
scarlehoff, line 68: Regarding this, I'd propose the following programs/scripts:

- (1) Eko generate: generates an eko runcard file.
- (2, depends on 1): Eko download: we compute the md5sum or whatever other algorithm on the runcard file and check for a corresponding file in the server, if it doesn't exit.
- (3, depends on 1, fires if 2 fails): Eko create: creates the operator
- (4, depends on 3): Eko upload: uploads the operator (takes the md5sum of the runcard).
- (5, depends on 1-2): Eko operate: just operator*input itself
- (6, depends on 5, [others]): Pineapplekocomb: produces the fktable  

I say md5 on the runcard but we can include the eko release tag or whatever in addition. Take md5 just as the first thing that came to mind.
Doing something like this is robust in that any little change (in the definition of the runcard or the code) would force a regeneration of all operators. It might be a bit annoying at the beginning but I believe that after release it should stabilize (for instance, apfelcomb would should point to a specific eko release and not to the head of the repo.

The reason why I think it is worthwhile to do this using the validphys tech are:
- It offers already the communication with the NNPDF server (parts 2, 4)
- It is a graph generator that manages the dependencies and that the people (at least the students) in NNPDF knows how to use*. I think the goal here should be to produce something that others can manage afterwards.
- I will be able to use (5) in the PDF fits and bring with it the rest of the technology.
+ I think having a validphys evolve action will open the door to new studies that people are not doing right now just because running Apfel is such a pain*, being able to use (5) automatically without having to thing about any of the other steps would be great.

*this is not necessarily a good thing, but some of the people _only know* validphys and struggle even with more general python let alone Fortran but ¯\_(ツ)_/¯

scarrazza:
> Why not, Pineko?

oh god, that's pretty good, in particular from its biology we see:

> It makes its shell thicker by gluing tree bark to its body. The glue is made of fluid that Pineco spits out, which hardens when it is exposed to air. Unbothered by the added weight of its shell, it patiently hangs from a tree branch in tropical and temperate forests, waiting to prey on flying insects. 

which is *exactly* what the code is supposed to do ;-P.

design/README.md
cschwan: That seems a good idea to me. Steps 1-5 seem like they are some kind high-level EKO 'helpers' API. Step 5 itself should just be linear algebra and quite fast. Given that we have steps 1-5 step 6 is probably almost trivial.

design/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Problems to be avoided
+
+Here a short summary of practical problems we would like to avoid:
+- avoid different data input format, today we have to maintain parsers for APFEL (DIS/FTDY), APPLgrid and PineAPPL,
+- avoid long runtime operations that can be cached and stored in a server a priori, e.g.: DGLAP evolution kernels, DIS and FTDY grids,
+- reduce to minimum data input processing,
+- delegate the FK Table construction to a single code/function,
+- provide tools to cache results and upload to server
+- provide tools to optimize the quality of FK tables.
```
cschwan, line 45: Here are more problems that we can avoid:

- after generation we want to check three numbers with a realistic PDF (a toy PDF is not enough!): 1) the PineAPPL predictions, 2) the predictions using EKO with a reduced number of grid points (measures the compression penalty), and 3) the prediction of the FK table (measures the effect of the down-evolution). The last two numbers are the ones shown already in apfelcomb, but we really need the first number as well, as seen with ATLAS Z high-mass DY
- personally, I don't like the format of apfelcomb's database at all: 1) why is there an ID for each grid? I'd rather invoke the program with the grid file and the theory identifier, 2) the ppbar switch should no longer be needed as this something that I'll implement in PineAPPL (see https://github.com/N3PDF/pineappl/issues/39), 3) the bin selection is very fragile - right now there's a '1' for every bin that's included (imagine doing that for ATLAS 3D DY with 504 bins), but we probably should do it the other around, by specifying cuts, 4) merge all the databases (DIS/FTDY/...) into a single one

AleCandido:
I support the Pineko proposal, we'll have also a logo ready designed (even if it will be actually copyright protected...)

design/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+## APFELcomb design and issues
+
+### Workflow
+
+The apfelcomb project compiles a program which connects several tools and
+databases from NNPDF and external providers in order to generate FK Tables,
+where observables are stored into tables with a predefined input scale energy value for PDFs evaluation.
+
+The user/develop should include the custom implementation of the combination operation for each specific process category. All operations involving DGLAP, DIS and Fixed-target DY are performed at runtime. Operations involving the generation of hadronic observables with or without electroweak corrections are performed off-line and stored on git-lfs repositories.
```
AleCandido, line 15: Sorry for nitpicking, but since this is a specs document: are you sure that you would like to rely on using `git-lfs` always?
(of course the problem it's not the document itself, but the code written after it)

design/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Proposal
+
+We propose to develop a code which performs the combination and provide tools to prepare, store and update its requirements.
+
+### Input
+
+The program should take as input:
+1. the theory number identification number, using the NNPDF theory database. Note that in opposite to apfelcomb we should take the number and not the full sqlite string.
+2. the dataset name following the NNPDF convention.
```
AleCandido, line 55: Here we need to decide if `eko` and `yadism` additional configs will be provided as:
- an additional database
- an additional configuration file
- connected to the respective datasets
- hardcoded in `Pineko` once for all

design/README.md
AleCandido: I completely agree with the proposed 1-6 steps, in particular I like the part of the `md5sum`, it should be a part of this document and we should optimize it step by step (of course I'm referring to the concept with the same caveats Juan wrote).
I also agree with Christopher:
> Step 5 itself should just be linear algebra and quite fast.

and step 6 will just consist of dumping the generated FKtable with the suitable metadata.

#### Validphys
On this I'm not sure. I had a look into validphys code some time ago, but it was really doing a lot of operations with a lot of hardly maintainable code (lucky us that Zahari is still around), while the `Pineko` business it's quite easy.

My personal opinion would be not to use validphys for things that are not strictly coupled to the fit, i.e. are happening offline. At the end of the day `Pineko` will expose its interface, so you will use that one whatever it is, and CLI/python lib it's in any case better than validphys interface.

The only thing I'm completely ignorant about it's how much it's difficult to interact with NNPDF server. Can't we just use something like [`requests`](https://requests.readthedocs.io/en/master/) and manage everything with http's `GET`/`POST`?

#### Future

> I think the goal here should be to produce something that others can manage afterwards.

This is always 100% true.

scarlehoff: wrt validphys, in this case is very simple, we want it for the graph dependencies + availability for the rest of NNPDF. We don’t need to create reports so that facilitates a lot what we have to do.

wrt network, there is something else to discuss. Should it really go to the NNPDF server and not to apfel (the computer, not the program)? If going to the NNPDF server I think we should all use the same API (specially if it is to be made public)

design/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Precomputed objects
+
+The software suite should provide tools for the preparation of:
+- EKO evolution kernels for NNPDF theories.
+- PineAPPL grids predictions for all datasets in NNPDF.
+
+Both objects must be stored in a NNPDF repository (server).
```
scarlehoff, line 63: About this, because it wasn’t clear to me in our discussions (or in the Friday meetings). When going public, does that include the grids and the eko operators or only the tools?

scarrazza: I think that server or repository, or say both are acceptable solutions. Personally I would prefer to have both.

design/README.md
AleCandido: First of all I'm not against `valiidphys` a priori, I'm just worried of *extra dependencies*. If it's not *extra* it's fine.

Then I see 3 reasons at the moment for which I could exploit a graph:
1. inlining the code
2. dispatching the tasks
3. managing asynchronous stuffs

The first one it's definitely not the case, also the second one I don't think we need, and for the third there is no benefit in using external packages since python has native syntax for it, in particular you can read in [`validphys`](https://github.com/NNPDF/nnpdf/blob/8cd331ffb718f44766e1f220b46bced3dabadc38/validphys2/src/validphys/loader.py#L627-L628)
```python
#TODO: Make this async someday
class RemoteLoader(LoaderBase):
```

The "task dependencies" I can't see how it is useful: in which way is it different from normal branching with `if... else` statements?
(sorry if I look aggressive, I'm just discussing and I would like to understand; I really believe you are more expert than me on this, but I simply decided to exhibit consistently my point of view)

scarlehoff: I think this example is a good motivation https://docs.nnpdf.science/vp/api.html#generic-example

The point of the graph is to create a number of providers that others can use. For instance, once the (6) points mentioned before exist I can create a new function in validphys;

```python
def do_something_funny_with_operators(pdf_at_scale, eko_operator):
    return pdf*eko_operator
```
Where `pdf_at_scale` is some other action that someone did that take some PDF and evolves with eko, which also uses some of the stuff we do here. 

without doing anything else I already have access to the whole machinery that we did here. Validphys will also internally decide what is the correct dependency tree (we have two dependencies with eko in those arguments!). As I've already mentioned, it is not just theoretical, there is already an use case which is using it at the end of the fits.

And more generally, I think that if we are doing something for NNPDF it should be made using the NNPDF customs and not create our own.

pd: It pains me to be the one advocating for validphys but is precisely in this context where it is useful x)

AleCandido: As far as I understand the use case you proposed it is simply *caching the output* locally (i.e. `eko_operator` is the (6), in the sense that represent it's output and it's able to run (6) if needed). Sorry if I'm dense: I'm really not familiar with `validphys` and it's not a straightforward object.

But if this is the problem we can in no time implement the local cache in `Pineko`, since it is just costing us to add a `cache_outputs = {}` somewhere, and retrieve stuffs according to the `md5sum` we are already planning.

> Also, I don't think it is an extra dependence when it is an internal tool.

Even `eko` is a dependency for `yadism`, not an *extra* one of course, but we paid so many time the price for this in many different ways that I became a bit cautious (and consider that the 2 projects are maintained by the same 2 developers).

scarlehoff: Rereading your comments I think the point you are missing is that I am not talking of `Pineko`. Of course it adds nothing for us _here_ but it adds a lot for NNPDF as a whole.

As I said, there is already an use case: those functions will be used in the fits.

felixhekhorn: - I'm of course on the same grounds as Alessandro: I never worked nor used validphys myself, but I can see that the thing is thousands of line long and has tons of modules
- that beeing said, I shared his doubts about using validphys here
- so here is someting I don't understand: do you want to make validphys a dependency of pineko or no? will there be validphys decorators?

design/README.md
scarrazza: Yes, but note that this program/step is independent of yadism.
We need another section in this docs for the specifications of the "theory generators" in particular (yadism, mg5).
We may follow the development pattern that @cschwan is using for mg5, and extend it if needed.

design/README.md
scarrazza: I think we can/should have both: git-lfs and server. In particular the server could render the content of the git-lfs repos for fktables, evolution kernels and pineappl grids.

The server is good because:
- easy & fast access
- public
- already implemented with backups

The repo is good because:
- we can check the revision if bugs are found
- is an extra source of backup

design/README.md
scarlehoff: We are building a python app, I don't think the amount of modules is a measure of anything. In particular if we want to truly be independent and slim we should be using C (after all, the only tool we need to interact with is pineappl). Eko will be always used as an external tool so no need for python bindings.

>  do you want to make validphys a dependency of pineko or no? will there be validphys decorators?
Yes.

But honestly, this is not the hill I want to die on. I people feel so strongly against vp that's fine by me, but bear in mind that most of the work we do here will be eventually duplicated on the NNPDF repository and then will need to be maintained in both places. And eventually one (this one I'm betting) will stop being maintained and used as some external tool that some people did (think apfel) so it will be rebuilt again.

design/README.md
AleCandido: That's fine, we noticed that dealing with `git-lfs` on the cluster it may be harder because you need to install an extra client, and you may not have admin privileges and so on and so forth.

If the `git` repo is used just for the generation (man-work) and then they are mirrored on the server it would be perfectly fine.

design/README.md
scarrazza: I fully agree with the first point. Concerning the second point, I think the main problem is that we have been using applgrids from multiple providers, with lack of quality control. The db is just reshaping the grids to match the NNPDF specs, in terms of datapoints, units, masks, normalization, etc. With pineappl we can get rid of all that and create a 1:1 match between the dataset and the theory grid.

design/README.md
AleCandido: I'm not aware of the way @cschwan is managing mg5 inputs at the moment, but I agree on having a separate db for "theory generators", it will be a unique source of truth and it will make the FKtables fully reproducible.

design/README.md
cschwan: Exactly; I think to summarise my second point, what we should avoid is to put the relevant information in too many places - this makes all the procedures very error-prone and fragile, as we all know.

design/README.md
scarrazza: Frankly, I don't think the vp discussion is something to be worried about. We can see the vp as a "theory/dglap generator", like a program which calls eko, yadism and (why not) mg5 to generate the specific tables using the NNPDF theory database and dataset conventions.

Obviously, you could write another python script from scratch, while in vp everything NNPDF oriented is already interfaced through an API. So, you could store the yadims/eko/mg5 specifications in the vp repo and call the respective actions that will launch the artifacts generation jobs, and finally upload then to the server and repo.

design/README.md
cschwan: @scarrazza: This is not exactly on-topic, but I was thinking about a *universal runcard generator*: A problem that we still have is that the theory predictions are calculated using very different programs, and usually it's required to have a precise knowledge of all the intricacies of those programs. Especially for `mg5_aMC` the amount of extra script lines that I had to write is not small. At some point we might want to leverage SHERPA, as not all the processes can be run with `mg5_aMC`, and let's not forget yadism.

I think a worthwhile project would be to write some kind of 'universal runcard generator', where we bundle our knowledge of all of those programs, where we can write a **single** specification for the PineAPPL grids that we'd like to generate, with all the cuts and parameters, and where we ideally can choose the theory program if multiple programs can calculate the process.

I wonder whether this is worth the effort. This is of course yet another instance of 'every problem can be solved be another level of indirection', but in the long run I think it this universal runcard generator might be worth having; at some point we surely want to generate NNLO grids, and the variety of programs there is even larger than at NLO.

AleCandido: Yes, maybe you're are right that we should think carefully if it's needed or we are just adding extra complications, but my personal feeling is that it would be worth, because this is really the last missing link in making the runs fully reproducible.

I was thinking that it may be worth to design the specs as an abstract [schema](https://json-schema.org/) and then make it "pluggable", in the sense that we support some basic management, but then the bindings for each program are bundled as separate plugins (even if maybe shipped within the same project).

Then should a PineAPPLgrid be bundled with it's own generating runcard?

design/README.md
```diff
@@ -0,0 +1,169 @@
...
+The current proposal for the `eko` output format is the following:
+
+- for each pair of `(Q2_0, Q2)` there will be a rank 4 tensor `(fout, xout, fin, xin)`, saved in _numpy_
+  binary format
+  - it is possible to include binary in a `yaml` or to make an explicit
+    `npy/npz` file instead
+- since `numpy` it's loosing the information on the `dtype` and `shape` on save
+  it should be shipped together in a metadata file
+- since `eko` it's able to produce operators for multiple `Q2` at the same time
+  it is possible to collect them in a single compressed archive
```
AleCandido, line 129: This point is delicate:
- for `eko` purpose shipping a single zipped archive for all the generated operators it's much more natural
- this will couple arbitrarily objects that are computed together, only because of creation time, and may break (or make it much more difficult) the `md5sum` implementation

design/README.md
AleCandido: > Edit: I am not joking with the point about C. If we are not using anything that needs python I would actually prefer to do something in some other language (go is great for networking stuff they say!)

Nothing against any other language, but proliferating them will make you have a severe problem of the expertise you require to your team.

design/README.md
scarlehoff: Agreed. Unless we need the grouping for storage size or something it's better to have them all separated.

design/README.md
scarlehoff: Regarding the *universal runcard generator*, @cschwan, I'm surprised this is not something experimentalists are asking Sherpa and Madgraph (and other generators) developers to agree upon. I mean, there was an agreement for the output files after all. Do you know whether there are any efforts in this direction?

cschwan: I don't think there's something like this. In the (update to the Binoth) Les Houches Accord there's a specification for matrix elements, but that's not really what I'm thinking of. I suppose experimentalist would be *very* happy if such a program would exist.

scarlehoff: Indeed. Going even more off-topic (sorry) as this is maybe decoupled from this project but it would be interesting to provide such a tool (and therefore define such a standard). A sort of runcard reader that projects can use as a dependency (instead of using whatever they use).
Beyond the usefulness for experimentalists and us, I remember that back when I was doing fixed-orders there were bugs and inconsistencies that traced back to using some different parameter between NNLOJET and MCFM or Sherpa/Madgraph. And it was very much not obvious to find what was wrong.*

It is a complicated problem (political and otherwise) but it could be fun.

* going slightly on-topic, errors due to small inconsistencies did propagate to published NNLO results so it was very much non trivial until the end.

cschwan: I think it's actually not that difficult; what we have to do is to read runcards (our own format) and then write runcards and patches for the Monte Carlos (which is completely independent from what the MCs do, so no political problem) and start them.

As a starting point we could try to do that with Madgraph5 for which the `run.sh` script in https://github.com/scarrazza/applgridew basically already does it. The first step would be to port it to Python or Rust, and then to change the syntax of the runcards. For the time being we're only interested in PineAPPL grids, so there's also a well-defined output.

scarlehoff: Can't access that repo.
I mean it would be complicated to convince people to adopt the `pineko` standard. I was being ambitious there.

cschwan: @scarrazza : Can you give @scarlehoff access to it?

AleCandido: I think you are referring to https://github.com/scarrazza/applgridphoton, but there is no `run.sh`, what were you pointing to?

Let me propose again to use json schema for the specs, there are bindings for both rust and python available.

cschwan: @AleCandido : No, it's https://github.com/scarrazza/applgridew, where we have all the patches and runcards for the generation of the PineAPPL grids. Something like JSON for our runcards is definitely a good idea, but that's already step 2 ;).

AleCandido: Then I cannot access that repo...

scarrazza: Now you (all) should have access to repo. I think that this universal generator is a great idea, and in fact the script from @cschwan is already a good starting point.

cschwan:
@scarrazza had an interesting idea, and I wonder whether it makes sense to already talk about it in this context, because it may save us some work. The idea is based on the following observation: An FKtable *is* (a very special case of) a PineAPPL grid. Let me explain: Essentially, what apfelcomb does when it converts a PineAPPL grid to an FKtable, is to 1) compress the number of grid points in x, 2) evolve all values of the factorisation scale to a single Q0^2, and 3) change the basis from the flavour basis to the evolution basis. The result of the three operations can still be represented as a PineAPPL grid, so I wonder whether we shouldn't drop FKtables altogether for `pineko`.

Technically, what would happen is that we write a toolchain that uses EKO to convert one PineAPPL grid into another one. The change of the flavour basis to the evolution basis isn't a problem for PineAPPL, because it never assumes any basis; the entries are simply associated to integers, and we can always define what they mean: in the flavour basis PDG ids, in the evolutions basis something else. I'm not quite sure what would be needed from the fitting side, but since PineAPPL already has a, what I assume to be, quite efficient convolution method (all PDF values are evaluated exactly once for each (x, Q^2, flavour) or never, if the grid at this point is zero) the main operation is already there.

The potential advantages are: we might actually save some work because we can forget about FKtables, which also means that we reduce the number of tools involved even further: since everything would be a PineAPPLgrid, we can simply use the `pineappl` binary to produce predictions. In that case we already have a very simple way of determining the impact of the three operations: we simply use `pineappl diff <grid1> <grid2>` to check both grids against each other.

Of course, this proposal would change our plan above quite a bit, so comments and questions are appreciated!

scarrazza:
Yes, that sounds natural. However let me point out that from the fit perspective there are 2 stages.

In the first stage we could use the pineappl format as fktables (modulo extending it's header, etc) and providing to n3fit a numpy representation of the flavour map and weight matrix. 

In a second stage we could replace the tensordot operations with a custom operator that uses the convolution method implemented in pineappl. However this may be delicate given that we will loose gradients and the possibility to use gpus, which can obviously be implemented but may require a non negligible effort.

AleCandido:
I completely agree on the first step, to me it sounds the natural way of doing it.

It's true that we should not add too many features too many features on a single component, to make it as isolated as possible, but in this case we are mostly not touching `pineappl` at all, so it's just software reuse, and it will become a single tool to maintain factorizing future efforts. I like it.

On the second step I'm not the *juan* who should tell anything, because I never worked on the nn, but I agree with @scarrazza that it's delicate, because the current operation it's optimized for the nn pipeline, and I would not do anything in `pineappl` that would make it coupled to the nn (instead if you simply wish to implement the convolution on gpu it's not bad of course, but if in practice it's not needed for anything else than the nn I would keep it out of the boundaries of `pineappl`).

cschwan:
@scarrazza:

> In the first stage we could use the pineappl format as fktables (modulo extending it's header, etc) and providing to n3fit a numpy representation of the flavour map and weight matrix.

I think that's probably the best option; we simply agree on a format how the numpy arrays must look like. If we want to couple n3fit to the (Python API) of PineAPPL, we can simply add methods there that return the arrays.

> It's true that we should not add too many features too many features on a single component, to make it as isolated as possible, but in this case we are mostly not touching `pineappl` at all, so it's just software reuse, and it will become a single tool to maintain factorizing future efforts. I like it.

Exactly, that's the idea! Basically, we can start working on the PineAPPL/EKO tool right away: everything we need is basically there!

scarlehoff:
As @cschwan mentions, it can be coupled to `n3fit` easily if there is a python array at any point. Can't pineappl return the grid as a python object already?

cschwan:
> As @cschwan mentions, it can be coupled to `n3fit` easily if there is a python array at any point. Can't pineappl return the grid as a python object already?

Not yet, but it's trivial to write. We basically have to wrap the corresponding C function in Python and make it numpy array.

AleCandido:
@cschwan probably the best way of doing it as the one you said, but consider also that a `numpy` array is somehow already a C object python-bound, so maybe it would be even easier to pass directly from C (but it's just an option, it may also be 10x more time-consuming as far as I know)

PR #5, opened 2021-02-17
Init raw pineko functions

felixhekhorn:


AleCandido:
[pineappl_grids.zip](https://github.com/N3PDF/pineko/files/6070766/pineappl_grids.zip)

AleCandido:
Looks like the numbers we are producing by `convolute_eko()` are perfect, up to:
```rust
subgrid.grid = Some(array)
```
So the way we're saving them make is ruining the correct interpretation of the numbers themselves. I don't know yet what it is, the lumi function should be the same, but somehow:
- the subgrid is not the same, even if it should contain the same numbers
- the grid is not the same (but `bins` and `orders` should be trivial, and the `lumi` should be the same)

In the screenshot below you can see the non-zero `array` elements from `convolute_eko()` above, while the bottom numbers are just the one loaded by the `.yaml`, and they are the exact same (as it should be by applying identity):
image

AleCandido:
Remember that being LO it's just a delta function, that's why there's only a single x point significantly different from zero (we are also picking a point that is almost perfectly on the grid).
Moreover all the difference in the grid channels are just the absolute electric charge of each flavor, and thus:
- s~ = d~ = d = s
- c~ = u~ = u = c

Further differences are only coming from the PDF.

AleCandido:
SUCCESS!

image

AleCandido:
Many loops deserve a progress indicator: https://github.com/mitsuhiko/indicatif

src/pineko/__init__.py
```diff
@@ -95,17 +91,16 @@ def load_pineappl_dy():
...
+)
+
 # for the time being replace with a fake one, for debugging
 #  operator_grid = eko_identity(operator_grid.shape)
 
-pineappl_grid_q0 = pineappl_grid.convolute_eko(
-    operators["q2_ref"],
-    alpha_s(q2_grid),
-    operators["pids"],
-    x_grid,
```
felixhekhorn: This was an error, right? because it was coming from pineappl and so not in the order of eko, as was assumed - using https://github.com/N3PDF/pineappl/pull/51/files/114fbad8ca6291aadaf96c59065b8ac6ac45f301..c1a242435f64409d124227ff2b068a9a18f7d06e#diff-c0ce134e4c3635c8c19d90ea8ecbb1019377aeedd8c58d4cdde01b17a0eeaf66R47 we should be able to revert the fix we applied yesterday evening @AleCandido do you agree?

felixhekhorn: The comment refers only to line 105 = xgrid

AleCandido: Yes, indeed it was an error I fixed, but it's consistent with the current version of pineappl.

AleCandido: https://github.com/N3PDF/pineappl/blob/c1a242435f64409d124227ff2b068a9a18f7d06e/pineappl/src/grid.rs#L1039-L1045

AleCandido: https://github.com/N3PDF/pineappl/blob/c1a242435f64409d124227ff2b068a9a18f7d06e/pineappl/src/grid.rs#L1039-L1045

AleCandido:
I added a progress-bar to pineappl, we can remove at the end, but during this exercise it's actually very useful (and has no impact on timing in practice, since it's incrementing by 1 each couple of seconds or even more):
- we're taking ~21h in order to do DY (I tried with DIS and worked perfectly in a 3-4 seconds, not exactly measured)
image

AleCandido:
I also implemented the index translation with `HashMap`, recomputing them as little as possible, but the time is the one above (it was already implemented for that run).

Since DIS it's working now I really believe we should worry about performance optimization, and there are few strategies we can pursue:
- for sure profile it to see where it's spending most of the time
- recomputing even less making assumptions on grid structure
- try to skip more (if possible) when we know there are empty elements/sectors
- exploit eko specific features (i.e. do the convolution in evolution basis)

The last one I don't like, because we probably have to make pineappl aware of evolution physics, but rotating a pineappl grid should take very little, and we're already able to generate ekos in evolution basis.
Since in evolution basis an eko is almost diagonal (and *here be dragons*: in FFNS it's perfectly true for the light flavors, but for the flavors that are not light it's diagonal in flavor basis instead, and for ZM-VFNS it's even more complicated) we can save the square root of operations, on both the high and low pid index, that for 2 ekos (DY) means we can *almost* replace 4 iterations by 2, where each PID iteration is O(10), and so saving about a factor of 100.
It's not so straightforward, the gain can be much less and the implementation effort much more, considerably reducing maintainability and tools mutual independence.

At this point it's not doing so many things in the inner loops, but @cschwan might have a clearer picture.

cschwan:
I agree that we should talk about performance optimisations, since anything of the order of hours makes testing impossible. My bet is that the main culprit (bet = I haven't tested it) is [cache trashing](https://en.wikipedia.org/wiki/Thrashing_(computer_science)) inside the closure that the `Subgrid::convolute` method accepts. The main operation there is symbolically:

    for q2 in ... {
        for x1 in ... {
            for x2 in ... {
                result += op[q2, p, x1, i, j] * op[q2, p, x2, i, j] * grid[q2, x1, x2];
            }
        }
    }

If, for instance, `x2` is incremented in the innermost loop, we jump from the evaluation of `op[q2, p, x2, i, j]` to `op[q2, p, x2 + 1, i, j]`, which is very likely not to be cached, since it is `dim(i) * dim(j)` entries away. The situation is even worse when `q2` is incremented. To fix this, we should change the position of the indices from `op[q2, p, x1, i, j]` to something like `op[i, j, p, q2, x1]`. If you think about all the loops in `convolute_eko` the innermost loop should increment the last index of the operator, the next-to-innermost loop the one-but-last index, and so on and so forth.

We could do this by changing the memory layout of the operator in `pineappl_py`, so that you don't have to change it in EKO; doing it once should take a few seconds tops. We need something like https://docs.rs/ndarray/0.14.0/ndarray/struct.ArrayBase.html#method.swap_axes, but not exactly this method because it doesn't actually change the memory layout.

AleCandido:
I’m  pretty sure that numpy should be able to do it in very little time, much less than 20 hours :)

I don’t know if it’s the main culprit or not, but for sure it’s an optimization, and I believe you are perfectly right we should do it, and we’ll see.

felixhekhorn:
> because we want to treat as our main case the PDF evolution itself, that of course it’s the complete opposite

> the innermost loop should increment the last index of the operator, the next-to-innermost loop the one-but-last index, and so on and so forth.

- in deed: eko here is consistent!

- I'm also a bit surprised by the 21h - because loop's can not be a problem, they are trivial, so it has to be the lookup (as @cschwan argued) but then again, this is also quite trivial to do - and with the correct container the lookup should be O(N)
- if we want to swap axis around we could do it with numpy just before passing it down to Rust

AleCandido:
I believe that when Christopher said in `pineappl_py` he was referring to the python layer of `pineappl_py`, and thus in numpy.

Furthermore, I like to make all the magic that is convolution specific happening in the interface, rather than in eko itself, such that we produce EKOs in a unique way, and then the interface consume the given EKO to make its job.

cschwan:
>     * I'm also a bit surprised by the 21h - because loop's can not be a problem, they are trivial, so it has to be the lookup (as @cschwan argued) but then again, this is also quite trivial to do - and with the correct container the lookup should be O(N)

It's not the lookup, but the fact that reading from RAM is *extremely* slow (you should really have a look at the table here: https://stackoverflow.com/a/4087315/812178) compared to the CPU's clock speed. A common optimisation that CPUs use to hide the latency of accessing RAM is to perform the following operation: When you access an element of an array, the CPU doesn't just copy the requested element into its cache, but rather a slice of the array surrounding the requested element. That, I suppose, makes the latency worse in the first place, but if you access the elements of an array in the way in which it is laid out in memory this operation results in an overall optimisation.

One interesting question to answer would be why DIS is so much slower than DY, and I would bet that it is a mixture of cash trashing and the fact that your DIS grid is very simple. I think cash trashing hits DY much harder than DIS, because in general it happens in the inner-most loop, which is

    op[q2, p1, x1, i, j] * op[q2, p2, x2, i, j] * grid[q2, x1, x2]

We repeat this operation extremely often, but in the general case when `p1 != p2` and `x1 != x2` the corresponding elements are very far from each other in memory. That means cash-trashing in the inner-most loop, repeated over and over again, driving up the run time of the program. I wouldn't be surprised if we can make our program at least 100 times faster using the index reordering alone.

That being said, it's possible that I'm entirely wrong ...

> 
>     * if we want to swap axis around we could do it with numpy just before passing it down to Rust

There's surely a way to do it in Rust - we should do it inside `convolute_eko` - but so far I haven't properly searched for it.

cschwan:
@AleCandido , @felixhekhorn Do we want to meet again today? Now that DY works in principle, we should stress-test it and make some plans. I also updated the `diff` subcommand, which now accepts a switch `--ignore_orders`, so that it works when we compare a PineAPPL grid with a PineAPPL-FKtable.

AleCandido:
- scale variations
- preparation loop: join high_grid on orders
- clean up & optimize
  - parallelize?
- QED evolution

Issue #6, opened 2021-02-19
MHOU implement scheme B+C


felixhekhorn:
We should implement schemes B and C of https://inspirehep.net/literature/1741422

cschwan:
I hope this is the right place to ask this question: What do we need for a fit with theory uncertainties from the grid side and/or from Pineko and EKO?

If my understanding it correct, [lines like this one in apfelcomb](https://github.com/NNPDF/apfelcomb/blob/f49871435792bff1e269d524c3e6861aa6dd3e28/src/core/fk_appl.cc#L303) were used to approximately reconstruct the scale-varied FK tables. Do we have something similar in EKO or do we not need this?

AleCandido:
Just to write down what we told yesterday about scale variations reconstruction.

### Ren scale variations
If you check https://inspirehep.net/literature/1741422, in chapter 3.2.1 and 3.2.3 it's fully spelled out how to derive ren scale variation for DIS and double hadronic processes, based on the perturbative expansion of coefficient functions and of the beta function of alphas.

This thing is not that hard to reproduce, and we can add on top of PineAPPL: having a grid without renormalization logs, we can fill them algorithmically, based on pure QCD orders.
For mixed EW I guess might be a bit more complicated, but most likely if a single scale is used for the two couplings (not mandatory, but frequent, and we can always restrict to this case) we can rederive an analog formula.

### Fact scale variations
This is a bit more complicated, since they are coming from DGLAP (instead of coupling running), thus they have a non-trivial flavor structure.

> If my understanding it correct, [lines like this one in apfelcomb](https://github.com/NNPDF/apfelcomb/blob/f49871435792bff1e269d524c3e6861aa6dd3e28/src/core/fk_appl.cc#L303) were used to approximately reconstruct the scale-varied FK tables. Do we have something similar in EKO or do we not need this?

The derivative is used in formula 3.37 of the aforementioned paper, but a derivative in fact scale can be always traded for a derivative in alphas (plus beta coefficients), and a derivative in alphas can always be obtained from the perturbative series, manipulating the series with the same coefficients.
This is what is done in 3.39, and in principle we could use that to obtain fact scale variations for PineAPPL whenever they are lacking.

In practice, here you need not only beta functions series coefficients, but even the series coefficients of anomalous dimensions (and the easiest way to get them at this point is querying EKO), and they mix flavors in a pretty non-trivial way (since they are sensible in evolution basis, but crazy in flavor basis, that is more common for processes).
[...]

PR #12, opened 2022-03-01
Move permanent part of fkutils here

felixhekhorn:
This PR should have really started at https://github.com/N3PDF/pineko/commit/6d937476c993b29d35188c0f44aedc5be37bafad

- [x] log time taken for each *expensive* operation


AleCandido (changes requested):
Improve CLI, please :)

src/pineko/cli/theory_ekos.py
```diff
@@ -0,0 +1,47 @@
...
+import click
+import eko
+import rich
+import yaml
+
+from .. import configs, parser, theory_card
+from ._base import command
+
+
+@command.command("theory_ekos")
```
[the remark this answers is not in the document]
felixhekhorn: I can regroup them - but the scope will be theory (as Emmanuele uses "theory")

src/pineko/configs.py
```diff
@@ -122,7 +104,7 @@ def load(path=None):
             return {"paths": {"root": pathlib.Path.cwd()}}
         else:
             print("Configuration path specified is not valid.")
-            quit()
+            sys.exit()
```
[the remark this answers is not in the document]
felixhekhorn: pylint suggested to do so (R1722) - and it seems ok: https://www.geeksforgeeks.org/python-exit-commands-quit-exit-sys-exit-and-os-_exit/

src/pineko/cli/theory_ekos.py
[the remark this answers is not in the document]
felixhekhorn: - things related to theory are now bound below `theory`
- the name `theory` does not relate to the fact, that we're dealing with "theory" stuff here, as opposed to "experimental" - but to the fact that we're expecting a `theory_id` (and hence `theory` is not that bad of a name, although, maybe `TheoryFileAssets` whould be better, but it's too long) and we're iterating over datasets
- now if you say "all we do is like this" - I still want to distinguish between file based operations (`check,compare,convolute,opcard`) and theory based operations; if you really still want to regroup, I can see two options: or the file based move into their own scope and theory becomes root (reverse to the current layout) or we could provide two commands/executables: `pineko` (really meaning `pineko_theory`) and `pineko_file`

felixhekhorn:
I wonder whether to change here:
https://github.com/N3PDF/pineko/blob/784dee8e4a016fc033461fe9ac4e21d1ead67294/src/pineko/configs.py#L79-L83

and add a `else` after the if, because otherwise you get something like this:
```
$ pineko -c /home/a.toml theory opcards 213 LHCB_Z_13TEV_DIELECTRON
Configurations loaded from '/home/a.toml'
Analyze LHCB_Z_13TEV_DIELECTRON
Success: Wrote card with 1 Q2 points to 
/home/felix/Physik/N3PDF/FK/pineko/data/operator_cards/LHCB_DY_13TEV_DIELECTRON.yaml
```
where `/home/a.toml` never existed and the config truely used the default one

AleCandido:
Don't change `detect`: if you don't like this behavior, you should change `load`

https://github.com/N3PDF/pineko/blob/784dee8e4a016fc033461fe9ac4e21d1ead67294/src/pineko/configs.py#L101-L102

or even change where it's used:

https://github.com/N3PDF/pineko/blob/784dee8e4a016fc033461fe9ac4e21d1ead67294/src/pineko/cli/_base.py#L20-L24

Indeed, `detect` is already passing all the information: since it no one existed, it's even raising an error, that is later caught by `load`. So, `detect` has done its job.

I'd suggest not to call `detect` directly inside `load`, but to pass the `detect` result as `load` argument.
Consequently, the exception management should not be done inside `load`, but inside `command`, and so you can decide to print a different message, or even to just fail, and keep propagating the exception.

AleCandido:
> * what is `add_scope` supposed to do?

It's a function I'm using in `runcardsrunner` corresponding instance, to add defaults for an entire section, but preserving values that have explicitly been specified.

> (which would correspond to the `else` I mentioned)

This is false: the `else` would correspond to the case in which no path is specified. In that case either you return an error or you adopt some defaults. What is happening there is that you attempt a few sensible defaults, and if even those are not working, then raise.

> * first, it does not need to be a file, also a path would go through

What's the difference between a path and a file? On CLI you can only write strings, i.e. paths...

> * I dropped the `return {"paths": {"root": pathlib.Path.cwd()}}` because it would result in a ValueError right after as I have some mandatory configs

This is fine, I've defaults for everything in `runcardsrunner`, but it's more of a choice.

> * (finally I also added more docs)

Always a good idea.

src/pineko/cli/theory_.py
```diff
@@ -32,11 +32,11 @@ def opcards(theory_id, datasets, overwrite):
 @theory_.command()
 @click.argument("theory_id", type=click.INT)
 @click.argument("datasets", type=click.STRING, nargs=-1)
-@click.option("--logs", is_flag=True, help="dump logs")
+@click.option("--no-logs", is_flag=True, help="suppress logs")
```
AleCandido, line 35: Here and below: if you put directly `--no-logs` follow the pattern on the [docs](https://click.palletsprojects.com/en/8.0.x/options/#boolean-flags) and provide both, and set the default to `False`.

src/pineko/cli/theory_.py
```diff
@@ -32,11 +32,11 @@ def opcards(theory_id, datasets, overwrite):
...
 @click.option("--overwrite", is_flag=True, help="Allow files to be overwritten")
-def ekos(theory_id, datasets, logs, overwrite):
+def ekos(theory_id, datasets, no_logs, overwrite):
     """Compute EKOs for all FK tables in all datasets."""
-    theory.TheoryBuilder(theory_id, datasets, overwrite).ekos(logs)
+    theory.TheoryBuilder(theory_id, datasets, overwrite).ekos(no_logs)
```
AleCandido, line 39:
```suggestion
    theory.TheoryBuilder(theory_id, datasets, overwrite).ekos(not logs)
```

src/pineko/theory.py
```diff
@@ -222,46 +227,50 @@ def eko(self, name, _grid, tcard, logs):
...
+        logger.info("Start computation of %s", name)
+        start_time = time.perf_counter()
         ops = eko.run_dglap(theory_card=tcard, operators_card=ocard)
         ops.dump_tar(eko_filename)
+        logger.info("Finished computation of %s - took %f s", name, time.perf_counter() - start_time)
         if eko_filename.exists():
             rich.print(f"[green]Success:[/] Wrote EKO to {eko_filename}")
 
-    def ekos(self, logs):
+    def ekos(self, no_logs):
```
AleCandido, line 261:
```suggestion
    def ekos(self, silence):
```

src/pineko/cli/theory_.py
felixhekhorn: `True` you mean? because you said logs by default

src/pineko/theory.py
felixhekhorn: silent it is

felixhekhorn:
@AleCandido @scarlehoff @cschwan @andreab1997 

Dear all,

- I now believe to have all elementary stuff needed to build FK tables moved from https://github.com/NNPDF/fktables/blob/main/fkutils/builder.py to here, please take a look if you can and provide me with some feedback
- a tentative "manual" is given in README.md
- some improvements (@andreab1997 ) over the old implementation:
   - logs are now generated for every run (no more piping needed `python build.py > logs/eko/bla.log`)
   - grids can be inherited by command (instead of manual linking)
   - ekos can be inherited
   - no more prefixing with theory ID (so @scarlehoff you will be able to just copy the folder as is)
   - no more dependency on apfelcomb database but only ymldb (@scarlehoff I'm only reading the members for the rest I don't care)
   - no more special treatment of APPL thanks to ymldb
- some problems still to solve (can be done here or in a seperate PR):
  - allow scale variations
  - add argument to `parser.get_yaml_information` to allow only to ask for the name, so we can run eko without the actual grids being present
  - the hacks I needed for jets (i.e. https://github.com/NNPDF/fktables/blob/main/split_grid.py https://github.com/NNPDF/fktables/blob/main/jets_build.sh https://github.com/NNPDF/fktables/blob/86927f9460f8bbaf7f4daab0789c2d32283d3f2d/fkutils/builder.py#L76 https://github.com/NNPDF/fktables/blob/86927f9460f8bbaf7f4daab0789c2d32283d3f2d/fkutils/builder.py#L173 ) I did *not* port, but I hope https://github.com/N3PDF/eko/pull/105 will be available soon
  - I'd be happy to have some more opinions on https://github.com/N3PDF/pineko/pull/12#discussion_r818837993
- once we're happy there is a final step, i.e. moving the files from the old infrastructure to the new one - which mainly consists in stripping the prefixed theory ID and then to adjust the compare script to the new layout ...

cschwan:
@felixhekhorn cheers, I already had a quick look at the README, see my comments and edits there.

felixhekhorn:
@enocera of course I should have tagged you as well - apologies ;-) please feel free to take a look

src/pineko/cli/theory_ekos.py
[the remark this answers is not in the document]
AleCandido: However, I'm perfectly satisfied with the current layout.

The only difference between this and what I suggested is one name: `theory` in place of `compute`. I still think that they are grouped not because they are related to a theory ID, but because they are part of the theory computation process, instead of being a utility command (so called "file based operation").

> * the name `theory` does not relate to the fact, that we're dealing with "theory" stuff here, as opposed to "experimental" - but to the fact that we're expecting a `theory_id`

That's perfect, so `theory_id` is a wonderful name for a positional argument.

src/pineko/parser.py
```diff
@@ -1,10 +1,12 @@
-import yaml
-# ATTENTION: this is a partial copy from 
+# -*- coding: utf-8 -*-
+# ATTENTION: this is a partial copy from
 # https://github.com/NNPDF/nnpdf/blob/ec73c9c5d3765c8b600e3015d3f5d6238dd89400/validphys2/src/validphys/fkparser.py
```
AleCandido, line 3: Do we actually need this file at the end? Because I'd like to have a single copy, either here or in `vp`

AleCandido (commented):
I still have to dig `pineko.evolve` and `pineko.theory`, but everything else it's good enough.

I left a few comments, mainly about code I suppose to be not needed, and I suggest to just drop.

.pylintrc
```diff
@@ -0,0 +1,504 @@
+[MASTER]
```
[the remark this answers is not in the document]
felixhekhorn: done in [d2018a4](https://github.com/N3PDF/pineko/pull/12/commits/d2018a467c3860212a2636ea71abb7c037290eac)

_run.py
```diff
@@ -11,7 +11,7 @@
 
 from pineko import convolute
 
-data = pathlib.Path(__file__).absolute().parent / "data1"
+data = pathlib.Path(__file__).absolute().parent / "_data"
```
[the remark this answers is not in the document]
felixhekhorn: done in [24674cf](https://github.com/N3PDF/pineko/pull/12/commits/24674cf9564ecde93532c5c3f18b128446f8de01)

data/grids/load.sh
```diff
@@ -0,0 +1,8 @@
+# copy from remote
+
+# common example
+# scp dom.mi.infn.it:/media/FK/fktables/data/grids/200-LHCB_DY_13TEV_DIELECTRON.pineappl.lz4 ./common/LHCB_DY_13TEV_DIELECTRON.pineappl.lz4
+
+# theory example
+# scp dom.mi.infn.it:/media/FK/fktables/data/grids/200-SLAC_NC_EM_D_F2.pineappl.lz4 ./200/SLAC_NC_EM_D_F2.pineappl.lz4
+# scp dom.mi.infn.it:/media/FK/fktables/data/grids/200-SLAC_NC_EM_P_F2.pineappl.lz4 ./200/SLAC_NC_EM_P_F2.pineappl.lz4
```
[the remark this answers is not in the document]
felixhekhorn: yes, this is required for debug reasons (as we discussed to download local stuff) - and will be also needed in the future ...

src/pineko/__main__.py
```diff
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
 from . import command
```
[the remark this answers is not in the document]
felixhekhorn: Agreed, it is superseded by the CLI. Dropped in [b69353e](https://github.com/N3PDF/pineko/pull/12/commits/b69353e7296efa6d7673dadbf519fb7b670848f9)

src/pineko/parser.py
scarlehoff: Since we will depend on `eko` and `pineappl` I think it is only natural to depend on `pineko`.

I'm fine with adding extra flags that change the behavior as long as the "default" is what vp is doing. Once this PR is merged and we can test that vp is able to simply import this from pineko and that's it we can remove it from vp.

scarlehoff: On a second thought, this might mean moving `pineko` to `NNPDF` since it will need to be accessible by everyone (as this will be the package that is able to build and read the theories)

felixhekhorn: but the package is already public, so everybody can use it ... and maybe sooner or later we want to publish even on PyPI

AleCandido: No strong opinion on this topic, I'm just glad we all agree to remove duplication

AleCandido (approved):

scarlehoff:
Once this is merged and a package created I can try and use it in vp.

scarlehoff:
I'm going to merge this if nobody complains.

felixhekhorn:
I opened #13 and #14 for the remaining problems (SV is tracked in #6)

scarlehoff:
Then I can merge, right? As soon as we merge we can tackle the different issues one by one.

felixhekhorn:
For future reference: a bin-by-bin-evolution (as needed for jets) is still possible if needed after https://github.com/N3PDF/eko/pull/105 , but this time I would rather hack outside `pineko` then inside: we could define a bunch of fake-datasets: `CMS_1JET_8TEV-bin00000` containing exactly one member which itself contains exactly one bin /cc @cschwan

PR #17, opened 2022-03-29
Move the parsing of fktables to pineko

scarlehoff:
This is basically what it would be required to have vp depend on pineko.

Cons:
    - It is a bit of a complication for something that is only depending on vp (that is, the format of the dataframe)

Pros:
    - Gives us a bit more freedom on how to do anything since the only limitation is that at the end we have to have a dataframe as defined in `pineappl_to_fktable`. How do we get there is up to us. Vp doesn't need to know how the initial state were define, or what the luminosity means or anything.

### DIFF
--- a/src/pineko/parser.py
+++ b/src/pineko/parser.py
@@ -1,10 +1,46 @@
-# -*- coding: utf-8 -*-
-# ATTENTION: this is a partial copy from
-# https://github.com/NNPDF/nnpdf/blob/7cb96fc05ca2a2914bc1ccc864865e0ca4e66983/validphys2/src/validphys/pineparser.py
+"""
+    Loader for pineappl-based FKTables
 
+    The FKTables for pineappl have ``pineappl.lz4`` as extension and can be utilized
+    directly with the ``pineappl`` cli as well as read with ``pineappl.fk_table``
+
+    Examples
+    --------
+    >>> from pathlib import Path
+    >>> from pineko.parser import get_yaml_information, pineappl_to_fktable
+    >>> dataset = "ATLAS_TTB_DIFF_8TEV_LJ_TRAPNORM"
+    >>> yaml_folder = Path("/usr/local/share/NNPDF/data/yamldb")
+    >>> pine_folder = Path("/usr/local/share/NNPDF/data/theory_200/pineappls/")
+    >>> yaml_file = yaml_folder / f"{dataset}.yaml"
+    >>> yaml_info, paths = get_yaml_information(yaml_file, pine_folder)
+    >>> fk = pineappl_to_fktable(yaml_info, paths[0])
+    >>> fk.dataframe
+15         16        17        18   ...       167       180       181       195
+data x1 x2                                           ...                                        
+0    0  0   0.000000   0.000000  0.000000  0.000000  ...  0.000000  0.000000  0.000000  0.000000
+        1   0.000000   0.000000  0.000000  0.000000  ...  0.000000  0.000000  0.000000  0.000000
+        2   0.000000   0.000000  0.000000  0.000000  ...  0.000000  0.000000  0.000000  0.000000
+        3   0.000000   0.000000  0.000000  0.000000  ...  0.000000  0.000000  0.000000  0.000000
+        4   0.000000   0.000000  0.000000  0.000000  ...  0.000000  0.000000  0.000000  0.000000
+...              ...        ...       ...       ...  ...       ...       ...       ...       ...
+4    99 95  0.000000  -0.724031 -0.000224  0.000140  ... -0.002200  0.000000 -0.013380  0.000000
+        96  0.000000  -0.722401 -0.000193 -0.000007  ... -0.002141  0.000000 -0.014283  0.000000
+        97  0.000000  -0.513297 -0.000059  0.000128  ... -0.001426  0.000000 -0.010557  0.000000
+        98  0.000000  -0.841811  0.000018  0.000349  ... -0.002255  0.000000 -0.016986  0.000000
+        99 -0.004565  74.340392  0.001172 -0.002813  ... -0.015495 -0.000411 -0.000548 -0.000183
+"""
+from collections import namedtuple
 import yaml
+import numpy as np
+import pandas as pd
+
+from eko.basis_rotation import evol_basis_pids
+from pineappl.fk_table import FkTable
 
 EXT = "pineappl.lz4"
+
+FkData = namedtuple("FkData", ["dataframe", "q0", "xgrid"])
+
 
 class YamlFileNotFound(FileNotFoundError):
     """ymldb file for dataset not found."""
@@ -35,7 +71,54 @@ def _load_yaml(yaml_file):
     return ret
 
 
-def get_yaml_information(yaml_file, grids_folder):
+def _apfelcomb_compatibility_flags(gridpaths, metadata):
+    """
+    Prepare the apfelcomb-pineappl compatibility fixes
+
+    Returns
+    -------
+        apfelcomb_norm: np.array
+            Per-point normalization factor to be applied to the grid
+            to be compatible with the data
+        apfelcomb_repetition_flag: bool
+            Whether the fktable is a single point which gets repeated up to a certain size
+            (for instance to normalize a distribution)
+        shift: list(int)
+            Shift in the data index for each grid that forms the fktable
+    """
+    operands = metadata["operands"]
+    apfelcomb_norm = None
+    if norms := metadata.get("apfelcomb_norm"):
+        # There's no easy way for an fktable to know its role in given operation:
+        for factors, grids in zip(norms, operands):
+            # Now check, in case of an operation, what is our index in this operation
+            if len(grids) == len(gridpaths) and all(
+                f.name == f"{o}.{EXT}" for f, o in zip(gridpaths, grids)
+            ):
+                apfelcomb_norm = np.array(factors)
+
+    # Check for the repetition flag, meaning we only want the first datapoint for this fktable
+    apfelcomb_repetition_flag = False
+    if metadata.get("repetition_flag"):
+        valid_targets = []
+        for operand, flagged in zip(operands, metadata["repetition_flag"]):
+            if flagged:
+                valid_targets.append(f"{operand[0]}.{EXT}")
+        # Now check whether the current fktable is part of the valid targets
+        apfelcomb_repetition_flag = gridpaths[0].name in valid_targets
+        if apfelcomb_repetition_flag and len(gridpaths) > 1:
+            raise ValueError(f"Repetition set for a group of fktables at once: {gridpaths}")
+
+    # afaik there's only dataset with shifts, but it needs to be considered
+    shifts = None
+    if metadata.get("shifts"):
+        if len(operands) > 1:
+            raise ValueError("Wrong shifts for {metadata['target_dataset']}")
+        shifts = [0 if shift is None else shift for shift in metadata["shifts"][0]]
+    return apfelcomb_norm, apfelcomb_repetition_flag, shifts
+
+
+def get_yaml_information(yaml_file, grids_folder, check_grid_existence=True):
     """Given a yaml_file, returns the corresponding dictionary and grids.
 
     The dictionary contains all information and we return an extra field
@@ -47,6 +130,8 @@ def get_yaml_information(yaml_file, grids_folder):
         path of the yaml file for the given dataset
     grids_folder : pathlib.Path
         path of the grids folder
+    check_grid_existence: bool
+        if True (default) checks whether the grid exists
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


AleCandido:
Just to tell it: I'm not a fan of linearizing dimensions.

We have the luck of to have a simple and powerful framework at hand, that is able to deal with N-dimensional tensors in a fast and expressive way.

FkTables are genuinely 4 dimensional objects (bin, lumi, x1, x2), so we should use 4 dimensional data structures, not try to fit it into a 2 dimensional one.

If you want labels, concatenation and other excel-like swag, giving up on part of the simplicity, than you should consider to use `xarray`, still based on `numpy`, but at least it doesn't have the tradeoff on the rank.

scarlehoff:
And you haven't seen the terrible crap in the other side to get the pandas df and recover exactly the 4rank tensor that I had at the beginning :__

Sadly vp uses the df everywhere (most notably, for the old tables). I can move the tensor to df part to the vp parser to that this side is clean. I think that makes sense.

AleCandido:
:disappointed:

src/pineko/parser.py
```diff
@@ -35,7 +71,54 @@ def _load_yaml(yaml_file):
...
+            to be compatible with the data
+        apfelcomb_repetition_flag: bool
+            Whether the fktable is a single point which gets repeated up to a certain size
+            (for instance to normalize a distribution)
+        shift: list(int)
+            Shift in the data index for each grid that forms the fktable
+    """
+    operands = metadata["operands"]
+    apfelcomb_norm = None
+    if norms := metadata.get("apfelcomb_norm"):
```
AleCandido, line 91: Please, avoid walrus...

AleCandido: And test for `bool`, not for truthiness...
(i.e. make the explicit check: if you expect float, test for a float, if you simply want to test if the field is present, test for `is not None`)

src/pineko/parser.py
```diff
@@ -35,7 +71,54 @@ def _load_yaml(yaml_file):
...
+        # There's no easy way for an fktable to know its role in given operation:
+        for factors, grids in zip(norms, operands):
```
AleCandido, line 93: Please,  `raise` if `len(norms) != len(operands)`,  `zip()` will silently cut one of the two, and this is an expected behavior for the user of the function.

src/pineko/parser.py
```diff
@@ -63,9 +148,132 @@ def get_yaml_information(yaml_file, grids_folder):
...
-            if not p.exists():
+            if not p.exists() and check_grid_existence:
                 raise GridFileNotFound(f"Failed to find {p}")
             tmp.append(p)
         ret.append(tmp)
 
     return yaml_content, ret
+
+
+def _pinelumi_to_columns(pine_luminosity, hadronic, flav_size=14):
```
AleCandido, line 159: If you're going to move this out, maybe I'll carefully review it only on the other side.

However, I haven't tested, but it looks mostly fine, apart from:
1. `co` is a cryptic name for a variable (I understand is a column, but took me a while, so maybe you can call it `col_indices`, `column_indices`, `col_labels`, or something like...)
2. the value 14 that is a default here, it's not modifiable for the user of `pineappl_to_fktable`, resulting in an hardcoded value on this side (`pineko`), that you will have to match on the other side (`vp`); one more reason to move it...

src/pineko/parser.py
```diff
@@ -35,7 +71,54 @@ def _load_yaml(yaml_file):
...
+        for factors, grids in zip(norms, operands):
+            # Now check, in case of an operation, what is our index in this operation
+            if len(grids) == len(gridpaths) and all(
+                f.name == f"{o}.{EXT}" for f, o in zip(gridpaths, grids)
+            ):
+                apfelcomb_norm = np.array(factors)
+
+    # Check for the repetition flag, meaning we only want the first datapoint for this fktable
+    apfelcomb_repetition_flag = False
+    if metadata.get("repetition_flag"):
```
AleCandido, line 102: Again, I'd test for `is not None`

src/pineko/parser.py
```diff
@@ -35,7 +71,54 @@ def _load_yaml(yaml_file):
...
+            if flagged:
+                valid_targets.append(f"{operand[0]}.{EXT}")
+        # Now check whether the current fktable is part of the valid targets
+        apfelcomb_repetition_flag = gridpaths[0].name in valid_targets
+        if apfelcomb_repetition_flag and len(gridpaths) > 1:
+            raise ValueError(f"Repetition set for a group of fktables at once: {gridpaths}")
+
+    # afaik there's only dataset with shifts, but it needs to be considered
+    shifts = None
+    if metadata.get("shifts"):
```
AleCandido, line 114: Again, I'd test for `is not None`

src/pineko/parser.py
```diff
@@ -35,7 +71,54 @@ def _load_yaml(yaml_file):
...
+            if len(grids) == len(gridpaths) and all(
+                f.name == f"{o}.{EXT}" for f, o in zip(gridpaths, grids)
+            ):
+                apfelcomb_norm = np.array(factors)
+
+    # Check for the repetition flag, meaning we only want the first datapoint for this fktable
+    apfelcomb_repetition_flag = False
+    if metadata.get("repetition_flag"):
+        valid_targets = []
+        for operand, flagged in zip(operands, metadata["repetition_flag"]):
```
AleCandido, line 104: Again, test for lengths, to avoid unexpected behavior

AleCandido (changes requested):
First batch, I still have to review the main API `pineappl_to_fktable`

src/pineko/parser.py
```diff
@@ -1,11 +1,47 @@
...
+    --------
+    >>> from pathlib import Path
+    >>> from pineko.parser import get_yaml_information, pineappl_to_fktable
+    >>> dataset = "ATLAS_TTB_DIFF_8TEV_LJ_TRAPNORM"
+    >>> yaml_folder = Path("/usr/local/share/NNPDF/data/yamldb")
+    >>> pine_folder = Path("/usr/local/share/NNPDF/data/theory_200/pineappls/")
+    >>> yaml_file = yaml_folder / f"{dataset}.yaml"
+    >>> yaml_info, paths = get_yaml_information(yaml_file, pine_folder)
+    >>> fk = pineappl_to_fktable(yaml_info, paths[0])
+    >>> fk.dataframe
```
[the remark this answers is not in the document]
scarlehoff: It's a path where the things are in dom. It should be reproducible.

scarlehoff:
>  I still have to review the main API pineappl_to_fktable

Wait for that one though, I'm going to remove the df part and move it back to vp.

src/pineko/parser.py
scarlehoff: > Please, avoid walrus...

Why?

src/pineko/parser.py
scarlehoff: No, if `len(norms) != len(operators)` that's fine. It will continue its path. I can make it not enter the loop if that's the case if you prefer it.

src/pineko/parser.py
scarlehoff: This can stay here though, since where the columns are is defined by eko. This one I don't know tbh.

src/pineko/parser.py
scarlehoff: But I want (and also below) to not enter in the case of `None`, in the case of `False` and in the case of empty list. This checks all 3 cases. What would you put instead?

src/pineko/parser.py
AleCandido: > Why?

It deteriorates readability. Having side effects on assignment makes less intuitive which is the actual expression (plus is easily avoidable 3.8 feature, we gain nothing to little, and we pointlessly reduce compatibility with older Python versions, but this is a minor detail, we explicitly do not support <3.8 in any case).

AleCandido: However, it's a detail in any case, so do as you wish.

src/pineko/parser.py
AleCandido: If `len(norms) == len(operators) - 1` because of a typo, you will have one element less used in `operators` silently...

src/pineko/parser.py
AleCandido:
```suggestion
    if metadata.get("repetition_flag", False):
```

If it's missing `get` will give you `False`, an explicit `null` in the yaml should not be supported (if the correct setting should be `false`).
For the empty list I'd say that the body should gracefully behave even in that case (if it's a meaningful empty list, corresponding to empty something else). If instead the empty list is an alias for `false`, only support `false`.

AleCandido: In this way you're currently having an expected behavior e.g. for a `0.` normalization: it should be multiply by `0.`, not skipping.

src/pineko/parser.py
scarlehoff: If there's a typo in the yaml file I cannot do anything. I can add checks for warnings for weird stuff, but the problem is the apfelcomb stuff are afterthoughts so they _can_ be weird stuff.
I can try to look through the yaml db to improve on them though. It will also fix some of the other issues you raised...

src/pineko/parser.py
AleCandido: In any case, I believe that actually the meaningful value to deactivate is `null`, and not `false`, since there is no meaningful case in which it would be `true` (the usual non-empty value is a float, so empty should be `null`).

So my suggestion would be to drop support for false, and actually use:
```suggestion
    if metadata.get("repetition_flag") is not None:
```

scarlehoff:
I will run through the yaml db to standarize how the apfelcomb flags are added since many were added manually anyway. That way I can fix most issues at once without adding checks for it in here.

src/pineko/parser.py
AleCandido: Then it should be accessible even from the caller, or if 14 it's really an hard-coded value, instead of putting as a default argument, make it a module level constant  (i.e. an uppercase named variable).

src/pineko/parser.py
felixhekhorn: I think I agree with @AleCandido : side-effects are a dangerous C heritage (also agree that it's a detail)

scarlehoff: Well, 2 against 1. I'll remove it. But [I'm crying](https://www.youtube.com/watch?v=TKuvJLTeJYY)

src/pineko/parser.py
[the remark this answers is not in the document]
AleCandido: I would invoke @felixhekhorn experience to check if it's possible to create a tiny simple mock environment for testing.

scarlehoff:
```
apfelcomb:
    normalization:
        ATLAS_TTB_8TEV_2L_TTRAP: [2.0]
    repetition_flag:
        CMSTTBARTOT8TEV-TOPDIFF8TEVTOT: True
```

So that instead of relying on things being by chance in the same order or whatever, the behavior is explicit for each subfktable. This is for `ATLAS_TOPDIFF_DILEPT_8TEV_TTRAPNORM` for instance.

If you guys agree I'll go ahead. In this case for instance only the `ATLAS_TTB_8TEV_2L_TTRAP` fktable needs a factor and only the `CMSTTBARTOT8TEV-TOPDIFF8TEVTOT` needs to set the repetition flag.

src/pineko/parser.py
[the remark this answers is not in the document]
scarlehoff: I guess we can have a mock theory in the repo.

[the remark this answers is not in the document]
felixhekhorn: Definetly - adding tests to `pineko` is one of the next things on my agenda

felixhekhorn:
> In this case for instance only the `ATLAS_TTB_8TEV_2L_TTRAP` fktable needs a factor and only the `CMSTTBARTOT8TEV-TOPDIFF8TEVTOT` needs to set the repetition flag

If things decouple, perfect!

src/pineko/parser.py
scarlehoff: I guess it's "hardcoded" but it should be hardcoded to len(whatever eko size is).

scarlehoff:
Meh, there's no good solution for this.

The apfelcomb flags basically set up the dataframe for vp so either everything is in vp or everything is in here.
I would prefer to have the apfelcomb flags here so that they actually get removed, if they get into the vp side they will be abused forever.

AleCandido:
> The apfelcomb flags basically set up the dataframe for vp so either everything is in vp or everything is in here. I would prefer to have the apfelcomb flags here so that they actually get removed, if they get into the vp side they will be abused forever just like apfelcomb was.

Fine, then I'd immediately open a separate issue for apfelcomb flags total disruption: as soon as we completed theory uncertainties and we have new commondata, we are stable enough and we drop backward compatibility (the last faints of it).

felixhekhorn:
I'm still no big fan of this:
- it adds highly non-trivial functionality beyond the scope of the package: the unique scope in my eyes is the production of FK tables
- this should instead belong to a CommonDataReader (as also said here: https://github.com/NNPDF/nnpdf/blob/646dd01c2b9e60ebf1ac98cf52b4ed5c6efa6f62/validphys2/src/validphys/pineparser.py#L16) - which most likely is the long-term solution
- for the short term I suggest to add a condensed, 5-line version of if: read the file and concatenate the content of `operands`
  - Pros: quick and easy
  - Cons: slight code duplication, though only until CDR gets settled
- the other unfavoured option is to give up on datasets:
  - Pros: no additional dependency what so ever
  - Cons: you have to know FK table member names, which seems unreasonable to users (but not to scripts): `ATLAS_WP_JET_8TEV_PT-atlas-atlas-wjets-arxiv-1711.03296-xsec002`
- the least favoured option is to depend on `vp`:
  - Pros: well ...
  - Cons: infinite (truly - since I'd like to get rid of non-trivial dependencies)

scarlehoff:
I'm currently reducing the amount of stuff here but

> it adds highly non-trivial functionality beyond the scope of the package: the unique scope in my eyes is the production of FK tables

Indeed

> this should instead belong to a CommonDataReader 

That's what will happen (but for now there is no such thing)

The (my) idea is that only the apfelcomb flags will remain here. I am very interested on this being the case because that way nobody can rely on them in vp (since they are coming from some external package).

felixhekhorn:
> The (my) idea is that only the apfelcomb flags will remain here. I am very interested on this being the case because that way nobody can rely on them in vp (since they are coming from some external package).

for sure we can intermediate stuff here, since everything is still under heavy development - but I believe the correct place for this should be https://github.com/NNPDF/runcards/issues/123 as I consider this under correct grid generation and not FK generation

scarlehoff:
But this is not only for converted grids, these flags sometimes are necessary because of a mismatch between cfactors / cuts etc and what pineappl grids have.

tbh, it can be anywhere but it needs to be something vp can import _and_ be somewhere the people writing vp cannot touch.

PR #18, opened 2022-05-19
Initialize documentation

AleCandido:

PR #19, opened 2022-05-19
Scale Variations

### LINKED ISSUE #6: MHOU implement scheme B+C

felixhekhorn:
Closes #6


AleCandido:
Remember to add to the documentation (see PR #18) considerations about schemes: `pineko` is the best place, since both EKO and yadism see only a part of it.

Moreover, you'll need most likely to patch something even in EKO. Indeed, if you apply factorization scale variations on the DGLAP side, and you use the expanded mode [1]:
1. `fact_to_ren` (still looking for a better name) has to set different from 1, but I'd say it should really be `xif`; indeed, as for Eqs. 3.34 - 3.35 of our beloved arXiv:1906.10698v2, you'll evolve the PDF up to `t + k = log(Q^2 * xiF^2) = log(muF^2)`, so the renormalization scale plays no role in here
2. `fact_to_ren` is not passed to main evolution step, following Eq. 3.35: we are now computing `f(t + k) := fbar(as(t + k), 0)`, so the end-point is changed, but no correction is applied here (to the "resummation" scale, according to the nomenclature of a recent popular talk ^^)
3. `fact_to_ren` is used in computing the expanded correction, explicitly quoted in Eq. 3.35

If `ModSv = exponentiated` is chosen, then no expanded correction has to be computed, but `fact_to_ren` has to be passed in step 2. to the shift the point in which anomalous dimensions are evaluated (i.e. to shift the resummation scale wrt the factorization scale in which we're evolving), that corresponds to Eq. 3.34.

@andreab1997 please include all of this (possibly even improved with links and better descriptions) in the docs.

[1]: default for scheme B, even though not the one implemented by APFEL (that use exponentiated, even though it should be almost the same thanks to the adoption of TRN solution)

felixhekhorn:
Closes #23

felixhekhorn:
let's start by merging this - a significant amount of the changes is due to #23

src/pineko/evolve.py
```diff
@@ -39,10 +42,10 @@ def write_operator_card_from_file(pineappl_path, default_card_path, card_path):
     pineappl_grid = pineappl.grid.Grid.read(pineappl_path)
     with open(default_card_path, "r", encoding="UTF-8") as f:
         default_card = yaml.safe_load(f)
-    return write_operator_card(pineappl_grid, default_card, card_path)
+    return write_operator_card(pineappl_grid, default_card, card_path, xif)
 
 
-def write_operator_card(pineappl_grid, default_card, card_path):
+def write_operator_card(pineappl_grid, default_card, card_path, xi):
```
andreab1997, line 48: Why do you want to call it `xi`? It is always `xif`, right? So maybe would be better to just call it `xif`

src/pineko/evolve.py
```diff
@@ -76,54 +82,80 @@ def evolve_grid(
...
+    max_al : int
+        maximum power of electro-weak coupling
+    xir : float
+        renormalization scale variation
+    xif : float
+        factorization scale variation
+    alphas_values : None or list
+        values of strong coupling used to collapse grids
+    assumptions : str
+        assumtions on the flavor dimension
```
andreab1997, line 112: assumptions :)

src/pineko/theory.py
```diff
@@ -1,10 +1,19 @@
...
+commenly refered to as 'theory'.
+"""
 import logging
 import time
 
 import eko
+import numpy as np
 import rich
 import yaml
+from eko import strong_coupling as sc
```
andreab1997, line 16: Maybe would be better to import eko only once but I don't know, just wondering

src/pineko/theory.py
```diff
@@ -192,7 +201,7 @@ def iterate(self, f, **kwargs):
                 f(name, grid, **kwargs)
             rich.print()
 
-    def opcard(self, name, grid):
+    def opcard(self, name, grid, tcard):
```
andreab1997, line 204: here we only need `xif` from the theory card, right? Then it is maybe better to pass only `xif` to `opcard` instead of the whole card

src/pineko/evolve.py
AleCandido: Just because @felixhekhorn likes to be able to click on the button...

src/pineko/theory.py
```diff
@@ -1,10 +1,19 @@
 # -*- coding: utf-8 -*-
+"""Tools related to generation of a list of FK tables.
+
+The typical use case of pineko is the generation of a list of FK tables,
+all with common theory parameters. The collective list of this FK tables
+together with other theory ingredients (such as C-factors) are often
+commenly refered to as 'theory'.
```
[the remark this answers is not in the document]
AleCandido:
```suggestion
commonly referred to as 'theory'.
```

src/pineko/theory.py
AleCandido: `import eko` is not enough, because `eko.strong_coupling` would not be defined (for that to happen, we need to `import strong_coupling` in `__init__.py`, that is not the case - and most likely we don't want to, at the moment, since we want to directly expose only a minority of things that we believe the user is going to use more often).

You'd need at least:
```py
import eko.strong_coupling
```
Not much different.

AleCandido: If you refer to import it multiple times in the package, that is not a problem (if a module is already loaded at runtime, Python won't load it again - modules are runtime singletons).

andreab1997: Yes I was thinking to use just `import eko` and then use the function as `eko.strong_coupling.whatever_func`. However,  if importing twice is not a performance problem as you said, it is just fine to import twice.

andreab1997 (approved):

PR #20, opened 2022-05-30
Update repo structure

andreab1997:
We want to have a new README with:
- [x] logo
- [x] badges
- [x] installation instructions 
- [x] docs
- [x] contributing 

PS: Do we want to add also `codecov`?


AleCandido:
Please gitignore MacOS files (i.e. `.DS_Store` files), and remove them from the repo :)

In general, I would suggest you not to put them in the repo `.gitignore`, because this way you should do for each project/repo, but much better to put globally in your git configurations (on Linux `$HOME/.config/git/ignore`, or even better `$XDG_CONFIG_HOME/git/ignore`, but you can set your favorite file in the configurations, like specified here: https://git-scm.com/docs/gitignore#_configuration)
Of course, the moment you do it for mac files, please add also VS code once for all ^^

AleCandido:
For the work flows actually you don't have to redo everything from scratch: if you look inside `eko` you will find out that it's making use of some reusable workflow.
Exploit the same (it will enhance maintenability).

andreab1997:
> For the work flows actually you don't have to redo everything from scratch: if you look inside `eko` you will find out that it's making use of some reusable workflow. Exploit the same (it will enhance maintenability).

Yes, I noticed. However the eko workflow includes also benchmarks and for the moment in pineko we miss even unit tests. In fact, I copied most of the EKO workflow but I simply removed the benchmark part (and I have called the file with a different name :) )

AleCandido:
> Yes, I noticed. However the eko workflow includes also benchmarks and for the moment in pineko we miss even unit tests. In fact, I copied most of the EKO workflow but I simply removed the benchmark part (and I have called the file with a different name :) )

Then you copied an old workflow, or `isobench.yml`, that contains **only** isolated benchmarks.

Since you are going for unit tests, the correct workflow is simply [`unittest.yml`](https://github.com/N3PDF/pineko/pull/20#issuecomment-1141815067), that contains nothing but unit tests, and that is the reusable one.
But if unit tests are already covered, why are you adding an extra `workflow.yml` (which is a too generic name) that is running again unit tests?

AleCandido:
Take into account that the reusable workflow already contains [Codecov upload](https://github.com/N3PDF/workflows/blob/9cd4b87663cd4e44a0fcec9fb30a1cd5f65b1352/.github/workflows/python-poetry-tests.yml#L64-L70) (otherwise it would not be possible that `eko` unit tests are uploaded).

AleCandido:
Other remark: install `pre-commit` in your local repository, there is some not normalized code.

.github/codecov.yml
```diff
@@ -0,0 +1,43 @@
...
+    project:
+      default:
+        # check only master and branches that should be merged directly into
+        # master
+        branches:
+          - main
+          - develop
+          - "feature/*"
+          - "release/*"
+          - "change_readme"
```
AleCandido, line 20: Remove all the restrictions on branches, we are not following git flow model in this repository.

AleCandido: Most likely, you can just kill the `status` key, but check for it.

.github/workflows/unittests.yml
```diff
@@ -22,3 +22,5 @@ jobs:
     uses: N3PDF/workflows/.github/workflows/python-poetry-tests.yml@main
     with:
       python-version: "3.10"
+
+
```
AleCandido, line 26: This is the effect of missing `pre-commit` installation. It would have removed automatically blank lines at the end of the file.

.gitignore
```diff
@@ -132,3 +132,6 @@ dmypy.json
 
 # visual studio
 .vscode/
+
+# mac files
+.DS_Store
```
AleCandido, line 137: This (and even the `.vscode/` above) I really suggest removing from here, and put in your system (actually user) gitignore.
That is the correct level to put this information, since these files are not generated by tools used by the project, but by tools present on your system (thus it is overhead for the project, and it should be repeated for every project).

Moreover, there are still 3 `.DS_Store` files checked in:
image

AleCandido (changes requested):

.github/workflows/unittests.yml
andreab1997: yes sorry for that. I have installed it now

.gitignore
andreab1997: I am doing that now

andreab1997: Ok solved now

AleCandido (approved):
Everything else can be done in a separate PR (e.g. a nice follow up to coverage setup would be to actually have some unit tests).

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,54 @@
...
+- :art: Run [black](https://github.com/psf/black) to style your code
+- :blue_book: Use [numpy documentation
+  guide](https://numpydoc.readthedocs.io/en/latest/format.html)
+
+### Git
+
+- :octocat: Make sure the commit message is written properly ([This
+  blogpost](https://chris.beams.io/posts/git-commit/) explains it
+  nicely)
+- :sailboat: Use [git-flow](https://github.com/nvie/gitflow) - note
```
felixhekhorn, line 50: if we're not planing on using `git-flow` let's remove it from here

AleCandido: You're right, I missed it

felixhekhorn (approved):

Issue #13, opened 2022-03-29
Do actual `optimize`


felixhekhorn:
While `optimize` *is* called here https://github.com/N3PDF/pineko/blob/ee6b9da3e7ed6078f2816543cf3952399beaba04/src/pineko/evolve.py#L112 it is currently called with it's default value `def optimize(self, assumptions = "Nf6Ind"):` see https://github.com/N3PDF/pineappl/blob/101cbdefdd3b9f6c0427de48e57bca53e0f3f843/pineappl_py/pineappl/fk_table.py#L37

instead we should add a field to the theory card and use that value

AleCandido:
Available in a moment with the `pineappl` upgrade to v0.5.2

cschwan:
Here you go: https://pypi.org/project/pineappl/0.5.2/

AleCandido:
@felixhekhorn @andreab1997 if you start replacing compatibility layer, take this into account.

AleCandido:
@felixhekhorn @andreab1997 what is missing here?

felixhekhorn:
Nothing - but time :upside_down_face: ; maybe @andreab1997 you can take care of this?

andreab1997:
> Nothing - but time 🙃 ; maybe @andreab1997 you can take care of this?

Yes, of course

andreab1997:
Only to be sure I have understood, we want to add an argument to the theory card (called something like `assumptions`) and use that value as second argument of the function `optimize` instead of using its default value, right?

AleCandido:
But let's start here the "compatibility" project.

This argument was not present before, but we can infer from:
- Q0
- thresholds (matching scales)
- IC

At some point we'll further change (e.g. an ICV, or similar, will appear, at which point might be easier to directly go for assumptions), but let's assume an adiabatic approximation: everything is as it has been for the past few years, changes will be applied on top.

andreab1997:
Do we want to close this issue now that the PR #22 has been merged?

felixhekhorn:
yes

PR #22, opened 2022-06-07
Provide automatically the assumption hash from tcard info

andreab1997:
We want to provide a function that automatically computes the assumption hash from the information in the theory card. At the moment we will consume Q0, the matching scales and IC. 

This should solve #13


andreab1997:
@AleCandido @felixhekhorn It is not completely clear to me how to consume the information of intrinsic charm (and eventually of intrinsic b and t). For what I have understood, the possible assumption hashes are listed [here](https://docs.rs/pineappl/latest/pineappl/fk_table/enum.FkAssumptions.html). However, I believe that for NF = 3 none of them allow the charm to be different from zero (as we would have in the intrinsic case). In fact, the difference between "Nf3Ind" and "Nf3Sym" is only in how they treat the strange. Am I missing something?

AleCandido:
Yes, the idea is that PineAPPL on its side doesn't care so much if an object is intrinsic or perturbative.

Let's focus on charm (don't care about bottom and top for the time being):
- if you are fitting in the FFNS4, then your Q0 is already determining that you are in the 4 flavors patch, and of course you have to choose `Nf4Ind` $\to$ the contributions of bottom and top are stripped, but all the possible contributions for charm are kept
- if you are fitting in the FFNS3, but without IC, then same thing with `Nf3Ind`
- if you are fitting in the FFNS3, with IC, but without ICA (intrinsic charm asymmetry, i.e. $c - \bar{c}$) then you have to choose `Nf4Sym` $\to$ you want to retain charm contribution, but you don't need the contributions for $c \neq \bar{c}$, since there would be no PDF associated
- if you are in the FFNS3, but you fit both IC and ICA, then you have the same contributions of FFNS4 available, so you need again `Nf4Ind`

andreab1997:
> PineAPPL only provides the different channels for the partonic cross section, knows nothing about the PDF.
> 
> But since we know what the PDF will be, and we know that some PDF elements will be just 0, then we can tell PineAPPL: "you see, I won't use these contributions, because the PDF will be 0, so you can even drop them". So you don't care as well if a PDF is intrinsic or perturbative, you only care if **it is different from 0** or it is not. But this you can infer for the number of active flavors (the patch you are fitting in), and the intrinsic objects you fit.
> 
> #### In practice...
> Actually, we fit in the 4 flavors patch, with IC, but we should not be interested in IC at all: the contribution is there for all charm channels, so you have to keep. But the full truth is that we use c- = 0 even if it is inconsistent: c+ is already small, c- will be even smaller, so it is only _slightly inconsistent_...
> 
> In any case, for this pass `Nf4Ind` and retain both charm contributions: neglecting c- will save very little memory, and it will be confusing the moment we decide to fit it. Do the consistent thing.

Ok perfect, Thanks for the exaplanation :)

src/pineko/theory_card.py
```diff
@@ -22,3 +22,36 @@ def load(theory_id):
...
+    act_flav = 6
+    mod = 'Ind'
+    if Q0 < match_scales['t']:
+        act_flav = 5
+    if Q0 < match_scales['b']:
+        act_flav = 4
+    if Q0 < match_scales['c']:
+        act_flav = 3
+        if ic:
+            act_flav += 1
```
AleCandido, line 55:
```suggestion
            act_flav += 1
            mod = 'Sym'
```

andreab1997: > In any case, for this pass Nf4Ind and retain both charm contributions: neglecting c- will save very little memory, and it will be confusing the moment we decide to fit it. Do the consistent thing

It seemed to me that you wanted to avoid "Sym" even if it is what we are doing at the moment. Have I misunderstood?

AleCandido: Slightly: I want to avoid `Sym` when we actually are in the 4 flavors (since it is inconsistent), and that is already what you are doing.

But if you see above in the list I told you that if you are parametrizing in the 3 flavors patch with IC (that is not what NNPDF is doing, at least for the time being), then you actually want `Sym`, because IC is unlocking only one degree of freedom (while the other would be behind something like ICA, that does not exist at the moment).

andreab1997: Ah okok now it is clear. :)

felixhekhorn (approved):

felixhekhorn:
> I am sorry for this. Problem is that I tried to run `pre-commit` but `black` was failing for some reason in my laptop (something related to not be able to import some of the click modules)

yes, that is [a known problem](https://github.com/psf/black/issues/2964)  and the solution is simply to update - and indeed v22.3.0 runs through

src/pineko/evolve.py
```diff
@@ -71,7 +71,13 @@ def write_operator_card(pineappl_grid, default_card, card_path):
 
 
 def evolve_grid(
-    pineappl_path, eko_path, fktable_path, max_as, max_al, comparison_pdf=None
+    pineappl_path,
+    eko_path,
+    fktable_path,
+    max_as,
+    max_al,
+    assumptions,
```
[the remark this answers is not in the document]
andreab1997: Ok done :)

felixhekhorn:
now you should be able to install :upside_down_face:

andreab1997:
> > I am sorry for this. Problem is that I tried to run `pre-commit` but `black` was failing for some reason in my laptop (something related to not be able to import some of the click modules)
> 
> yes, that is [a known problem](https://github.com/psf/black/issues/2964) and the solution is simply to update - and indeed v22.3.0 runs through

Oh yes, now is working. Thanks :)

felixhekhorn:
fine for me to merge! - @AleCandido ?

AleCandido:
I had a quick look and its mostly fine + I trust the two of you.

Just one question: I don't exactly understand what the CLI for `convolute` and `evolve` is exactly supposed to do (please, provide a CLI walkthrough in the online docs), but can you reassure me that regardless of their default, if they are used to compute FkTables they will respect the assumptions determined in the way discussed above?
(The answer might also be: "they are not used for the FkTables", just to make sure that a theory that should be optimized with `Nf4Ind` will not end up with `Nf6Ind`).

felixhekhorn:
> Just one question: I don't exactly understand what the CLI for `convolute` and `evolve` is exactly supposed to do (please, provide a CLI walkthrough in the online docs), but can you reassure me that regardless of their default, if they are used to compute FkTables they will respect the assumptions determined in the way discussed above? (The answer might also be: "they are not used for the FkTables", just to make sure that a theory that should be optimized with `Nf4Ind` will not end up with `Nf6Ind`).

yes, the CLI is not used for "typical FkTables" (meaning "theories"), but instead are just an entry point to a specific part of the code (say evolution or convolution) to which the user can/has to provide the settings; instead in the "theory" mode the program chooses the settings according to the theory runcard (which actually *is* available here - in contrast to "file based mode")

AleCandido (approved):

PR #24, opened 2022-06-13
Allow pineko to check if the grid contains SV 

andreab1997:
We want to allow pineko to check if scale variations are available in a grid if we are computing a scale-varied FK table.


src/pineko/check.py
```diff
@@ -41,3 +41,29 @@ def check_grid_and_eko_compatible(pineappl_grid, operators):
...
+    Parameters
+    ----------
+        pineappl_grid : pineappl.grid.Grid
+            grid
+        theory_card : dict
+            theory card
+    """
+    xir = theory_card["XIR"]
+    xif = theory_card["XIF"]
+    ftr = theory_card["fact_to_ren_scale_ratio"]
```
AleCandido, line 60: Don't propagate the `theory_card` too much, take these three elements as input of the function itself (it is more transparent, you can document it better, and use without the need of an actual `theory_card`).

src/pineko/check.py
```diff
@@ -41,3 +42,30 @@ def check_grid_and_eko_compatible(pineappl_grid, operators):
...
+        grid_path : pathlib.Path
+            path to grid
+        xir : float
+            log of renormalization scale ratio to central
+        xif : float
+            log of factorization scale ratio to central (scheme-C)
+        ftr : float
+            log of factorization scale ratio to central (scheme-B)
+    """
+    if xir == 1 and xif == 1 and ftr == 1:
```
felixhekhorn, line 62: the function should take a single argument, the grid and this if-condition goes in the calling function in the theory module

src/pineko/check.py
```diff
@@ -41,3 +42,30 @@ def check_grid_and_eko_compatible(pineappl_grid, operators):
...
+        return 0
+    pineappl_grid = pineappl.grid.Grid.read(grid_path)
+    order_list = [order.as_tuple() for order in pineappl_grid.orders()]
+    for order in order_list:
+        if order[-1] != 0 or order[-2] != 0:
+            return 0
```
felixhekhorn, line 68: also `return` is a valid python expression (without any value)

src/pineko/cli/check_scalevar.py
```diff
@@ -0,0 +1,19 @@
...
+import rich
+
+from .. import check, theory_card
+from ._base import command
+
+
+@command.command("check_scalevar")
+@click.argument("pineappl_path", metavar="PINEAPPL", type=click.Path(exists=True))
+@click.argument("theory_ID", metavar="ID", type=click.Path(exists=True))
+def subcommand(pineappl_path, theory_ID):
```
felixhekhorn, line 13: - let's keep this consistent with the rest of the "file-based" CLI: there is no theory - it just checks the availablibily
- if it fails print a nice message instead of an exception (as in the other check function)

src/pineko/cli/check_scalevar.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
```
felixhekhorn, line 1: let's move this to the cli/check module and rename the other function in there accordingly ...

src/pineko/check.py
felixhekhorn: also the check needs to use `isclose` (float remember ...)

src/pineko/check.py
```diff
@@ -41,3 +42,30 @@ def check_grid_and_eko_compatible(pineappl_grid, operators):
     # x-grid
     if not np.all(in1d(np.unique(operators["targetgrid"]), np.array(x_grid))):
         raise ValueError("x grid in pineappl grid and eko operator are NOT compatible!")
+
+
+def check_grid_contains_sv(grid_path, xir, xif, ftr):
+    """Raises a `ValueError if the theory_card asks for scale-variations but they are not
```
felixhekhorn, line 48: I wonder whether we should split into two functions: one for ren and one for fact - since they are required in different situations

felixhekhorn (changes requested):
- this PR should sit on top of #19 since even scheme C is not available in master
- I assume this will also trigger more minor issues (e.g. the numpy doc style ...)

src/pineko/check.py
AleCandido: (irrelevant:) a valid python *statement*

andreab1997:
> * #19

Should I rebase this on #19 then? @felixhekhorn

felixhekhorn:
> Should I rebase this on #19 then? @felixhekhorn

yes, please

andreab1997:
@felixhekhorn I have done but now there are some problems on stuffs that I have not touched. Do you know why?

felixhekhorn:
> @felixhekhorn I have done but now there are some problems on stuffs that I have not touched. Do you know why?

you mean here in the workflow? keep in mind that this only works together with https://github.com/N3PDF/pineappl/pull/138 so an unpublished version - check locally instead

src/pineko/check.py
```diff
@@ -65,3 +66,32 @@ def check_grid_and_eko_compatible(pineappl_grid, operators, xif):
...
         raise ValueError("x grid in pineappl grid and eko operator are NOT compatible!")
+
+def check_grid_contains_fact_sv(grid_path):
+    """Checks whether factorization scale-variations are available in the pineappl grid.
+    Parameters
+    ----------
+        grid_path : pathlib.Path
+            path to grid
+    """
+    pineappl_grid = pineappl.grid.Grid.read(grid_path)
```
felixhekhorn, line 77: please change the argument to be the object (instead of the path)

src/pineko/cli/check.py
```diff
@@ -28,3 +28,35 @@ def subcommand(pineappl_path, eko_path, xif):
         rich.print("[green]Success:[/] grids are compatible")
     except ValueError as e:
         rich.print("[red]Error:[/]", e)
+
+
+@command.command("check_scalevar")
+@click.argument("pineappl_path", metavar="PINEAPPL", type=click.Path(exists=True))
+@click.option(
```
felixhekhorn, line 35: if it is required make it an argument (instead of an option)

felixhekhorn (changes requested):

src/pineko/check.py
AleCandido: Yes, better to operate on objects, and load the grid only ones.
Otherwise you slow down everything, and it would make much harder to use for other functions (what should I do if I have an object? dump it and pass a path? much better the opposite: if it accepts object, and you have a path, just load it).

felixhekhorn:
Could you please do me a favour and also adjust the arguments of https://github.com/N3PDF/pineko/blob/b4dc401e4feef5cffc4b137081c246b66c8ea8d8/src/pineko/evolve.py#L79 such that in `TheoryBuilder.fk` we open the grid only once?

andreab1997:
> Could you please do me a favour and also adjust the arguments of
> 
> https://github.com/N3PDF/pineko/blob/b4dc401e4feef5cffc4b137081c246b66c8ea8d8/src/pineko/evolve.py#L79
> 
> such that in `TheoryBuilder.fk` we open the grid only once?

So you want that also `evolve_grid` takes the grid itself as argument and not the path, right?

felixhekhorn:
> > 
> 
> @felixhekhorn Ok but then we need to give up the fancy print with the grid path because we don't know that information anymore. Is it ok for you?
> 
> https://github.com/N3PDF/pineko/blob/e12c6482a32c7c8f902bfa2d2504e4bfe451fd19/src/pineko/evolve.py#L100

Mmm, actually, I like that one because it is telling me what is happening ... I guess we can lift it to the calling functions ... maybe on the "theory" side we can even adjust it a bit (since there the key ingredient is actually repeated 3 times ...)

andreab1997:
> > > 
> > 
> > 
> > @felixhekhorn Ok but then we need to give up the fancy print with the grid path because we don't know that information anymore. Is it ok for you?
> > https://github.com/N3PDF/pineko/blob/e12c6482a32c7c8f902bfa2d2504e4bfe451fd19/src/pineko/evolve.py#L100
> 
> Mmm, actually, I like that one because it is telling me what is happening ... I guess we can lift it to the calling functions ... maybe on the "theory" side we can even adjust it a bit (since there the key ingredient is actually repeated 3 times ...)

Ok this is done now. However I have two comments:

1.  I believe that `_x_grid, _pids, mur2_grid, _muf2_grid = pineappl_grid.axes()` is breaking something because I tried with **LHCB_DY_13TEV_DIMUON.pineappl.lz4** and It seems that its axes are only three and not four. Is this a problem of the grid (maybe it is just old or wrong) or a problem with the code?
2. Since we are now passing the grid itself to `evolve_grid` I believe that would be sensible to pass the eko itself as well. Do you agree? If yes maybe it is better to open another PR in order to keep this one as little as possible. Do you agree even on that?

cschwan:
@andreab1997 concerning 1. I believe that's the difference between PineAPPL master (three variables) and the [scale-variation branch](https://github.com/N3PDF/pineappl/pull/138) (four variables). I think you need the branch.

andreab1997:
> @andreab1997 concerning 1. I believe that's the difference between PineAPPL master (three variables) and the [scale-variation branch](https://github.com/N3PDF/pineappl/pull/138) (four variables). I think you need the branch.

Ok thanks. Then again the problem is with my version of pineappl

AleCandido:
> Ok thanks. Then again the problem is with my version of pineappl

You can install that branch cloning the repository and running `maturin install` from inside the Poetry environment (you enter in it with `poetry shell` run from inside Pineko repository).

felixhekhorn:
>     2. Since we are now passing the grid itself to `evolve_grid` I believe that would be sensible to pass the eko itself as well. Do you agree? If yes maybe it is better to open another PR in order to keep this one as little as possible. Do you agree even on that?

I agree on both

but please remember to activate pre-commit! (so please install now, run and commit again)

AleCandido:
> * 
> 
> * Since we are now passing the grid itself to `evolve_grid` I believe that would be sensible to pass the eko itself as well. Do you agree? If yes maybe it is better to open another PR in order to keep this one as little as possible. Do you agree even on that?

Once, in a distant future, I'll be able to completely N3PDF/eko#105, than the idea is to pass a callable, such that Q2 values can be loaded one at a time, not to blow up memory usage...

andreab1997:
> scale-variation branch

Thank you but I managed to do that in another way (due to usual lhapdf problem). However @cschwan since you marged the sv branch now the master branch should be ok as well right?

felixhekhorn:
> > scale-variation branch
> 
> Thank you but I managed to do that in another way (due to usual lhapdf problem). However @cschwan since you marged the sv branch now the master branch should be ok as well right?

yes, master should work now - still the procedure is the same as @AleCandido descibed (`maturin` etc.) (since it is not released)

cschwan:
@andreab1997 if you install it from git and not with pip, yes!

src/pineko/cli/convolute.py
```diff
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 """CLI entry point to convolution."""
 import click
+import pineappl as pine
```
felixhekhorn, line 4: let's not change the `pineappl` import but rather rename the conflicting argument

src/pineko/check.py
```diff
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 """Tools to check compatibility of EKO and grid."""
 import numpy as np
+import pineappl
```
felixhekhorn, line 4: is this still used?

felixhekhorn (changes requested):

src/pineko/check.py
AleCandido: Pylint should tell you :)

andreab1997:
What do we still need to merge this? @felixhekhorn @AleCandido

AleCandido:
I'll review in a moment

cschwan:
It's even contained in an official release now: <https://pypi.org/project/pineappl/>.

AleCandido:
> It's even contained in an official release now: https://pypi.org/project/pineappl/.

Thank you @cschwan, I thought one moment to block this waiting on release, but then I decided that we could even merge and upgrade in `main`.

Now there is no tradeoff any longer, I'm upgrading now :D

AleCandido:
@andreab1997 please try to rebase on `feature/sv`, I'm getting conflicts...

AleCandido:
> What do we still need to merge this? @felixhekhorn @AleCandido

@andreab1997 trivial observation: workflows are failing...

andreab1997:
> 703c316



> > What do we still need to merge this? @felixhekhorn @AleCandido
> 
> @andreab1997 trivial observation: workflows are failing...

I am taking care of this now

andreab1997:
> @andreab1997 please try to rebase on `feature/sv`, I'm getting conflicts...

I should have solved both the problems (the conflicts and the workflow failing)

src/pineko/cli/check.py
```diff
@@ -28,3 +28,33 @@ def subcommand(pineappl_path, eko_path, xif):
...
+
+
+@command.command("check_scalevar")
+@click.argument("pineappl_path", metavar="PINEAPPL", type=click.Path(exists=True))
+@click.argument(
+    "tocheck",
+    metavar="TOCHECK",
+    type=str,
+)
+def subcommand_sv(pineappl_path, tocheck):
```
AleCandido, line 40: This one I want to try to refactor myself, I'll publish a candidate as soon as I have time.

AleCandido (changes requested):
Few improvements, I'll give a tiny contribution as well and then we can merge.

debug.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+import eko
```
AleCandido, line 2: Do you actually want to merge this?
(might be fine, but maybe I'd not put top level)

In case you used for testing correctness of `pineko`, consider converting in a test (if makes use of `lhapdf` in an "isolated benchmark")

eko_debug.py
```diff
@@ -0,0 +1,47 @@
+# -*- coding: utf-8 -*-
+import eko
```
AleCandido, line 2: Same thing of `debug.py`.

src/pineko/check.py
```diff
@@ -65,3 +65,31 @@ def check_grid_and_eko_compatible(pineappl_grid, operators, xif):
...
+            Pineappl grid
+    """
+    order_list = [order.as_tuple() for order in grid.orders()]
+    for order in order_list:
+        if order[-1] != 0:
+            return
+    raise ValueError("Factorization scale variations are not available for this grid")
+
+
+def check_grid_contains_ren_sv(grid):
```
AleCandido, line 84: This and the former function I'd slightly improve the name:
- `check` is already the name of the file, and since it is already the namespace there is no need to repeat
- `grid` is already the object is acting on (so little to no need even here, I'd drop)
- `sv` is already implicit in `ren` and `fact`, no need to repeat

In the end they will be `contains_fact` and `contains_ren`. I'd like to avoid even the contain, with a more pythonic construction like
```py
"factoriziation" in grid
```
(`in` is actually the operator `__contains__`, so it's perfect), but I see plenty of problems. First you have to overwrite the default behavior of the `pinappl.grid.Grid.__contains__`, that is already defined somewhere else.

So fine by `contains_fact` and `contains_ren`.

yadism_debug.py
```diff
@@ -0,0 +1,72 @@
+# -*- coding: utf-8 -*-
+import numpy as np
```
AleCandido, line 2: Same thing of `debug.py`.

AleCandido (changes requested):

src/pineko/theory.py
```diff
@@ -346,6 +347,15 @@ def fk(self, name, grid_path, tcard, pdf):
         do_log = self.activate_logging(
             paths["logs"]["fk"], f"{self.theory_id}-{name}-{pdf}.log"
         )
+        # check if grid contains SV if theory is requesting them
+        xir = tcard["XIR"]
+        xif = tcard["XIF"]
```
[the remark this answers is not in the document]
felixhekhorn: yes, we have there "activateFact" (or similar name) in order to decide whether to compute the SV grids - however for `pineko` this is not sufficient since we need to collapse on the actual value - or what were you thinking?

debug.py
andreab1997: Actually I did not commit this (and not even coded). Maybe it is from @felixhekhorn

eko_debug.py
andreab1997: same here

debug.py
felixhekhorn: I will resolve this

yadism_debug.py
andreab1997: and same here

src/pineko/theory.py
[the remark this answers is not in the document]
AleCandido: Sorry, so it is just the comment to be referred to the code below.

Maybe I'd just move it: first `# check if sv are requested` then `# in case of sv, check they are available in the grid`.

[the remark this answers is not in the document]
andreab1997: you mean just the comments right? because the code is already doing that: first it checks if the theory is asking for scale-variations and then it checks whether they are available in the gird.

felixhekhorn:
I hope github is joking 
grafik

felixhekhorn:
this PR is next in line ... open issues are
- https://github.com/N3PDF/pineko/pull/24#discussion_r917791533
- https://github.com/N3PDF/pineko/pull/24#discussion_r917279856

andreab1997:
for this, should I add a couple of tests or we add them in a separate PR? because codecov is complaining....

AleCandido:
@andreab1997 can you test the new refactored version of `check` subcommand? 975b8fe0a194d122a99e69071bc3cc7b9da68061

andreab1997:
> @andreab1997 can you test the new refactored version of `check` subcommand, please? [975b8fe](https://github.com/N3PDF/pineko/commit/975b8fe0a194d122a99e69071bc3cc7b9da68061)

sure

andreab1997:
> @andreab1997 can you test the new refactored version of `check` subcommand, please? [975b8fe](https://github.com/N3PDF/pineko/commit/975b8fe0a194d122a99e69071bc3cc7b9da68061)

I tried that and it works. However I believe we should not add unit tests in this PR but rather in #25 mainly because we need actual grids to test the cli (as are provided by #25). So I think we can merge this now

felixhekhorn:
> I tried that and it works. However I believe we should not add unit tests in this PR but rather in #25 mainly because we need actual grids to test the cli (as are provided by #25). So I think we can merge this now

let's close this PR and also #25, #27, #28 - and have a restart from a new `main` branch; just try to keep track of some minor comments such as [this one](https://github.com/N3PDF/pineko/pull/24#commitcomment-78293005)

andreab1997:
> > I tried that and it works. However I believe we should not add unit tests in this PR but rather in #25 mainly because we need actual grids to test the cli (as are provided by #25). So I think we can merge this now
> 
> let's close this PR and also #25, #27, #28 - and have a restart from a new `main` branch; just try to keep track of some minor comments such as [this one](https://github.com/N3PDF/pineko/pull/24#commitcomment-78293005)

Yes, I am asking the review only for the test part, then I believe we can merge

felixhekhorn (approved):

andreab1997:
Can I merge this? @felixhekhorn @AleCandido

PR #25, opened 2022-06-15
Provide some tests

andreab1997:
We want to provide some unit test for pineko


felixhekhorn:
- Very good your start this project ;-) however, if you feel it is easier to do benchmarks before unit tests (something I would do) feel free to do so ...
- @AleCandido I guess we want to keep the split between unit tests in `tests` that have no external dependencies and benchmarks?
- benchmarks will be required (at some point) as they test the actual thing - specific here: the interplay with eko, lhapdf as pineko really is a glue ...

andreab1997:
> * Very good your start this project ;-) however, if you feel it is easier to do benchmarks before unit tests (something I would do) feel free to do so ...
> * @AleCandido I guess we want to keep the split between unit tests in `tests` that have no external dependencies and benchmarks?
> * benchmarks will be required (at some point) as they test the actual thing - specific here: the interplay with eko, lhapdf as pineko really is a glue ...

Yes, I agree that benchmarks is what we really need. However I would start doing unit tests to understand better how the code works before starting to write benchmarks.

AleCandido:
> Yes, I agree that benchmarks is what we really need. However I would start doing unit tests to understand better how the code works before starting to write benchmarks.

Benchmark will be useful at some point, and I want to keep the split.

Nevertheless, what we really need are unit tests: they are useful not only as a consistency check (like benchmarks), but also to lead development and refactoring (à la TDD, we have already seen them in action).

As usual, LHAPDF will make a bit more cumbersome to install dependencies, but maybe I can try to properly repackage it, and update on PyPI (and we might avoid a further container this way).

tests/test_files/pineko.toml
```diff
@@ -0,0 +1,13 @@
+[paths]
+# inputs
+ymldb = "data/yamldb"
```
felixhekhorn, line 3: you can even skip the `data` prefix, I'd say ...

andreab1997: Yes but I don't know at the moment if I will need some other files to be put in `test_files`. In that case I would prefer to keep them separate from the data

AleCandido:
> So I looked into the function and, for example, I did not understand why there is `a[1,-1]` in the place of `a`.

As cited in the docstring, the function is taken from here:
https://github.com/numpy/numpy/issues/7784#issuecomment-848036186

It is just a trick to retrieve the indices to compare to the elements, and e.g. in case `b[0] <= a[1]` you want index `0`, to later compare to `a[0]` and `a[1]`.

I'm pretty confident it is going to work for a sorted array, but it might even work for an unsorted one...

The only way to understand it is analyzing a few cases, i.e. testing it:
- tests don't have to be realistic, they are "unit tests", so they are testing the behavior of the function in isolation
- even if you are getting coverage, it might well be that the only way to be confident enough that a function is working, it's testing it with multiple inputs (so, despite coverage, one might be not enough)

andreab1997:
> > So I looked into the function and, for example, I did not understand why there is `a[1,-1]` in the place of `a`.
> 
> As cited in the docstring, the function is taken from here: [numpy/numpy#7784 (comment)](https://github.com/numpy/numpy/issues/7784#issuecomment-848036186)
> 
> It is just a trick to retrieve the indices to compare to the elements, and e.g. in case `b[0] <= a[1]` you want index `0`, to later compare to `a[0]` and `a[1]`.
> 
> I'm pretty confident it is going to work for a sorted array, but it might even work for an unsorted one...
> 
> The only way to understand it is analyzing a few cases, i.e. testing it:
> 
> * tests don't have to be realistic, they are "unit tests", so they are testing the behavior of the function in isolation
> * even if you are getting coverage, it might well be that the only way to be confident enough that a function is working, it's testing it with multiple inputs (so, despite coverage, one might be not enough)

Yes, that is the point. I have tested it with a not realistic input and it failed. Now I am going to understand why, however, even if it is working with realistic inputs, maybe we don't want to keep a function that fails in some cases.

AleCandido:
> Yes, that is the point. I have tested it with a not realistic input and it failed. Now I am going to understand why, however, even if it is working with realistic inputs, maybe we don't want to keep a function that fails in some cases.

Definitely we don't. So the correct behavior is keeping the tests, update the function :)
(e.g. you can sort the array, and store the inverse sorting to compare)

tests/conftests.py
```diff
@@ -0,0 +1,37 @@
...
+        try:
+            yield
+        finally:
+            os.chdir(prevdir)
+
+    return wrapped
+
+
+@pytest.fixture
+def lhapdf_path():
```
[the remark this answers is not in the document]
andreab1997: Yes but for the moment I am not using it. I will remove it at the end if not needed.

tests/test_files/logs/eko/208-HERA_CC_318GEV_EM_SIGMARED.log
```diff
@@ -0,0 +1,70 @@
+2022-06-16 11:19:58,734 pineko.theory/INFO: Start computation of HERA_CC_318GEV_EM_SIGMARED
```
AleCandido, line 1: Are you sure that this had to be commited? @andreab1997

AleCandido: Same for the others, actually...

Are you sure that this commit is worth?

tests/conftests.py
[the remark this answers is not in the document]
AleCandido: If needed, you should move in benchmarks, if not needed, you should remove. In any case this is not going to remain here :)

tests/test_files/logs/eko/208-HERA_CC_318GEV_EM_SIGMARED.log
andreab1997: No sorry, I can remove the logs. But i need the directories

tests/conftests.py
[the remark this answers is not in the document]
andreab1997: Yes, I removed it :)

AleCandido:
In general, unit tests should test things in isolation. So, please, do your best in order to limit the amount of non-source files you are using in test.

In particular: do your best to remove completely the `test_files` folder. It is awful to have all these files committed in the repo, and the more files you have, the more things you have to keep updated: if the programs generating them will change, it might change the specific format and content of the files.

The best pattern would be to make fixtures, that are generating required files at runtime, in temporary folders.

tests/test_files/logs/eko/208-HERA_CC_318GEV_EM_SIGMARED.log
AleCandido: Make a fixture that creates the tree

andreab1997:
> In general, unit tests should test things in isolation. So, please, do your best in order to limit the amount of non-source files you are using in test.
> 
> In particular: do your best to remove completely the `test_files` folder. It is awful to have all these files committed in the repo, and the more files you have, the more things you have to keep updated: if the programs generating them will change, it might change the specific format and content of the files.
> 
> The best pattern would be to make fixtures, that are generating required files at runtime, in temporary folders (if really needed).

Yes, I would have liked to do everything at runtime. However I need at least one eko to test and it takes forever to compute it at runtime, even for small datasets

AleCandido:
> Yes, I would have liked to do everything at runtime. However I need at least one eko to test and it takes forever to compute it at runtime, even for small datasets

Even computing it on a 5 points grid?

andreab1997:
> > Yes, I would have liked to do everything at runtime. However I need at least one eko to test and it takes forever to compute it at runtime, even for small datasets
> 
> Even computing it on a 5 points x-grid?
> 
> (However, in case there is no way to get a small one, one EKO is fine, but then use only that one, and not the full folders hierarchy)

Yes, of course I will take care of the folders. The fact is that I need an actual eko (with the correct grid) because I want also to test the checks.

felixhekhorn:
> Yes we could but the problem is that then I cannot test the checks because the eko would not be the real one

but you don't have a real PDF either (since you're not having lhapdf) - in fact, the only PDF you could use in a unit test is toyPDF

AleCandido (changes requested):
I'm happy to accept your tests: I don't have time to check all of them in details, but I trust you, and I'm happy they are passing.

The only thing I request to you is the following: move `test_files` out of tests (and all related tests), and of course rename it.
You can go for something like "isolated benchmarks" (isobench) as we currently have in EKO.

But please, remove any test that involves non-`pip` dependencies (essentially LHAPDF, since PineAPPL is fine instead) from tests. Tests should be unit tests, and they should work even for the very basic installation:
```
pip install pineko
```
in a fresh clean environment.

AleCandido:
Please, also merge/rebased on `main`, right after merging #28

felixhekhorn:
https://github.com/N3PDF/pineko/pull/25/commits/688e4cc240d1223db192a17ed60dc022b2039fbe hm? are you not running pre-commit? (or did you ignore intentionally?)

andreab1997:
> [688e4cc](https://github.com/N3PDF/pineko/commit/688e4cc240d1223db192a17ed60dc022b2039fbe) hm? are you not running pre-commit? (or did you ignore intentionally?)

I am running pre-commit but for some reason black did not work on that specific file and I had an error from the pre-commit action of github. So I run black and it solved the problem

andreab1997:
@AleCandido @felixhekhorn Since this is getting old, I would try to merge this as soon as possible even if it is not completed yet. I have tried to split benchmarks from unit tests and to keep only a limited number of files, as you requested :)

AleCandido:
I agree with you, we should try to merge soon. The only thing I don't like are the 132 files (I'd like more to generate/donwload them)...

By the way, if you provide a small test for `evolve.evolve_grid` you'll get back the green tick :)

AleCandido:
> > 
> 
> Yes, I would like to avoid them too but I don't know how. Of course I could try to download them from dom but it takes time (not a lot but still) and dom is not really the official place for the grids and the fks.

Indeed, I agree. We should not use `dom`, also because it is not public.

> At the end most of the files come from the pdf, but I can just avoid to put there all the members (which is something I should have noticed before :) ).

I was actually thinking about the PDF set, and to download it from LHAPDF, that is official.
In any case, once you find the best set, we can just remove everything else also from `git` history, and reduce the size of the repo (not a pleasure in general, but possible and useful).

andreab1997:
> > > 
> > 
> > 
> > Yes, I would like to avoid them too but I don't know how. Of course I could try to download them from dom but it takes time (not a lot but still) and dom is not really the official place for the grids and the fks.
> 
> Indeed, I agree. We should not use `dom`, also because it is not public.
> 
> > At the end most of the files come from the pdf, but I can just avoid to put there all the members (which is something I should have noticed before :) ).
> 
> I was actually thinking about the PDF set, and to download it from LHAPDF, that is official. In any case, once you find the best set, we can just remove everything else also from `git` history, and reduce the size of the repo (not a pleasure in general, but possible and useful).

Actually now that I have removed all the members but one, the pdf set is just two files so I don't know if it is worth to download them.

andreab1997:
> > > 
> > 
> > 
> > Yes, I would like to avoid them too but I don't know how. Of course I could try to download them from dom but it takes time (not a lot but still) and dom is not really the official place for the grids and the fks.
> 
> Indeed, I agree. We should not use `dom`, also because it is not public.
> 
> > At the end most of the files come from the pdf, but I can just avoid to put there all the members (which is something I should have noticed before :) ).
> 
> I was actually thinking about the PDF set, and to download it from LHAPDF, that is official. In any case, once you find the best set, we can just remove everything else also from `git` history, and reduce the size of the repo (not a pleasure in general, but possible and useful).

And now we have again the green tick

AleCandido:
> > 
> 
> Actually now that I have removed all the members but one, the pdf set is just two files so I don't know if it is worth to download them.

At this point is just the same of the grids: either we remove them all, or we keep them.
For the time being they are sufficiently small, so I'd rather do the easiest thing and keep them where they are, to merge sooner this PR. If ever we'll reconsider benchmarks, we can rediscuss.

AleCandido (changes requested):
I'd say it is almost perfect.

If you can quickly fix the few things, I would merge right after.

AleCandido (approved):

AleCandido:
The benchmark workflow you can improve in this PR, or also directly on `main` (in case you're going to do it).

AleCandido:
So, the cache did work:
image

But it is not used at the moment:

AleCandido:
Ok, you did all well. But, at this point, I believe also the other workflow is not perfect.

Revert 26d5b1e9c2d9ead1796d844257cb20452b1cde6b. I considered telling you before, but then you were doing already fine, so there was no reason: you can't do the same, because you are inside another container, I believe is more complicated.

andreab1997:
> Ok, you did all well. But, at this point, I believe also the other workflow is not perfect.
> 
> Revert [26d5b1e](https://github.com/N3PDF/pineko/commit/26d5b1e9c2d9ead1796d844257cb20452b1cde6b). I considered telling you before, but then you were doing already fine, so there was no reason: you can't do the same, because you are inside another container, I believe is more complicated.

Ah ok, I was beginning to suspect something like that :) 
So most likely I can just revert the last three commits and go back after https://github.com/N3PDF/pineko/pull/25/commits/5e0fd126ca4ada0d9686aed333ab570dfc1d2f1a

.github/workflows/bench.yml
```diff
@@ -0,0 +1,60 @@
...
+        run: |
+          # same poetry env
+          PIP="$(head -n1 $(which poetry) | cut -c 3-) -m pip"
+          ${PIP} install poetry-dynamic-versioning
+      - name: Install dependencies
+        run: poetry install --no-interaction --no-root
+        if: steps.cache.outputs.cache-hit != 'true'
+      - name: Install project
+        run: |
+          poetry install --no-interaction ${{ inputs.poetry-extras }}
```
AleCandido, line 42:
```suggestion
      - name: Install dependencies
        run: poetry install --no-interaction --no-root ${{ inputs.poetry-extras }}
        if: steps.cache.outputs.cache-hit != 'true'
      - name: Install project
        run: |
          poetry install --no-interaction --only-root
```

AleCandido: Explanation: the extras should go together with the other dependencies.

Moreover, if everything is already fine and skipped, the extra step should only install the project.

AleCandido (changes requested):
Try to install project only (no dependencies), when cache is used.

AleCandido:
ec5f792f6aeebf65c2f0f2a3f76e4af6b80db61b did not work: for unknown reasons `pylint` was missing, while it should have been in the cached environment (since `poetry install` by default should install dev-deps as well).

The only guess is that since Poetry 1.2.0 dev is just another group, so maybe we should specify `--with dev-dependencies`. But it is working also without, since `dev-dependencies` is (by default) a non-optional group.

So, there is a weird interplay with cache. I would suggest merging and forgetting, the status is the same of `N3PDF/workflows`, at some point we will re-investigate for both.

andreab1997:
> [ec5f792](https://github.com/N3PDF/pineko/commit/ec5f792f6aeebf65c2f0f2a3f76e4af6b80db61b) did not work: for unknown reasons `pylint` was missing, while it should have been in the cached environment (since `poetry install` by default should install dev-deps as well).
> 
> The only guess is that since Poetry 1.2.0 dev is just another group, so maybe we should specify `--with dev-dependencies`. But it is working also without, since `dev-dependencies` is (by default) a non-optional group.
> 
> So, there is a weird interplay with cache. I would suggest merging and forgetting, the status is the same of `N3PDF/workflows`, at some point we will re-investigate for both.

Ok but I need to ask you again the review

felixhekhorn (changes requested):
since we're going for `pytest`, let's use it! i.e.
- move `utils` in `conftest` and make `lhapdf_path` a fixture (as we do in `eko`)
- make `test_files` + `test_pdf` a fixture
- I also suggest a fixture `def grid_path(id)` and similar
- a minor style comment: let's move file names from suffix to prefix as we do for tests and as in eko

then let's merge!

benchmarks/bench_theory.py
```diff
@@ -0,0 +1,149 @@
+# -*- coding: utf-8 -*-
+import os
+import pathlib
+
+import pineko
+
+theory_obj = pineko.theory.TheoryBuilder(208, ["LHCB_Z_13TEV_DIMUON"])
+theory_obj_Hera = pineko.theory.TheoryBuilder(208, ["HERACOMBCCEM"])
```
AleCandido, line 8: Respect naming convention, all lower case:
```suggestion
theory_obj_hera = pineko.theory.TheoryBuilder(208, ["HERACOMBCCEM"])
```

AleCandido (changes requested):

benchmarks/bench_theory.py
AleCandido: And the rest accordingly

AleCandido:
> *  I also suggest a fixture `def grid_path(id)` and similar

This I'm not sure is really useful: they are a bunch of functions only used a few times each. However, I'm not strongly opinionated, do whatever you prefer, but don't spend too much time, it is already good enough.

andreab1997:
> > * I also suggest a fixture `def grid_path(id)` and similar
> 
> This I'm not sure is really useful: they are a bunch of functions only used a few times each. However, I'm not strongly opinionated, do whatever you prefer, but don't spend too much time, it is already good enough.

Actually I was thinking about the same thing. I believe that the lhapdf_path fixture is useful and the test_pdf and test_files ones are useful as well but a grid_path fixture I don't think it would be. So for me we can merge, do you agree?

AleCandido:
> So for me we can merge, do you agree?

Almost, find and replace: `s/theory_obj_Hera/theory_obj_hera/g` in the whole `bench_theory.py`. You did only for the definition, and it is obviously breaking the rest...

andreab1997:
> > So for me we can merge, do you agree?
> 
> Almost, find and replace: `s/theory_obj_Hera/theory_obj_hera/g` in the whole `bench_theory.py`. You did only for the definition, and it is obviously breaking the rest...

Yes, I know :)

andreab1997:
Actually now docker is failing for some reason...

AleCandido:
> Actually now docker is failing for some reason...

Network/server issue, I believe. Let's retry with the new commit, otherwise in a while.

andreab1997:
Now that the workflow is running correctly, can we merge this? @AleCandido @felixhekhorn

benchmarks/bench_theory_card.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+import pathlib
+
+import pineko
+
+test_files = pathlib.Path(__file__).parents[0] / "data_files/"
```
felixhekhorn, line 6: remove in favour of fixture

felixhekhorn (changes requested):

Issue #26, opened 2022-06-21
FONLL-B DIS FK table need different PTO


felixhekhorn:
I'm sorry to say again, but NLO FK tables are still wrong (only FK tables this time though - so it's not expensive):
https://github.com/N3PDF/pineko/blob/e12c6482a32c7c8f902bfa2d2504e4bfe451fd19/src/pineko/theory.py#L325

this is wrong in the case of NLO (so `PTO=1`) and FONLL-B because FONLL-B also contains bits at $O(a_s^2)$ which will be neglected with this statement ...

the solution is
- find out (from the theory card) whether we're in FONLL-B
- check whether we're looking at a DIS grid (from the lumi, I guess - we can cross check with https://github.com/NNPDF/nnpdf/pull/1529)
- eventually do a correction of `max_as`

felixhekhorn:
- if this is done - close it
- did you compare you're `is_dis` determination against https://github.com/NNPDF/nnpdf/pull/1578 ?
- actually https://github.com/N3PDF/pineko/blob/aa1a068edffc75d8ee1a4ed8e5c44fd8630963e1/src/pineko/check.py#L87 it might be a good idea to not rely on the fact that the lepton is always the second particle in the lumi

andreab1997:
Should we close this? @felixhekhorn @AleCandido

PR #27, opened 2022-07-05
Change max_as during FONLL-B DIS fk calculation

andreab1997:
We need pineko to change the `max_as` value from 1 to 2 when computing FONLL-B DIS fktables. 
Solves #26


src/pineko/check.py
```diff
@@ -43,19 +43,29 @@ def check_grid_and_eko_compatible(pineappl_grid, operators):
...
 
     Returns
     -------
             : bool
             true if the fktable is a FONLL-B DIS fktable
     """
-    if fns == "FONLL-B":
+    import ipdb
+
+    ipdb.set_trace()
```
AleCandido, line 63: if possible, avoid to commit debugging instructions (`pre-commit` should actually warns you about, are you using it?)

andreab1997: Actually in this case yes, I have used that.

AleCandido: Which one of the two?
1. you used and it didn't warn
2. you had a strong need to commit `ipdb`

Because this `pre-commit` action is actually checking the presence of leftover debugging instructions:
https://github.com/N3PDF/pineko/blob/main/.pre-commit-config.yaml#L12

In case of 2., please avoid as much as possible!

andreab1997: I used and I didn't get the warning. But ok fine, maybe I am just wrong and I have actually forgot to run pre-commit

felixhekhorn: well, the whole idea of pre-commit is to *not* run it yourself, but to let git do it ...

src/pineko/check.py
```diff
@@ -41,3 +41,28 @@ def check_grid_and_eko_compatible(pineappl_grid, operators):
     # x-grid
     if not np.all(in1d(np.unique(operators["targetgrid"]), np.array(x_grid))):
         raise ValueError("x grid in pineappl grid and eko operator are NOT compatible!")
+
+
+def check_fonll_b(fns, lumi):
```
felixhekhorn, line 46:
```suggestion
def is_fonll_b(fns, lumi):
```

andreab1997:
same here: should I add tests here?

AleCandido:
> same here: should I add tests here?

I'm not connecting to what "same" is referred, but definitely **yes**. Add as many *relevant* tests as you can!

andreab1997:
> > same here: should I add tests here?
> 
> I'm not connecting to what "same" is referred, but definitely **yes**. Add as many _relevant_ tests as you can!

yes sorry, I forgot to tag the other PR. Same of #24

andreab1997:
again, I am asking the review only for the test

felixhekhorn (approved):

src/pineko/check.py
```diff
@@ -66,6 +66,29 @@ def check_grid_and_eko_compatible(pineappl_grid, operators, xif):
...
+    Returns
+    -------
+            : bool
+            true if the fktable is a FONLL-B DIS fktable
+    """
+    isDIS = True
+    for lists in lumi:
+        for el in lists:
+            if el[1] != 11:
+                isDIS = False
```
AleCandido, line 88: Instead of defining `isDIS` (even wrong case convention), just early return

AleCandido:
```suggestion
                return False
```

AleCandido: Of course remove `isDIS` even above and below, and add a comment explaining why you are early returning.

AleCandido (changes requested):

src/pineko/theory.py
```diff
@@ -364,6 +364,13 @@ def fk(self, name, grid_path, tcard, pdf):
                 rich.print(f"Skipping existing FK Table {fk_filename}")
                 return
         max_as = 1 + int(tcard["PTO"])
+        # Check if we are computing FONLL-B fktable and eventually change max_as
+        pine_grid = pineappl.grid.Grid.read(grid_path)
```
AleCandido, line 368: If you wish `grid` is even enough:
```suggestion
        grid = pineappl.grid.Grid.read(grid_path)
```

Minimal, and non-ambiguous (a grid is by definition a PineAPPL object)

src/pineko/theory.py
```diff
@@ -364,6 +364,13 @@ def fk(self, name, grid_path, tcard, pdf):
...
+        if check.is_fonll_b(
+            tcard["FNS"],
+            pine_grid.lumi(),
```
AleCandido, line 371:
```suggestion
            grid.lumi(),
```

src/pineko/check.py
```diff
@@ -66,6 +66,29 @@ def check_grid_and_eko_compatible(pineappl_grid, operators, xif):
...
+    if fns == "FONLL-B" and isDIS:
```
AleCandido, line 89:
```suggestion
    if fns == "FONLL-B":
```

src/pineko/check.py
```diff
@@ -66,6 +66,29 @@ def check_grid_and_eko_compatible(pineappl_grid, operators, xif):
...
+            flavor number scheme (from the theory card)
+        lumi : list(list(tuple))
+            luminosity info
+
+    Returns
+    -------
+            : bool
+            true if the fktable is a FONLL-B DIS fktable
+    """
+    isDIS = True
```
AleCandido, line 84:
```suggestion
```

AleCandido (changes requested):

src/pineko/check.py
```diff
@@ -67,6 +67,31 @@ def check_grid_and_eko_compatible(pineappl_grid, operators, xif):
...
+            luminosity info
+
+    Returns
+    -------
+            : bool
+            true if the fktable is a FONLL-B DIS fktable
+    """
+    for lists in lumi:
+        for el in lists:
+            if el[1] != 11:
```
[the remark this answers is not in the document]
andreab1997: Oh yes sure. Just to understand, are there grids with other leptons at the moment?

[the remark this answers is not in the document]
AleCandido: Yes, for CC it's very relevant (because they decide which W boson is, and so which flavors/antiflavors are probed).

They are specified through a member of the observable card:
```yaml
projectileDIS: ["electron", "positron", "neutrino", "antineutrino"]
```

AleCandido:
For me now would be even fine, but tests are failing, I guess because I forgot a `not` in the `if` condition:

```py
if not 10 < abs(el[1]) < 17:
``'

AleCandido:
If tests will pass, you can merge @andreab1997

PR #28, opened 2022-07-08
Changing the eko argument in evolve_grid

andreab1997:
We want to pass the eko object rather than the path to evolve_grid

### DIFF
--- a/src/pineko/theory.py
+++ b/src/pineko/theory.py
@@ -400,8 +400,8 @@ class TheoryBuilder:
             f"with max_as={max_as}, max_al={max_al}, xir={xir}, xif={xif}",
         )
         _grid, _fk, comparison = evolve.evolve_grid(
-            operators,
             grid,
+            operators,
             fk_filename,
             max_as,
             max_al,


andreab1997:
@felixhekhorn I am not sure but maybe it is better to merge this to #24 rather than to main. Am I wrong?

src/pineko/cli/convolute.py
```diff
@@ -2,6 +2,7 @@
 """CLI entry point to convolution."""
 import click
 import pineappl as pine
+import eko as ek
```
felixhekhorn, line 5: again, let's not change the import name

src/pineko/theory.py
```diff
@@ -369,10 +370,9 @@ def fk(self, name, grid_path, tcard, pdf):
         astrong = sc.StrongCoupling.from_dict(tcard)
         # ocard = self.load_operator_card(name)
         # q2_grid = ocard["Q2grid"]
+        #loading ekos
```
felixhekhorn, line 373: you're still not running pre-commit :upside_down_face:

andreab1997: This time I was sure 😩

AleCandido:
```
pre-commit install
```
in the repository.

Right after, you can check the content of `.git/hooks/pre-commit`. It should be clear from the code that you installed it (otherwise the file will contain only comments)

felixhekhorn: https://github.com/N3PDF/pineko/pull/27#discussion_r916710466 - don't worry it is just a minor thing, which however is useful

andreab1997: Ok thanks this is what I was missing

src/pineko/theory.py
```diff
@@ -11,6 +11,7 @@
 
 import eko
 import numpy as np
+import pineappl
```
AleCandido, line 14: Here you're adding this `import`, but I don't see it used anywhere,

Am I missing something?

andreab1997: I am using pineappl here 
https://github.com/N3PDF/pineko/blob/842a31793c9e5b71498dc6c748404f4abffee1e2/src/pineko/theory.py#L355

AleCandido: Actually, I'm rather surprised Pylint didn't notify

andreab1997: yes because this branch originates from sv_check where that change happened

AleCandido: Can you (at least quickly) investigate why that branch passes all checks, even if we know it to be invalid?

andreab1997: But is not invalid, why should it be?

AleCandido: > But is not invalid, why should it be?

Either you need to import, or you don't need. However, @felixhekhorn explained the mystery: it was, it has simply been fixed in both branches afterwards (so the `import pineappl` appears in the "Files Changed" of this PR, and it is needed even if there is no new `pineappl.<something>` anywhere)

felixhekhorn:
Can this then be merged? or we wait for #24 to be merged first?

AleCandido (approved):

AleCandido:
Before rebase, please

andreab1997:
> Before rebase, please

on which branch?

felixhekhorn:
> Before rebase, please

as you know, I don't like this "rewriting history" business :upside_down_face: so I started to reread the reason why people are doing this and the ("simple") rule seems to be ["You can get the best of both worlds: rebase local changes before pushing to clean up your work, but never rebase anything that you’ve pushed somewhere."](https://git-scm.com/book/en/v2/Git-Branching-Rebasing) aka ["only rebase branches you, and only you, are working on."](https://medium.com/osedea/git-rebase-powerful-command-507bbac4a234)

so I'd say there should not be a rebase visible on github - hence you can just merge `sv_check` here (and before merge `master` on `sv_check`)

AleCandido:
> to main I guess

On the one your conflicting with, i.e. the one on which the PR is based:

image

andreab1997:
> Can this then be merged? or we wait for #24 to be merged first?

I believe we can merge now, I solved the conflicts

AleCandido:
> so I'd say there should not be a rebase visible on github - hence you can just merge `sv_check` here (and before merge `master` on `sv_check`)

I'm not saying you should do what I wish, if you do not agree with me, do whatever you wish. But I'll keep doing as well :)

The points are the following:
1. linear histories are simpler: when you want to explore when changes entered, there is a before and an after; at the time you merge your `main` was arrived at some point, so the commit you're inserting are arriving afterwards - this is simple, we solve only at merge time
2. if you do locally, then you never do: we always push everything, just to keep other collaborators updated, so it's always remote

Baseline: I will keep rebasing, feel free to do whatever you think it's best :)

andreab1997:
So at the end I am a bit confused, what do we want to do with this?

AleCandido:
> So at the end I am a bit confused, what do we want to do with this?

I have an opinion, @felixhekhorn has the opposite. Pick your favorite for your own reasons :)

(However, either merge or rebased, but something has to be done)

andreab1997:
Is everything ok with this?

felixhekhorn (approved):

PR #30, opened 2022-07-14
Check also first element in lumi for is_fonllb

andreab1997:
we want that `check.is_fonllb` checks also the first entry in  each `grid.lumi` element. 
Solve #26

### DIFF
--- a/src/pineko/check.py
+++ b/src/pineko/check.py
@@ -84,7 +84,7 @@ def is_fonll_b(fns, lumi):
     """
     for lists in lumi:
         for el in lists:
-            if not (10 < abs(el[1]) < 17):
+            if (not (10 < abs(el[0]) < 17)) and (not (10 < abs(el[1]) < 17)):
                 # in this case we are sure it is not DIS so for sure it is not FONLL-B
                 return False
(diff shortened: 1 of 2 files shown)


felixhekhorn (approved):

src/pineko/check.py
```diff
@@ -84,7 +84,7 @@ def is_fonll_b(fns, lumi):
     """
     for lists in lumi:
         for el in lists:
-            if not (10 < abs(el[1]) < 17):
+            if (not (10 < abs(el[0]) < 17)) and (not (10 < abs(el[1]) < 17)):
```
AleCandido, line 87: Since I'm writing on this PR in any case: instead of repeating twice the condition, make a brief `is_lepton` function (even in this function scope, or top level in the file, as you wish).

AleCandido:
@andreab1997 I know this PR has already been merged, but it was more convenient to write here, since it is next to the code.

andreab1997:
> @andreab1997 I know this PR has already been merged, but it was more convenient to write here, since it is next to the code.

Yes, sorry to have merged it without waiting you: I thought it was trivial :)

PR #31, opened 2022-07-15
Splitting the func islepton and updating test

andreab1997:
This is just a minor change to #30 due to @AleCandido suggestions.

### DIFF
--- a/tests/test_check.py
+++ b/tests/test_check.py
@@ -55,10 +55,10 @@ def test_is_fonll_b():
     fns = "FONLL-B"
     lumi = [[(1, 11, 1.0), (3, 11, 5.0)]]
     assert pineko.check.is_fonll_b(fns, lumi) is True
-    lumi.append([(-12, 1, 2.0), (-13, 1, 5.0)])
-    assert pineko.check.is_fonll_b(fns, lumi) is True
-    lumi.append([(1, 1, 4.0), (2, 11, 3.0)])
-    assert pineko.check.is_fonll_b(fns, lumi) is False
-    lumi.pop(-1)
+    new_lumi = [[(-12, 1, 2.0), (-13, 1, 5.0)]]
+    assert pineko.check.is_fonll_b(fns, new_lumi) is True
+    new_lumi.append([(1, 1, 4.0), (2, 11, 3.0)])
+    assert pineko.check.is_fonll_b(fns, new_lumi) is False
+    new_lumi.pop(-1)
     fns = "FONLL-C"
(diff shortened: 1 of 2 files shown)


tests/test_check.py
```diff
@@ -53,9 +53,11 @@ def test_contains_ren():
 
 def test_is_fonll_b():
     fns = "FONLL-B"
-    lumi = [[(1, 11, 3, 4), (3, 11, 5, 6)], [(9, 11, 0, 3), (-12, 3, -2, -1)]]
+    lumi = [[(1, 11, 1.0), (3, 11, 5.0)]]
     assert pineko.check.is_fonll_b(fns, lumi) is True
-    lumi.append([(1, 11, 2, 3), (2, 4, 5, 6)])
+    lumi.append([(-12, 1, 2.0), (-13, 1, 5.0)])
```
felixhekhorn, line 58: don't use append (`lumi` would still become crazy) - construct a new one

tests/test_check.py
```diff
@@ -53,10 +53,12 @@ def test_contains_ren():
...
     fns = "FONLL-B"
-    lumi = [[(1, 11, 3, 4), (3, 11, 5, 6)], [(9, 11, 0, 3), (-12, 3, -2, -1)]]
+    lumi = [[(1, 11, 1.0), (3, 11, 5.0)]]
     assert pineko.check.is_fonll_b(fns, lumi) is True
-    lumi.append([(1, 11, 2, 3), (2, 4, 5, 6)])
-    assert pineko.check.is_fonll_b(fns, lumi) is False
-    lumi.pop(-1)
+    new_lumi = [[(-12, 1, 2.0), (-13, 1, 5.0)]]
+    assert pineko.check.is_fonll_b(fns, new_lumi) is True
+    new_lumi.append([(1, 1, 4.0), (2, 11, 3.0)])
```
felixhekhorn, line 60: Same as before: let's make sane tests (such that if we ever implement a sanity check they would still pass ...); e.g.
- rename `lumi` -> `lumi_second`
- rename `new_lumi` -> `lumi_first`
- you can keep the crazy case, but just create a dedicated variable which has the thing clearly written, e.g. `lumi_crazy`
- and then check in FONLL-C *all* variables (all returning `False`)

felixhekhorn (changes requested):

felixhekhorn (approved):

andreab1997:
@AleCandido Can this be merged?

AleCandido (approved):

Issue #32, opened 2022-07-15
SV check does not take into account the PTO


andreab1997:
Currently our SV check does not take into account the fact that renormalization sv only start at NLO where NLO actually means next with respect to the first non-zero order of the process. 
Therefore, it does not allow the computation of a sv Fktable for DIS when PTO=1 (which in DIS case is the first non-zero order).

andreab1997:
Can this be closed?

felixhekhorn:
decide yourself :upside_down_face:

andreab1997:
> decide yourself 🙃

Then let's wait to have the full check. At least we won't forget :)

felixhekhorn:
> Can this be closed?

@andreab1997 now?

PR #33, opened 2022-07-15
Fixing sv check

andreab1997:
Solve #32

### DIFF
--- a/src/pineko/check.py
+++ b/src/pineko/check.py
@@ -101,9 +101,19 @@ def contains_fact(grid):
             Pineappl grid
     """
     order_list = [order.as_tuple() for order in grid.orders()]
+    as_orders = []
     for order in order_list:
-        if order[-1] != 0:
-            return
+        as_orders.append(order[0])
+    min_as = min(as_orders)
+    order_is_present = False
+    for order in order_list:
+        # fact sv starts at NLO with respect to the first non zero order
+        if order[0] == min_as + 1:
+            order_is_present = True
+            if order[-1] != 0:
+                return
+    if not order_is_present:
+        return
     raise ValueError("Factorization scale variations are not available for this grid")
 
 
@@ -116,7 +126,17 @@ def contains_ren(grid):
             Pineappl grid
     """
     order_list = [order.as_tuple() for order in grid.orders()]
+    as_orders = []
     for order in order_list:
-        if order[-2] != 0:
-            return
+        as_orders.append(order[0])
+    # ren sv starts one order after the first order with as
+    min_as = 1 if min(as_orders) == 0 else min(as_order)
+    order_is_present = False
+    for order in order_list:
+        if order[0] == min_as + 1:
+            order_is_present = True
+            if order[-2] != 0:
+                return
+    if not order_is_present:
+        return
     raise ValueError("Renormalization scale variations are not available for this grid")
--- a/src/pineko/cli/check.py
+++ b/src/pineko/cli/check.py
@@ -46,6 +46,7 @@ def sub_compatibility(grid_path, operator_path, xif):
 def sub_scvar(grid_path, scale):
     """Check if PineAPPL grid contains requested scale variations."""
     grid = pineappl.grid.Grid.read(grid_path)
+    grid.optimize()
     if scale == "ren":
         try:
             check.contains_ren(grid)
--- a/src/pineko/theory.py
+++ b/src/pineko/theory.py
@@ -353,6 +353,8 @@ class TheoryBuilder:
         ftr = tcard["fact_to_ren_scale_ratio"]
         # loading grid
         grid = pineappl.grid.Grid.read(grid_path)
+        # remove zero subgrid
+        grid.optimize()
         if not np.isclose(xir, 1.0):
             check.contains_ren(grid)
         if not (np.isclose(xif, 1.0) and np.isclose(ftr, 1.0)):
--- a/tests/test_check.py
+++ b/tests/test_check.py
@@ -27,8 +27,8 @@ class Order:
 
 def test_contains_fact():
     first_order = Order((0, 0, 0, 0))
-    second_order = Order((1, 1, 0, 0))
-    third_order = Order((0, 0, 1, 1))
+    second_order = Order((1, 0, 0, 0))
+    third_order = Order((1, 0, 0, 1))
     order_list = [first_order, second_order, third_order]
     mygrid = Fake_grid(order_list)
     assert pineko.check.contains_fact(mygrid) is None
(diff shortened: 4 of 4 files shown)


felixhekhorn (changes requested):
as we said: you need to take PTO from the outside and I think you should use `create_mask`

andreab1997:
> as we said: you need to take PTO from the outside and I think you should use `create_mask`

Ok so we want to check if the sv are available for the order we are asking, not if they are just available in the grid.

felixhekhorn:
> > as we said: you need to take PTO from the outside and I think you should use `create_mask`
> 
> Ok so we want to check if the sv are available for the order we are asking, not if they are just available in the grid.

I would say that depends a bit on what we actually want from that function, i.e. how strict we want to be ...
- we could just do a simple "is LO? then skip else proceed as before"
- or we do a more strict check: "I'm asking for NLO - check that *all required* ingredients are there"

andreab1997:
> > > as we said: you need to take PTO from the outside and I think you should use `create_mask`
> > 
> > 
> > Ok so we want to check if the sv are available for the order we are asking, not if they are just available in the grid.
> 
> I would say that depends a bit on what we actually want from that function, i.e. how strict we want to be ...
> 
> * we could just do a simple "is LO? then skip else proceed as before"
> * or we do a more strict check: "I'm asking for NLO - check that _all required_ ingredients are there"

Yes, I believe the second option to be better. This is what I am doing now. Moreover I splitted the sv for `as` and `al` in such a way in the future we will be fine.

src/pineko/check.py
```diff
@@ -92,31 +93,100 @@ def is_fonll_b(fns, lumi):
...
     ----------
         grid: pineappl.grid.Grid
             Pineappl grid
+        max_as: int
+            max as order
+        max_al: int
+            max al order
+    Returns
+    -------
+        : bool
```
felixhekhorn, line 109: please check https://numpydoc.readthedocs.io/en/latest/format.html#returns

src/pineko/check.py
```diff
@@ -92,31 +93,100 @@ def is_fonll_b(fns, lumi):
...
+        is fact scale-variation available for as
+        : bool
+        is fact scale-variation available for al
     """
-    order_list = [order.as_tuple() for order in grid.orders()]
+    order_array = np.array([order.as_tuple() for order in grid.orders()])
+    order_mask = pineappl.grid.Order.create_mask(grid.orders(), max_as, max_al)
+    order_list = order_array[order_mask]
+    as_orders = []
+    al_orders = []
```
felixhekhorn, line 118: I like the idea of providing both `as` and `al` - however there is too much repetition here ... and the check for both is essentially the same ...

src/pineko/check.py
```diff
@@ -92,31 +93,100 @@ def is_fonll_b(fns, lumi):
...
     for order in order_list:
-        if order[-1] != 0:
-            return
-    raise ValueError("Factorization scale variations are not available for this grid")
-
-
-def contains_ren(grid):
+        as_orders.append(order[0])
+        al_orders.append(order[1])
+    min_as = min(as_orders)
```
felixhekhorn, line 122: Let's use the tools we already have - to find out `min_as` you can just call `create_mask` again

src/pineko/check.py
```diff
@@ -92,31 +93,100 @@ def is_fonll_b(fns, lumi):
...
+    min_al = min(al_orders)
+    order_as_is_present = False
+    order_al_is_present = False
+    sv_as_present = False
+    sv_al_present = False
+    for order in order_list:
+        # fact sv starts at NLO with respect to the first non zero order
+        if order[0] == min_as + 1:
+            order_as_is_present = True
+            if order[-1] != 0:
```
felixhekhorn, line 132: since you said you like the idea of doing a stricter test and this I really meant here: we can exactly predict the necessary powers of logs: at each order you need to have exactly `0..d`  logs, where d is the difference in powers of coupling; e.g. at NLO you need 0,1 logs

src/pineko/check.py
felixhekhorn: also for parameters

felixhekhorn (changes requested):
of course for renormalization the same comments apply ...

src/pineko/check.py
andreab1997: Do you mean that maybe we need a separate function to do the checks? Or do you have something else in mind? 
Because of course is repetitive but they are not exactly the same. Moreover, I don't believe that splitting the check on `as` and the check on `al` would be worth.

src/pineko/check.py
andreab1997: Yes but `order_mask` is just returning a list full of `True` or `False` according to which order you need to take in the same order of the 'grid.orders()' list. So I believe that of course it is possible to get the `min_as` and the `min_al` values using `order_mask` but it would be less trivial than just using a loop and then the `min` func.

src/pineko/check.py
andreab1997: In practice at the moment the test relies on the fact that a grid can either have the scale variations or not and in the former case the grid would have the scale-variations for all the (available) orders. But of course this is probably not guaranteed

src/pineko/check.py
felixhekhorn: as you prefer ...

src/pineko/check.py
felixhekhorn: yes, I thought about defining an (inline) function to be called twice ... however, I agree it is a matter of taste, so you can either decide yourself or ask for a third opinion, e.g. @AleCandido

src/pineko/check.py
felixhekhorn: I agree the mixed case would be complicated (but that is, if at all, a long term problem) ... again it boils down to the question on [how strict we want to be](https://github.com/N3PDF/pineko/pull/33#issuecomment-1185519566) ... for this let's definitely ask for an opinion to @AleCandido - just to say, your current version is in between the two possibilities I said (and of course viable)

AleCandido: I guess in the long term, the "very strict" might be useful. We can even tune according to feedbacks, and expose flags/options to make customization available to the user. From very detailed option to something like `--strictness 3`.

AleCandido: In the short term the answer is: close this as soon as possible, it's not a priority, while we're a bit in a hurry for theory covmat. If we have spare time at the moment (meaning "all fktables/fits running, I can't do anything to speed up"), maybe you (@andreab1997) can take over @felixhekhorn task of checking different SV methods.
My feeling is that is a black hole of time, and we'll get almost nothing good, so we should limit the resources dedicated to it. On the other hand, the only good way to close this long-lasting task is to come with a theory covmat paper @NNLO. And if we want a paper, it wouldn't be a bad thing to have a *sufficient* understanding of what's going on - the *sufficient* concept will change with time, of course (and it is already changed).

andreab1997: Ok I agree. Then my proposal is to leave the test as it is now and, for the moment, to rely on the fact that the grids can only either have the scale variations or not. At least in this way I can compute the fktables now. Then I will update the check and for sure I would do what you said (@AleCandido0, i.e. asking the user for a `strictness` parameter. Do you agree?

felixhekhorn:
Discussing with @AleCandido we agreed to close this PR want so ever and start computing the FK tables *now*

src/pineko/check.py
AleCandido: I agree. About the **then** part we will rediscuss *then* :)

AleCandido (approved):

felixhekhorn (approved):
I guess you checked that now we can actually do the real life examples?

andreab1997:
> I guess you checked that now we can actually do the real life examples?

I tried with a couple of grids and it worked. I can try it on more if you wish :)

felixhekhorn:
Fine :upside_down_face:

PR #41, opened 2022-08-18
Fixing integrability opcards

andreab1997:
We realized that the opcards for the integrability grids were wrongly generated by pineko. In particular the point in targetgrid was not contained in the interpolation_x_grid list.

### DIFF
--- a/src/pineko/evolve.py
+++ b/src/pineko/evolve.py
@@ -71,7 +71,7 @@ def write_operator_card(pineappl_grid, default_card, card_path, xif):
     q2_grid = (xif * xif * muf2_grid).tolist()
     operators_card["targetgrid"] = x_grid.tolist()
     operators_card["Q2grid"] = q2_grid
-    if is_integrability(pineappl_grid.key_values()):
+    if "integrability_version" in pineappl_grid.key_values():
         operators_card["interpolation_polynomial_degree"] = 1
         x_grid_int = copy.deepcopy(x_grid.tolist())
         x_grid_int.append(1.0)
@@ -79,15 +79,6 @@ def write_operator_card(pineappl_grid, default_card, card_path, xif):
     with open(card_path, "w", encoding="UTF-8") as f:
         yaml.safe_dump(operators_card, f)
     return x_grid, q2_grid
-
-
-def is_integrability(grid_key_values):
-    """Check if the grid is an integrability grid."""
-    try:
-        int_version = grid_key_values["integrability_version"]
-    except KeyError:
-        return False
-    return True
 
 
 def evolve_grid(
--- a/src/pineko/evolve.py
+++ b/src/pineko/evolve.py
@@ -15,7 +15,9 @@ import yaml
 from . import check, comparator
 
 
-def write_operator_card_from_file(pineappl_path, default_card_path, card_path, xif):
+def write_operator_card_from_file(
+    pineappl_path, default_card_path, card_path, xif, integrability
+):
     """Generate operator card for a grid.
 
     Parameters
@@ -28,6 +30,8 @@ def write_operator_card_from_file(pineappl_path, default_card_path, card_path, x
         target path
     xif : float
         factorization scale variation
+    integrability : bool
+        whether or not the grid is an integrability grid
 
     Returns
     -------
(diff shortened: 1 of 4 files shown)


felixhekhorn (approved):
- I would appreciate if your commit message contains more information then just "Fixing" :upside_down_face: e.g. "Add int parameter to API" or "Propagate integrability info" or whatever ;-)
- if this fix is doing the trick - I'm happy to merge for now ...

for the future we may however reconsider:
- writing unit tests (if you have some spare time now, even do it now)
- and we should make ourselves aware, that this behaviour indeed relies on float arithmetic, which is potentially dangerous
  - because the way we're hacking it, we're relying on the fact that EKO will return an identity and hence we can get away with what ever xgrid ... 
  - to be specific this relies on the fact that the number here https://github.com/NNPDF/runcards/blob/3a5b7c9dc053707bb74d181c69b5add50c5a385f/pinecards/NNPDF_INTEG_XT3_40/integrability.yaml#L4 is inside float equality equal to Q0^2 in the theory card (which is even outside our scope)

andreab1997:
> * I would appreciate if your commit message contains more information then just "Fixing" 🙃 e.g. "Add int parameter to API" or "Propagate integrability info" or whatever ;-)

Yes sorry but they were just some minor stuff I forgot. 

> * if this fix is doing the trick - I'm happy to merge for now ...
I am testing now

> 
> for the future we may however reconsider:
> 
> * writing unit tests (if you have some spare time now, even do it now)
> * and we should make ourselves aware, that this behaviour indeed relies on float arithmetic, which is potentially dangerous
>   
>   * because the way we're hacking it, we're relying on the fact that EKO will return an identity and hence we can get away with what ever xgrid ...
>   * to be specific this relies on the fact that the number here https://github.com/NNPDF/runcards/blob/3a5b7c9dc053707bb74d181c69b5add50c5a385f/pinecards/NNPDF_INTEG_XT3_40/integrability.yaml#L4 is inside float equality equal to Q0^2 in the theory card (which is even outside our scope)

andreab1997:
> > * I would appreciate if your commit message contains more information then just "Fixing" 🙃 e.g. "Add int parameter to API" or "Propagate integrability info" or whatever ;-)
> 
> Yes sorry but they were just some minor stuff I forgot.
> 
> > * if this fix is doing the trick - I'm happy to merge for now ...
> 
> I am testing now

Ok I can confirm it works, should I merge? @felixhekhorn

felixhekhorn (changes requested):
actually, inside the theory command we should determine the "integrability-property" from the grid - those grids have a marker: the `integrability_version`

felixhekhorn:
> actually, inside the theory command we should determine the "integrability-property" from the grid - those grids have a marker: the `integrability_version`

I guess this works even for the file based API since at opcard generation time you need to have a grid present

andreab1997:
> actually, inside the theory command we should determine the "integrability-property" from the grid - those grids have a marker: the `integrability_version`

Right, this seems a better solution than asking for a flag

src/pineko/evolve.py
```diff
@@ -71,11 +71,25 @@ def write_operator_card(pineappl_grid, default_card, card_path, xif):
...
+        x_grid_int.append(1.0)
+        operators_card["interpolation_xgrid"] = list(x_grid_int)
     with open(card_path, "w", encoding="UTF-8") as f:
         yaml.safe_dump(operators_card, f)
     return x_grid, q2_grid
 
 
+def is_integrability(grid_key_values):
+    """Check if the grid is an integrability grid."""
+    try:
```
felixhekhorn, line 86: why not just `return "integrability_version" in grid_key_values`? and since it then becomes a one-line function used a single time, I'd suggest to inline ..

andreab1997: Because if the grid is not integrability, it won't have the key `integrability_version` at all and so that would result in an error.

andreab1997: That is the reason for the `try/except`

felixhekhorn: hum? that's exactly what `in` does, isn't it?
```
In [1]: d = {"a": 1}

In [2]: "a" in d
Out[2]: True

In [3]: "b" in d
Out[3]: False
```

andreab1997: yes right :)

felixhekhorn (approved):
I like short PRs ;-)

andreab1997:
@scarlehoff is this ok to merge for you?

PR #48, opened 2022-10-05
Refactor sub scvar

andreab1997:


felixhekhorn (changes requested):

AleCandido:
I would propose to use [`enum`](https://docs.python.org/3/library/enum.html) for them, something like:

```py
from enum import Enum, auto

class Scale(Enum):
    REN = auto()
    FACT = auto()

@click.argument(
    "scale",
    metavar="SCALE",
    type=click.Choice(list(map(lambda x: x.name.lower(), Scale)), case_sensitive=False),
)
...
```

It is more explicit, and may give more introspection in debugging.

andreab1997:
> I would propose to use [`enum`](https://docs.python.org/3/library/enum.html) for them, something like:
> 
> ```python
> from enum import Enum, auto
> 
> class Scale(Enum):
>     REN = auto()
>     FACT = auto()
> 
> @click.argument(
>     "scale",
>     metavar="SCALE",
>     type=click.Choice(list(map(lambda x: x.name.lower(), Scale)), case_sensitive=False),
> )
> ...
> ```
> 
> It is more explicit, and may give more introspection in debugging.

Do you mean something similar to what I have done in the last commit?

src/pineko/cli/check.py
```diff
@@ -36,54 +38,46 @@ def sub_compatibility(grid_path, operator_path, xif):
...
+    FACT = auto()
+
+
 @subcommand.command("scvar")
 @click.argument("grid_path", metavar="PINEAPPL", type=click.Path(exists=True))
 @click.argument(
     "scale",
     metavar="SCALE",
-    type=click.Choice(["ren", "fact"]),
+    type=click.Choice(list(map(lambda x: x.name.upper(), Scale)), case_sensitive=False),
```
AleCandido, line 53: You already wrote them uppercase, and so the `map` should be redundant:
```suggestion
    type=click.Choice(list(el.name for el in Scale), case_sensitive=False),
```

And in any case you set `case_sensitive=False`, so the rest should be up to him.

src/pineko/cli/check.py
```diff
@@ -36,54 +38,46 @@ def sub_compatibility(grid_path, operator_path, xif):
...
+    success = "[green]Success:[/] grids contain"
+    error = "[red]Error:[/] grids do not contain"
+    scale_variations = {
+        Scale.REN.name: " renormalization scale variations",
+        Scale.FACT.name: " factorization scale variations",
+    }
+    if scale not in scale_variations.keys():
+        raise ValueError("Scale variation to check can be one between ren and fact")
+    sv = scale_variations[scale]
+    funcs = {Scale.REN.name: check.contains_ren, Scale.FACT.name: check.contains_fact}
```
AleCandido, line 70: Also this one, you can assign to the value of `Scale` (just make it a tuple, or a dedicate object).

E.g.:
```py
class Scale(Enum):
    REN = ("renormalization scale variations", check.contains_ren)
    ...
```

There is also the intermediate option, you can make it a [`namedtuple`](https://docs.python.org/3/library/collections.html?highlight=namedtuple#collections.namedtuple):

```py
from collections import namedtuple

ScaleValue = namedtuple('ScaleValue', ['descr', 'check'])

class Scale(Enum):
    REN = ScaleValue("renormalization scale variations", check.contains_ren)
    ...
    
Scale[myscale].value.check
```

AleCandido: > Named tuples assign meaning to each position in a tuple and allow for more readable, self-documenting code.

src/pineko/cli/check.py
```diff
@@ -36,54 +38,46 @@ def sub_compatibility(grid_path, operator_path, xif):
...
+    func_to_call = funcs[scale]
+    # Call the function
+    is_sv_as, is_sv_al = func_to_call(grid, max_as_order, max_al_order)
+    conditions = {" for as": is_sv_as, " for al": is_sv_al}
```
AleCandido, line 74: I would define an `Enum` also here: let's start limit dictionaries when not required.

`dict` is very little transparent for the code: you won't get autocompletion or static check, and it is mutable also when you don't want it.

AleCandido:
> Do you mean something similar to what I have done in the last commit?

Similar yes, but we can still improve: I added a bunch of comments.

I trust that is working, but since our code is growing, I'm trying to focus on improving the quality, to make it more maintainable (just because we're scaling, and trying to stabilize our infrastructure).

src/pineko/cli/check.py
```diff
@@ -38,11 +39,14 @@ def sub_compatibility(grid_path, operator_path, xif):
...
+ScaleValue = namedtuple("ScaleValue", ["descr", "check"])
+
+
 class Scale(Enum):
     """Auxiliary class to list the possible scale variations."""
 
-    REN = auto()
-    FACT = auto()
+    REN = ScaleValue(" renormalization scale variations", check.contains_ren)
+    FACT = ScaleValue(" factorization scale variations", check.contains_fact)
```
AleCandido, line 49: The initial space is kind of weird: drop it, and just add it back in the place you are using it
```suggestion
    REN = ScaleValue("renormalization scale variations", check.contains_ren)
    FACT = ScaleValue("factorization scale variations", check.contains_fact)
```

src/pineko/cli/check.py
```diff
@@ -60,24 +64,18 @@ def sub_scvar(grid_path, scale, max_as_order, max_al_order):
...
+        raise ValueError("Scale variation to check can be one between ren and fact")
     conditions = {" for as": is_sv_as, " for al": is_sv_al}
     for cond in conditions.keys():
         to_write = ""
         if conditions[cond]:
             to_write += success
         else:
             to_write += error
-        to_write += sv
+        to_write += Scale[scale].value.descr
```
AleCandido, line 79: That is here:
```suggestion
        to_write += " " + Scale[scale].value.descr
```

src/pineko/cli/check.py
andreab1997: Ok I believe for the other case I have done what you suggested. However, in this case, it is not clear to me how to use enum (and actually how to avoid the dictionary).

src/pineko/cli/check.py
```diff
@@ -36,54 +39,43 @@ def sub_compatibility(grid_path, operator_path, xif):
...
+            to_write += success
         else:
-            rich.print(
-                "[red]Error:[/] grids do not contain factorization scale variations for al"
-            )
-    else:
-        raise ValueError("Scale variation to check can be one between xir and xif")
+            to_write += error
+        to_write += " " + Scale[scale].value.descr
+        to_write += cond
```
AleCandido, line 80: That is here:
```suggestion
        to_write += f" for {coupling.name}"
```
I used the `coupling` variable because at this point it will be just the name of the coupling.

src/pineko/cli/check.py
AleCandido: 1. You are using the `dict` only to iterate on them
    - and, by the way, if use the value as well, do not iterate on `.keys()`, and access for the value, just iterate on `.items()`
    - moreover, if you iterate on a dictionary is already iterating on its keys, so calling the `.keys()` method is not making any difference
2. you are creating the keys only to write them later, but you see that there is already redundancy: the `" for "` part is repeated twice, you can just drop it and put that below

At this point, you can also avoid the mapping and just iterate over the couples, i.e. start defining an `Enum`:

```py
CouplingInfo = namedtuple("CouplingInfo", ["descr", "theory"])

class Coupling(Enum):
    as = CouplingInfo("strong", "QCD")
    al = CouplingInfo("electromagnetic", "QED")
```

src/pineko/cli/check.py
```diff
@@ -36,54 +39,43 @@ def sub_compatibility(grid_path, operator_path, xif):
...
-            )
+    success = "[green]Success:[/] grids contain"
+    error = "[red]Error:[/] grids do not contain"
+    # Call the function
+    try:
+        is_sv_as, is_sv_al = Scale[scale].value.check(grid, max_as_order, max_al_order)
+    except KeyError:
+        raise ValueError("Scale variation to check can be one between ren and fact")
+    conditions = {" for as": is_sv_as, " for al": is_sv_al}
+    for cond in conditions.keys():
```
AleCandido, line 73: And here you use them to iterate:
```suggestion
    for coupling, condition in [(Coupling.as, is_sv_as), (Coupling.al, is_sv_al)]:
```

src/pineko/cli/check.py
```diff
@@ -36,54 +39,43 @@ def sub_compatibility(grid_path, operator_path, xif):
...
+        to_write = ""
+        if conditions[cond]:
```
AleCandido, line 75:
```suggestion
        if condition:
```

src/pineko/cli/check.py
AleCandido: Most likely this class can also be lifted somewhere else (later on), since it is a rather general definition.

src/pineko/cli/check.py
AleCandido: I guess we can do even shorter than this, with something like:

```py
        conditions = Scale[scale].value.check(grid, max_as_order, max_al_order)
    ...
    for coupling, condition in zip(Coupling, conditions):
```

But given that they are just two, here is not very relevant...

andreab1997:
Ok thank you @AleCandido for the suggestions, this should be fine now

felixhekhorn (approved):

AleCandido (approved):

PR #52, opened 2022-10-19
Refactor benchmark workflow

andreab1997:
As @AleCandido suggested, I am refactoring the workflow in order to download the files needed for benchmarks directly from the server (https://data.nnpdf.science/)


andreab1997:
This should be ready but I believe we need to install wget on the container, right? @AleCandido

felixhekhorn:
as @AleCandido said: please separate a script that you call in the workflow, but that one can run even with out (i.e. locally)

andreab1997:
As you can see, the tests after [ebf6342](https://github.com/NNPDF/pineko/pull/52/commits/ebf634204a3637fdbdd93ceb318fb6feb5280538) work. This is because I have just used `apt-get install wget` in the workflow (and of course this is not what we want). However, at least we know that the only thing that we miss is to install `wget` in the container.

AleCandido:
> However, at least we know that the only thing that we miss is to install `wget` in the container.

Just to spread the knowledge about how containers are made:
1. you should go here https://github.com/N3PDF/workflows/blob/main/containers/bench-evol/Containerfile and add `wget` to the list of packages to be installed, [in the second stage](https://github.com/N3PDF/workflows/blob/fb7177a97f2bb2e78635027cc9be0d1ff3525aea/containers/bench-evol/Containerfile#L35)
2. you should follow the instructions here https://github.com/N3PDF/workflows/blob/main/containers/README.md to build and upload the new image

But don't do it, there is already `curl` installed, so replace `wget` by `curl` (`curl` main purpose is making HTTP requests, so it is not downloading files by default, but you can with a couple of options).

AleCandido:
Ah no, wait. My bad, you are using recursive download, so [you need `wget`](https://superuser.com/a/705827)

andreab1997:
> 2\. https://github.com/N3PDF/workflows/blob/main/containers/README.md

Anyway I believe I need some to provide the github token in some way because if I do `podman pull ghcr.io/n3pdf/bench-evol:latest` I get the error `Error: initializing source docker://ghcr.io/n3pdf/bench-evol:latest: Requesting bearer token: invalid status code from registry 403 (Forbidden)`

AleCandido:
> ```sh
> # login to GitHub registry with user credentials (not organization), see [3]
> echo ${PAT} | podman login ghcr.io -u <username> --password-stdin
> ```
> [...]
> [3]: official [GitHub registry docs](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry)

As specified in the linked docs, you should give your PAT the `write:packages` scope.

AleCandido:
No, wait, I'm stupid. I pointed you to the `bench-evol` container, but here you are using the `lhapdf` one...

andreab1997:
> No, wait, I'm stupid. I pointed you to the `bench-evol` container, but here you are using the `lhapdf` one...

Here in `bench.yml` I am using `bench-evol` that is why I am trying to build it. Should I use something else? Anyway I need `lhapdf` for the benchmarks

AleCandido:
If you just need `lhapdf`, and not APFEL or Pegasus or QCDNUM, then use the more restricted `lhapdf` container (same repo).

However, I tried to run `install.sh` and I had no issue. Try yourself to do it in a fresh environment (create with `virtualenv env` and just activate), and (if still not working) try on a Linux machine.

andreab1997:
> If you just need `lhapdf`, and not APFEL or Pegasus or QCDNUM, then use the more restricted `lhapdf` container (same repo).
> 
> However, I tried to run `install.sh` and I had no issue. Try yourself to do it in a fresh environment (create with `virtualenv env` and just activate), and (if still not working) try on a Linux machine.

But actually the install script fails on the container (so after having done `./build.bash bench-evol`) which is a Linux machine, right?

AleCandido:
> `g++: fatal error: Killed signal terminated program cc1plus compilation terminated. make[2]: *** [Makefile:557: liblhapdf_yaml_cpp_la-emitterutils.lo] Error 1` Any clues?

**Killed signal terminated program** is telling you that something killed it, like when you press Ctrl-C...
Compiled on `dom`, and it is just working.

AleCandido:
Ok, I actually have the new image on `dom`, but I can't upload it, because of the following error:
```
Error: writing blob: initiating layer upload to /v2/n3pdf/bench-evol/blobs/uploads/ in ghcr.io: denied: permission_denied: write_package
```

Unfortunately, GitHub just introduced [fine-grained PAT](https://github.blog/2022-10-18-introducing-fine-grained-personal-access-tokens-for-github/), and my classic token, with `write:packages` permission, it is not enough to publish a package to the @N3PDF organization.
Moreover, I can't issue a fine-grained token with access to `N3PDF/workflows`, because it is not allowed by @N3PDF organization.

Both (classic and fine-grained) can be allowed to perform actions on @N3PDF, but @scarrazza or @scarlehoff has to do it explicitly at:
https://github.com/organizations/N3PDF/settings/personal-access-tokens-onboarding

AleCandido:
Ok, now I've done everything I could: we also have a releasing workflow, such that packages can be built in the CI and released automagically.
https://github.com/N3PDF/workflows/blob/v2/.github/workflows/publish-containers.yml

But it is still failing:
https://github.com/N3PDF/workflows/actions/runs/3302837504
with the usual error:
```
Error: writing blob: initiating layer upload to /v2/n3pdf/lhapdf/blobs/uploads/ in ghcr.io: denied: permission_denied: write_package
```

So, not even the `${{ github.token }}` (or `${GITHUB_TOKEN}`, is just the same) is able to release, here the default permissions:
https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

Therefore, it is really up to @N3PDF admins, since the `Default access (permissive)` has the `write:packages` scope, this means it has been somehow set to `Default access (restricted)`.

Here the guide on how to control permissions for the `${GITHUB_TOKEN}`.
https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token

The moment it is fixed, we can get the new container just by adding a tag, or rerunning the last workflow.

Now, I can't do anything more for this PR, it is totally up to @scarrazza and @scarlehoff

AleCandido:
Latest update: everything still holds for new packages, I have no idea on how to publish them if the tokens are not accepted.

However, for existing packages of which I'm the owner I could solve it. So, not `bench_evol`, since @niclaurenti is the owner, but I can instruct him on how to do it.

But here the dependency is only `lhapdf`, and for that there is an `lhapdf` container I could publish (since I'm the owner of that). `bench_evol` has been tailored for EKO benchmarks, providing APFEL, QCDNUM, Pegasus, and so on.

So, the solution was to switch benchmark.

Now it is still failing because of the dependency on `lz4`, since it requires to build the wheel and `gcc` is not available in the container. But this is in the range of things that have a straight solution :)
(now it is more a problem that there are more than one, so I will try to pick the best, but at some point any is fine to begin with)

AleCandido:
Problem with `lz4` solved on EKO side, now it is fine.

Current problem is that I had to release the version with Composite Output struct, so the old `output.Output` is not any longer there.
Two solutions:
1. we revert the version of EKO, and we merge with the failing workflow (then open a PR to update the dependency and the EKO API usage)
2. we update immediately here

I believe it should not require a major rework to use the new structure, it's just a matter of updating the usage here and there. I would go for 2.

felixhekhorn:
> Current problem is that I had to release the version with Composite Output struct, so the old `output.Output` is not any longer there.

for that exact reason: can you please drop the new releases on the 0.10 branch and commit instead a 0.11 version? Otherwise we run into lot's of trouble all over the place, because poetry thinks they are compatible, which they are not

AleCandido:
I actually thought the were compatible, I didn't remember I dropped the struct. However, you can keep going and reshuffle the tags :)

felixhekhorn:
I think a tag can not be renamed, can it? (that wouldn't make sense to me ...) and, I guess, the problem is PyPI (since poetry will look there), right?

AleCandido:
With "reshuffle tags" I mean:

- delete the tag from the repo (actually, from v0.10.3 to v0.10.5 included)
- *yank* the release from PyPI (there is only 0.10.5, the other workflows failed)
- release v0.11.0, and it will automatically generate the new version on PyPI

felixhekhorn:
ok, I believe to have done ...

AleCandido:
If you can fix it would be great :)

andreab1997:
> If you can fix it would be great :)
> 
> P.S.: try, if it's too complicated I'll do it

Actually I believe that I just need to change every `eko.output.Output.load()` to `eko.output.EKO.load()`. The only problem is that the ekos that I use for the tests are old and eko gives me errors when trying to load them.

AleCandido:
> Actually I believe that I just need to change every `eko.output.Output.load()` to `eko.output.EKO.load()`. The only problem is that the ekos that I use for the tests are old and eko gives me errors when trying to load them.

Maybe you want to use `output.legacy.load_tar()`, or something else in the `legacy` module. If they really work, they are supposed to be used for this.

andreab1997:
> > Actually I believe that I just need to change every `eko.output.Output.load()` to `eko.output.EKO.load()`. The only problem is that the ekos that I use for the tests are old and eko gives me errors when trying to load them.
> 
> Maybe you want to use `output.legacy.load_tar()`, or something else in the `legacy` module. If they really work, they are supposed to be used for this.

Yes but it seems that something fishy is happening. I have `eko==0.11.0` but it has no `output.legacy` module. Is this a problem of my installation or do you have the same problem?

AleCandido:
You need to import it:
```py
import eko.output.legacy
```

I just downloaded eko v0.11.0 and it is there :)

AleCandido:
> Yes, I know that I need to import it :)

It is going to change, see NNPDF/eko#145

andreab1997:
Ok I fixed the installation. Anyway still it is not working. I believe that the best choice we have is to recompute the new ekos (I am going to do that anyway for sv so it is zero effort).

andreab1997:
Ok I fixed the installation. Anyway still it is not working. I believe that the best choice we have is to recompute the new ekos (I am going to do that anyway for sv so it is zero effort).

andreab1997:
> > Ok I fixed the installation. Anyway still it is not working. I believe that the best choice we have is to recompute the new ekos (I am going to do that anyway for sv so it is zero effort).
> 
> Nothing will change: the computation is still the old one, so it is just converting old to new at some point. `load_tar` should do the exact same.

Ok then it is not. I will try to understand why.

AleCandido:
For the time being, the problem I found is that the operator card is missing a lot of expected pieces. This is what I see:
```
dict_keys(['Q2grid', 'eko_version', 'inputgrid', 'inputpids', 'interpolation_is_log', 'interpolation_polynomial_degree', 'interpolation_xgrid', 'q2_ref', 'targetgrid', 'targetpids'])
```
But the loader is looking for:
```
interpolation_xgrid
pids
inputgrid
targetgrid
inputpids
targetpids
ev_op_max_order
ev_op_iterations
interpolation_polynomial_degree
interpolation_is_log
backward_inversion
n_integration_cores
```

Bypassing them, it is still missing an attribute "Q0". Providing it, there is still a mismatch in the layout from old style operators (dict with keys `["operators", "alphas", "operator_errors"]`) and new class (dataclass with members `operator` and `error`).

Recasting manually, it is able to load the EKO.

AleCandido:
In practice, we need a new version of EKO, that is properly bridging the old format with the new one. It has been purely tested.

(I can't tell I expected it, but I didn't follow the last steps so carefully, so I was worried...)

AleCandido:
@andreab1997 try to upgrade to EKO v0.11.1, and then tell me if there is any error still

andreab1997:
> @andreab1997 try to upgrade to EKO v0.11.1, and then tell me if there is any error still

Now it works but a lot of things changed....
I need to change a lot of things. For example, is there a method that substitute the old `xgrid_reshape()`?

AleCandido:
> I need to change a lot of things. For example, is there a method that substitute the old `xgrid_reshape()`?

It's inside `output.manipulate`. Now it is a stand-alone function that takes an `EKO` object as first parameter.

andreab1997:
> > I need to change a lot of things. For example, is there a method that substitute the old `xgrid_reshape()`?
> 
> It's inside `output.manipulate`. Now it is a stand-alone function that takes an `EKO` object as first parameter.

Ok I solved a bunch of problems but I have two left that I don't know how to solve:
1. (maybe easy) in `evolve.evolve_grid` (in `alphas_values = [op["alphas"] for op in operators.Q2grid]`) I need the `alphas` values from the EKO that now are no more there.
2. In `check_grid_and_eko_compatible` I solved most of the problems but still I get an error from the test.
I am pushing what I have so you can see the problems and the checks that are failing.

AleCandido:
> 1. I need the `alphas` values from the EKO that now are no more there.

This could have been a problem on its own: because of scale variations, alphas values in the operator were deprecated, they had to be replaced.
You should initialize a `Coupling` object, and then query it for the value. About this also @niclaurenti should be able to help you.

> 2. I solved most of the problems but still I get an error from the test.

Checking, on my way :)

AleCandido:
So, tests are passing, but I'm not sure about how much this is telling us.
Main reason why benchmark are still failing is because you forgot to upgrade the dependency. Maybe there is another reason behind, we'll see in a moment.

AleCandido:
https://github.com/NNPDF/pineko/blob/8cc14346729fd6e559309d07378bb014d616969e/benchmarks/bench_checks.py#L24-L26

`targetgrid` is supposed to be of `eko.interpolation.XGrid` type, and you passed a list. Type hint here would have helped:
```py
     def xgrid_reshape(
        eko: EKO,
        targetgrid: Optional[interpolation.XGrid] = None,
        inputgrid: Optional[interpolation.XGrid] = None,
    ):
    ```

Then we have a problem in the PineAPPL interface:
```
/usr/local/lib/python3.10/site-packages/pineappl/grid.py:281: in convolute_eko
    [op["operators"] for op in operators["Q2grid"].values()]
```
this should be replaced by:
```
[op[q2].operator for q2 in operators]
```
or something similar. And eventually should be suppressed, because this way you are loading the full EKO in memory, nullifying the effort of keeping only one in memory.

**Comment by AleCandido** [comment]
The PineAPPL problem will be solved temporarily in the simple way (the one suggested above). First you should try to run locally, and once it's working we'll notify Christopher to make a new release.

The xgrid and alphas problem should be solvable inside Pineko.

**Comment by andreab1997** [comment]
Ok now I believe that the only left problem is inside `convolute_eko` so it is `pineappl` business. @AleCandido Do you agree?

**Comment by AleCandido** [comment]
9e37cff introduced a new issue (before tests were not failing...)

**Comment by andreab1997** [comment]
So the last problem is that inside `convolute` I don't have the theory card and so I cannot construct the `Coupling` object to be passed to `evolve_grid`. So, when the theory card will be finally available inside the eko, we will be able to load it from it and we will have no problems. Currently the theory card is not contained so the only solution I can see is to add an argument for the tcard path and load it.

**Comment by AleCandido** [comment]
To construct the [`Couplings`](https://github.com/NNPDF/eko/blob/e3466fae35fd245c7e1a99158d8582f590276555/src/eko/couplings.py#L359-L370) object, you don't need the full theory card.

In any case, the theory should be available in Pineko, because it has EKO is constructed by Pineko. However, we'll have a look together if propagating through the EKO file is the most sensible choice (or the only sensible one).

**Comment by andreab1997** [comment]
> To construct the [`Couplings`](https://github.com/NNPDF/eko/blob/e3466fae35fd245c7e1a99158d8582f590276555/src/eko/couplings.py#L359-L370) object, you don't need the full theory card.
> 
> In any case, the theory should be available in Pineko, because it has EKO is constructed by Pineko. However, we'll have a look together if propagating through the EKO file is the most sensible choice (or the only sensible one).

Yes but the `convolute` command was exactly meant to avoid using all the configs of `pineko` and just convolute a grid a with an eko.

**Comment by AleCandido** [comment]
> Yes but the `convolute` command was exactly meant to avoid using all the configs of `pineko` and just convolute a grid a with an eko.

For the time being, that convolution is not well-defined on its own. Most likely you are right, we just need the theory card inside the EKO. But eventually, that theory will contain only values affecting the EKO itself, so I wonder if we need something more for alphas.

**Comment by andreab1997** [comment]
> > Yes but the `convolute` command was exactly meant to avoid using all the configs of `pineko` and just convolute a grid a with an eko.
> 
> For the time being, that convolution is not well-defined on its own. Most likely you are right, we just need the theory card inside the EKO. But eventually, that theory will contain only values affecting the EKO itself, so I wonder if we need something more for alphas.

In fact at the moment that convolution is not working but I would say is not blocking this: we can fix it later

**Comment by andreab1997** [comment]
Ok now everything should be fine except for `pineappl`. What do we want to do for that?

File: benchmarks/bench_evolve.py [hunk 1]
```diff
@@ -29,6 +29,16 @@ def benchmark_write_operator_card_from_file(tmp_path, test_files):
...
+    # In order to check if the operator card is enough for eko, let's compute the eko
+    tcard = pineko.theory_card.load(208)
+    # I need smaller x and q grids in order to compute a small eko
+    small_x_grid = [1.0e-6, 1.0e-5, 1.0e-4, 1.0e-3, 1.0e-2, 1.0e-1, 1.0]
+    small_q2_grid = [100.0]
+    myopcard["xgrid"] = small_x_grid
+    myopcard["targetgrid"] = small_x_grid
+    myopcard["inputgrid"] = small_x_grid
+    myopcard["Q2grid"] = small_q2_grid
+    ops = eko.run_dglap(theory_card=tcard, operators_card=myopcard)
```
AleCandido, line 41: Thank you!

AleCandido:
@andreab1997 now benchmarks should be working (to be tested), apart for a single issue: the function supposed to load the theory card is failing.
https://github.com/NNPDF/pineko/blob/a94a9bcc3447455c6728a9e86842bf0e07899ace/benchmarks/bench_evolve.py#L21

Plus, I modified the functions writing the opcard, and now they also require the theory card to be provided. Of course, there are a couple of calls in the code in which is not. The linter is able to find them, and it's causing the tests to fail. Also that one should be fixed :)

AleCandido:
Ok @andreab1997: I started improving a few things, but they are inconsistently scattered. In any case, it works :)

As soon as you review it, we can merge and start using EKO v0.11

AleCandido:
Just rebased on `main`. There was a conflict just with the renaming `q2_grid` -> `muf2_grid` and similar.

andreab1997:
I am reviewing it now :)

src/pineko/evolve.py
```diff
@@ -1,21 +1,36 @@
...
+DEFAULT_PIDS = [-5, -4, -3, -2, -1, 21, 22, 1, 2, 3, 4, 5]
 
-def write_operator_card_from_file(pineappl_path, default_card_path, card_path, xif):
+
+def write_operator_card_from_file(
+    pineappl_path: os.PathLike,
+    default_card_path: os.PathLike,
+    card_path: os.PathLike,
+    xif: float,
+    tcard_path: os.PathLike,
```
andreab1997, line 26: Not really relevant but for me it would be more sensible to have `tcard_path` together with the other paths (i.e. before `xif`).

andreab1997:
Everything seems fine now (of course to be also tested on an actual FK computation). Anyway tests are passing and I don't see anything wrong. So I am asking your review again I then I believe we can merge. Thanks for this! @AleCandido @felixhekhorn

andreab1997:
@felixhekhorn I have added you at least you know what it is happened :)

src/pineko/evolve.py
AleCandido: I considered it as well, but in the end is something temporary, and we might want to drop it at some point (together with all the compatibility stuffs).

It is not there to stay, and at the moment I'd prioritize merging and keep going. We will reconsider in any case ^^

AleCandido (approved):

Issue #51, opened 2022-10-17
Refactor configs


AleCandido:
After a brief experience with the current configs

- [x] drop non-local paths:
    https://github.com/NNPDF/pineko/blob/4b1d511b46889e8ccfdce6aa8107842a1eccbe5a/src/pineko/configs.py#L104-L106
- [ ] include git-like resolution, i.e. check parent folders for a `pineko.toml` file
- [ ] test paths semantics
    - absolute paths are kept as they are
    - relative paths are relative to `pineko.toml`

Essentially, the best way to achieve the last one should be something like:
```py
pinekopath = pathlib.Path(...).absolute()

def absolute_or_pineko(path: pathlib.Path):
    try:
        return path.relative_to(pinekopath)
    except ValueError:
        return path

# or, almost the same
def absolute_or_pineko2(path: pathlib.Path):
    if path.is_absolute():
        return path
    return path.relative_to(pinekopath)
```

AleCandido:
It's already not a copy any longer: runcards still have the complicate one wiht `nestupdate` and everything else, while the thing here is just a stripped down minimal version of there.

I know you just copy-pasted from runcards, but afterwards we massively changed, just check the content and the git history.

So, it is not a bad idea to decouple the two. If in the end there is one that is particular better than the other, then we will consider switching also on the other side.
Whenever you have time @andreab1997, please, keep going :)

PR #56, opened 2022-10-20
Refactor configs

andreab1997:
This il linked to #51


src/pineko/configs.py
```diff
@@ -94,16 +94,24 @@ def detect(path=None):
...
+                "Provided path is not pointing to (or does not contain) the pineko.toml file"
+            )
+    # If no path is provided we need to look after the file.
+    # We want to check pwd and all its parent folders (without their subfolders of course) up to home.
+    paths = []
+    p = pathlib.Path.cwd()
+    home = pathlib.Path.home()
+    while p != home:
+        paths.append(p)
+        p = p.parent
```
felixhekhorn, line 114: you need to check whether `p` is contained in `home`: e.g. `/media/FK/pineko` vs. `/home/felix`

AleCandido: I would not stop at home, but just at root. So:
```py
while p is not p.parent:
    ...
```

Otherwise you would have an infinite loop if you are not working in your home (for example if you are inside `/tmp`, `/media`, `/var`, ...).
And in any case, there is no reason that `/pineko.toml` or `/home/pineko.toml` should exist, so better the quitting alternative :)

AleCandido: However, even easier:
```py
paths.extend(p.absolute().parents)
```

andreab1997:
As you may have noticed, I decided that if the user provides a guess, we need to find the `pineko.toml` in the folder he provided, otherwise we raise an error. This is because an user can have multiple `pineko.toml` file and we don't want to find the wrong one looking in all the parent folders if he provides what should be the correct folder to look in.

src/pineko/configs.py
```diff
@@ -94,16 +94,22 @@ def detect(path=None):
...
-    paths.append(pathlib.Path(appdirs.site_config_dir()))
+        configs_file = path / name if path.is_dir() else path
+        if configs_file.is_file():
+            return configs_file
+        else:
+            raise ValueError(
+                "Provided path is not pointing to (or does not contain) the pineko.toml file"
+            )
+    # If no path is provided we need to look after the file.
+    # We want to check pwd and all its parent folders (without their subfolders of course) up to root.
```
felixhekhorn, line 108:
```suggestion
    # We want to check cwd and all its parent folders (without their subfolders of course) up to root.
```

AleCandido: Done, just running `poetry update`

andreab1997:
Is this ok to be merged? (of course taking into account that the workflows are failing for #52 )

AleCandido:
Let's wait a second: we fix #52, merge that one, then rebase this and merge.

Let's try not to merge broken stuff: it's true that we are still work in progress on many aspects, but we are also trying to move towards stability.

AleCandido:
Rebased and changed PR base.

In any case, I'd not merge before #52 (that is almost ready, just waiting for review). The moment #52 is merged, the base will change again automatically :)

AleCandido:
Now that #52 has been merged, should we merge also this? @andreab1997

AleCandido (approved):

andreab1997:
> Now that #52 has been merged, should we merge also this? @andreab1997

Yes, let me only review this first. I am doing now

andreab1997 (commented):
Fine for me, everything seems good!

PR #59, opened 2022-11-02
Fix compatibility to eko 0.11

andreab1997:
We want to have `pineko` fully compatible with `eko==0.11`. For the moment this is the list of things to be done:

- [ ] Check `eko` tar file is legacy before loading and use the correct method
- [ ] Dump the eko in the correct way (using the new method)
- [ ] Fix the `inputpids` and `targetpids` entries of the opcard (most likely `rotations.inputpids` and `rotations.targetpids` should be transformation matrices).

### DIFF
--- a/src/pineko/check.py
+++ b/src/pineko/check.py
@@ -1,4 +1,6 @@
 """Tools to check compatibility of EKO and grid."""
+import tarfile
+
 import numpy as np
 import pineappl
 
@@ -198,3 +200,15 @@ def contains_ren(grid, max_as, max_al):
     if not order_al_is_present:
         sv_al_present = True
     return sv_as_present, sv_al_present
+
+
+def check_eko_is_legacy(eko_name):
+    """Check if the eko tar file is a legacy eko file."""
+    tar = tarfile.open(eko_name)
+    file_names = [file.name for file in tar.getmembers()]
+    legacy_patterns = ["operators.npy.lz4", "operator_errors.npy.lz4"]
+    for name in file_names:
+        for pattern in legacy_patterns:
+            if pattern in name:
+                return True
+    return False
--- a/src/pineko/theory.py
+++ b/src/pineko/theory.py
@@ -6,6 +6,7 @@ together with other theory ingredients (such as C-factors) are often
 commonly referred to as 'theory'.
 """
 import logging
+import tarfile
 import time
 
 import eko
@@ -397,7 +398,10 @@ class TheoryBuilder:
         # q2_grid = ocard["Q2grid"]
 
         # loading ekos
-        operators = eko.output.legacy.load_tar(eko_filename)
+        if check.check_eko_is_legacy(eko_filename):
+            operators = eko.output.legacy.load_tar(eko_filename)
+        else:
+            operators = eko.output.EKO.load(eko_filename)
         muf2_grid = operators.Q2grid
         # PineAPPL wants alpha_s = 4*pi*a_s
         # remember that we already accounted for xif in the opcard generation


src/pineko/check.py
```diff
@@ -198,3 +200,15 @@ def contains_ren(grid, max_as, max_al):
     if not order_al_is_present:
         sv_al_present = True
     return sv_as_present, sv_al_present
+
+
+def check_eko_is_legacy(eko_name):
+    """Check if the eko tar file is a legacy eko file."""
+    tar = tarfile.open(eko_name)
```
AleCandido, line 207: Use a context manager (i.e. `with` instruction)

AleCandido (changes requested):

benchmarks/bench_checks.py
```diff
@@ -15,9 +15,11 @@ def benchmark_check_grid_and_eko_compatible(test_files):
...
         test_files / "data/grids/208/NUTEV_CC_NU_FE_SIGMARED.pineappl.lz4"
     )
-    ekoop = eko.output.legacy.load_tar(
-        test_files / "data/ekos/208/HERA_CC_318GEV_EM_SIGMARED.tar"
-    )
+    eko_filename = test_files / "data/ekos/208/HERA_CC_318GEV_EM_SIGMARED.tar"
+    if pineko.check.check_eko_is_legacy(eko_filename):
+        ekoop = eko.output.legacy.load_tar(eko_filename)
+    else:
+        ekoop = eko.output.EKO.load(eko_filename)
```
AleCandido, line 22: Good, but just a suggestion: you are calling three functions with a single input (`eko_filename`), and this pattern is possibly going to repeat.

Just replace with a single call:
```suggestion
   ekoop = pineko.ekompatibility.load(eko_filename)
```

And define the function exactly in that way :)
(We could also move inside EKO, but I'd not go that far; since EKO is drastically moving in the new direction, let's keep compatibility in here as much as possible)

benchmarks/bench_evolve.py
```diff
@@ -81,7 +82,10 @@ def benchmark_evolve_grid(tmp_path, lhapdf_path, test_files, test_pdf):
     pine_path = test_files / "data/grids/208/HERA_CC_318GEV_EM_SIGMARED.pineappl.lz4"
     pinegrid = pineappl.grid.Grid.read(pine_path)
     eko_path = test_files / "data/ekos/208/HERA_CC_318GEV_EM_SIGMARED.tar"
-    eko_op = eko.output.legacy.load_tar(eko_path)
+    if pineko.check.check_eko_is_legacy(eko_path):
+        eko_op = eko.output.legacy.load_tar(eko_path)
+    else:
+        eko_op = eko.output.EKO.load(eko_path)
```
AleCandido, line 88: As said before, replace repeated pattern with function definition.

src/pineko/check.py
AleCandido:
```suggestion
     with tarfile.open(eko_name) as tar:
```

src/pineko/cli/check.py
```diff
@@ -30,7 +30,10 @@ def sub_compatibility(grid_path, operator_path, xif):
 
     """
     pineappl_grid = pineappl.grid.Grid.read(grid_path)
-    operators = eko.output.legacy.load_tar(operator_path)
+    if check.check_eko_is_legacy(operator_path):
+        operators = eko.output.legacy.load_tar(operator_path)
+    else:
+        operators = eko.output.EKO.load(operator_path)
```
AleCandido, line 36: Pattern repeated third time, the function is really the best choice at this point :)

src/pineko/cli/convolute.py
```diff
@@ -51,7 +51,10 @@ def subcommand(
     PDF is an optional PDF set compatible with the EKO to compare grid and FK table.
     """
     grid = pineappl.grid.Grid.read(grid_path)
-    operators = eko.output.legacy.load_tar(op_path)
+    if check.check_eko_is_legacy(op_path):
+        operators = eko.output.legacy.load_tar(op_path)
+    else:
+        operators = eko.output.EKO.load(op_path)
```
AleCandido, line 57: Fourth time...

src/pineko/theory.py
```diff
@@ -397,7 +398,10 @@ def fk(self, name, grid_path, tcard, pdf):
         # q2_grid = ocard["Q2grid"]
 
         # loading ekos
-        operators = eko.output.legacy.load_tar(eko_filename)
+        if check.check_eko_is_legacy(eko_filename):
+            operators = eko.output.legacy.load_tar(eko_filename)
+        else:
+            operators = eko.output.EKO.load(eko_filename)
```
AleCandido, line 404: Fifth.......

AleCandido (changes requested):

andreab1997:
Regarding the `targetpids` and `inputpids` issue, first of all I am not sure why now we have them both at top level and inside the key `rotations` in the opcard. Is only one of the two versions used? (@AleCandido, @felixhekhorn  ). 
Apart from that, I believe the best strategy would be to change them to the matrices we need and dump them in the operator card. However:
1. Is there a function inside eko that computes these matrices? Otherwise, is there another way to get them or should I compute them manually? 
2. I believe that the natural place to implement it is inside `eko.compatibility.update` (not in `pineko`). Do you agree?

AleCandido:
> Regarding the `targetpids` and `inputpids` issue, first of all I am not sure why now we have them both at top level and inside the key `rotations` in the opcard. Is only one of the two versions used? (@AleCandido, @felixhekhorn ).

Only the one in `rotations`. At least it should be like this.

benchmarks/bench_evolve.py
```diff
@@ -34,8 +34,12 @@ def benchmark_write_operator_card_from_file(tmp_path, test_files, test_configs):
     assert np.allclose(myopcard["rotations"]["xgrid"], x_grid)
     assert np.allclose(myopcard["rotations"]["targetgrid"], x_grid)
     assert np.allclose(myopcard["rotations"]["inputgrid"], x_grid)
```
AleCandido, line 36: Drop also `"targetgrid"` and `"inputgrid"`, they are as useless as the other two now :)

andreab1997: I believe that @felixhekhorn thinks that they are not useless and that `eko` would crash without them (and even without `targetpids` and `inputpids`). I am not an expert of `eko` so I wait for you

AleCandido: But here you are dropping `targetpids` and `inputpids`, so if you are dropping them, you can drop also the other two :)

andreab1997:
This should be fine now

AleCandido (approved):

benchmarks/bench_evolve.py
```diff
@@ -30,10 +32,16 @@ def benchmark_write_operator_card_from_file(tmp_path, test_files, test_configs):
     myopcard = yaml.safe_load(target_path.read_text(encoding="utf-8"))
     # Check if it contains all the information for eko
     assert np.allclose(myopcard["rotations"]["xgrid"], x_grid)
-    assert np.allclose(myopcard["rotations"]["targetgrid"], x_grid)
-    assert np.allclose(myopcard["rotations"]["inputgrid"], x_grid)
-    assert np.allclose(myopcard["rotations"]["inputpids"], pineko.evolve.DEFAULT_PIDS)
-    assert np.allclose(myopcard["rotations"]["targetpids"], pineko.evolve.DEFAULT_PIDS)
+    # Checking that inputpids and targetpids have been correctly removed
+    with pytest.raises(KeyError):
+        myopcard["rotations"]["inputpids"]
```
felixhekhorn, line 37: I think the `with` breaks here and the rest is never executed (so you need to repeat it - consider a loop ;-) )

felixhekhorn (changes requested):

benchmarks/bench_evolve.py
andreab1997: Oh you are right :)

benchmarks/bench_evolve.py
```diff
@@ -33,16 +33,12 @@ def benchmark_write_operator_card_from_file(tmp_path, test_files, test_configs):
...
-        myopcard["targetpids"]
-        myopcard["rotations"]["inputgrid"]
-        myopcard["rotations"]["targetgrid"]
-        myopcard["inputgrid"]
-        myopcard["targetgrid"]
-
+    to_check = ["inputpids", "targetpids", "inputgrid", "targetgrid"]
+    for elem in to_check:
+        with pytest.raises(KeyError):
+            myopcard["elem"]
```
felixhekhorn, line 39:
```suggestion
            myopcard[elem]
```

:upside_down_face:

AleCandido: `pytest.raises` also allows you to specify what do you expect to be contained in the error message. Add that the missing key should be actually `elem` or something like (if you were consistently quoting, it would have worked the same, but at least you should have made two consistent mistakes xD)

src/pineko/check.py
```diff
@@ -70,10 +70,13 @@ def check_grid_and_eko_compatible(pineappl_grid, operators, xif):
             "Q2 grid in pineappl grid and eko operator are NOT compatible!"
         )
     # x-grid
-    if not np.all(
-        in1d(np.unique(operators.rotations.targetgrid.tolist()), np.array(x_grid))
-    ):
-        raise ValueError("x grid in pineappl grid and eko operator are NOT compatible!")
+    # We are now checking only the shape of the operator
```
AleCandido, line 73: Do not understand why...

giacomomagni:
Hi, is it possible to merge this PR? If not what is missing? (I can contribute to finish it)

felixhekhorn:
maybe @andreab1997 can comment? but at the moment I'm not sure eko v0.11.x is usable at all (even with the fix) for pineko ... due to the output rework business ...

AleCandido:
Soon™ a new version of EKO will come (most likely v0.12), and right after with distributed computation support. Since it is getting messy, I'd wait a moment for that one.

giacomomagni:
> maybe @andreab1997 can comment? but at the moment I'm not sure eko v0.11.x is usable at all (even with the fix) for pineko ... due to the output rework business ...

What do you mean? Isn't it possible to fix it? I need to compute FK tables with `0.11`

andreab1997:
> > maybe @andreab1997 can comment? but at the moment I'm not sure eko v0.11.x is usable at all (even with the fix) for pineko ... due to the output rework business ...
> 
> What do you mean? Isn't it possible to fix ? I need to compute FK tables with `0.11`

I am Sorry but at the moment `eko==0.11` Is not usable together with pineko. I was working on the fix for a while but then we decided to do the proper job (which Is currently being done by @AleCandido and @felixhekhorn ). I will also join the effort and I Hope It won't take so much time but at the moment I am using `0.10` for scale variations

andreab1997:
I believe that most of the things here will be wrong once `eko=0.11` will be fixed. Do you agree? @AleCandido @felixhekhorn 
If yes maybe we can just close this and I will open a new one later.

felixhekhorn:
Agreed

Issue #50, opened 2022-10-15
Scaffolding


AleCandido:
At the moment, we need a given hierarchy of folders to run.

Unfortunately, this may complicate the usage for a new user (or any non-developer in general), since you have to be manage part of the structure manually, but then it has to respect some constraints.

To improve UX, I propose to implement to following features, keeping them simple, but useful:

- [x] a unique subcommand to manage a project structure, let's call it `pineko scaffold` (name to decide, take it as a placeholder)
- [x] a command to setup a new project `pineko scaffold new` (possibly aliased to `pineko new` for simplicity)
    - it should create all the directories, and possibly download some stuffs (e.g. yamldb)
- [x] a function to verify correctness of the project structure, since it should be declared in the config file `pineko.toml`, `check_folders()`
    - it should be run by every subcommand of `pineko scaffold` before running (simple to implement in `click`, we just need to put the call in the group function)
    ```py
    @command.group("scaffold")
    def subcommand():
        check_folders()
    ```
- [x] a subcommand `pineko scaffold check` to just run `check_folders()` (and print a simple and nice report if something goes wrong)

Since paths here are crucial, this is related to #51 

Optionally:
- a subcommand `pineko scaffold update` to download a new version of remote resources (e.g. yamldb)
  - in case, better to keep the old one as backup (put in a `.old-data` folder somewhere, and mangle with a timestamp)
- a subcommand `pineko scaffold repair`, to fix the structure according to the current one spelled out in `pineko.toml`

AleCandido:
> We may reconsider our choice of scoping "theory commands" and keep "file commands" global (such as `compare`)

Definitely: it looks to me that the theory mode is by now more useful than the file mode. But we can discuss :)

PR #67, opened 2022-12-06
Scope op card template in theory

### LINKED ISSUE #47: Make opcard template theory dependent
A practical limitation is the current pineko design: because pineko currently assumes a global `operator_card_template`: https://github.com/NNPDF/pineko/blob/707f21fba8d3715b68cc6ad4371229db84acdded/pineko.debug.toml#L6
- which defines the `interpolation_xgrid` which is the internal grid used to compute all operators
- but it could even hold a `inputgrid`, which instead defines the grid exposed to the fit (consider that for eko input=fitting scale)
- I guess we should change this design to make the template theory dependent, i.e. scope it by the theory id as all the other stuff

_Originally posted by @felixhekhorn in https://github.com/NNPDF/pineko/issues/42#issuecomment-1227099523_

felixhekhorn:
Closes #47 

if we agree on the strategy then I should
- [x] update the doc (here and in https://github.com/NNPDF/pineline - note that we have a slight duplication issue)
- [ ] add some tests

note that users now have to provide in addition to the external grids also the template operator card (could be also addressed in #50 )

cc @scarlehoff

### DIFF
--- a/benchmarks/bench_cli.py
+++ b/benchmarks/bench_cli.py
@@ -57,7 +57,9 @@ def benchmark_opcard_cli(tmp_path, test_files):
     grid_path = pathlib.Path(
         test_files / "data/grids/208/LHCB_DY_13TEV_DIMUON.pineappl.lz4"
     )
-    default_card_path = pathlib.Path(test_files / "data/operator_cards/_template.yaml")
+    default_card_path = pathlib.Path(
+        test_files / "data/operator_cards/208/_template.yaml"
+    )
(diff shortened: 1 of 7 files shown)


andreab1997:
Once the doc here is updated we can merge I believe

andreab1997 (approved):

andreab1997:
For the tests I believe our tests are already testing this (in fact they failed before)

andreab1997:
Can we merge this?

PR #68, opened 2022-12-09
Mathematica notebook to generate scale variation terms

andreab1997:
I am adding a mathematica notebook that we can use to generate scale variation terms for DIS and hadronic case. What it misses at the moment is the N3LO. @AleCandido @felixhekhorn


felixhekhorn:
@andreab1997 in your notebook there is too many repetition for my taste - we should predict e.g. the linear diff between alphas and pdfs; I started a new notebook, maybe you can have a look and tell me what you think? we need this to predict the N3LO stuff

AleCandido:
I perfectly agree with @felixhekhorn: in principle, there should only be the two equations we need
- $\alpha_s$ running (beta function)
- DGLAP equations
 
that are the two equations regulating the dependence on the nonphysical scales, and the perturbative solutions, with the deriving algorithm, not manually encoded.

Essentially, you want to have a place where you put your parameter to `n=2` and obtain NNLO, and `n=3` and obtain N3LO. But in principle, you should be able to put also `n=50` (even though the calculation might explode, but I don't care about `n=50`).

We need this because validating the algorithm on lower orders we can be more confident about the solution for higher orders, otherwise we gain little to nothing.

andreab1997:
> @andreab1997 in your notebook there is too many repetition for my taste - we should predict e.g. the linear diff between alphas and pdfs; I started a new notebook, maybe you can have a look and tell me what you think? we need this to predict the N3LO stuff

I a having a look right now :)

felixhekhorn:
sv is now explained in the MHOU paper together with some small MMa snippet

PR #69, opened 2022-12-15
Scaffolding

### LINKED ISSUE #50: Scaffolding

andreab1997:
Close #50 

@AleCandido @felixhekhorn


andreab1997:
I would like to rebase this on main once #67 will be merged. Do you agree?

andreab1997:
> I would like to rebase this on main once #67 will be merged. Do you agree?

Even because otherwise benchmarks will keep failing

AleCandido:
> Even because otherwise benchmarks will keep failing

But wait, they are not failing in main, so why the new command should make them failing?

andreab1997:
> > Even because otherwise benchmarks will keep failing
> 
> But wait, they are not failing in main, so why the new command should make them failing?

If you try to run now in main they will fail because I changed the benchmark files in the server in such a way the tests in #67 don't fail.

src/pineko/scaffold.py
```diff
@@ -0,0 +1,59 @@
...
+from .configs import needed_keys
+
+
+def set_up_project(configs):
+    """Set up all the folders spelled up in the configs dictionary.
+
+    Parameters
+    ----------
+    configs : dict
+        configs dictionary containing all the paths to be setted up
```
felixhekhorn, line 15:
```suggestion
        configs dictionary containing all the paths to be set up
```

src/pineko/scaffold.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+    for path in configs["paths"]:
+        if path == "root":
+            continue
+        if isinstance(configs["paths"][path], pathlib.Path):
+            if configs["paths"][path].suffix == ".yaml":
+                continue
+            os.makedirs(configs["paths"][path], exist_ok=True)
```
felixhekhorn, line 23: `pathlib` should reimplement `os` - e.g. here you can do instead https://docs.python.org/3/library/pathlib.html?highlight=pathlib#pathlib.Path.mkdir

src/pineko/scaffold.py
```diff
@@ -0,0 +1,59 @@
...
+    Parameters
+    ----------
+    configs : dict
+        configs dictionary containing all the paths to be setted up
+    """
+    for path in configs["paths"]:
+        if path == "root":
+            continue
+        if isinstance(configs["paths"][path], pathlib.Path):
+            if configs["paths"][path].suffix == ".yaml":
```
felixhekhorn, line 21: maybe https://docs.python.org/3/library/pathlib.html?highlight=pathlib#pathlib.Path.is_file instead?

src/pineko/scaffold.py
```diff
@@ -0,0 +1,59 @@
...
+                if isinstance(configs["paths"][path][log_path], pathlib.Path):
+                    os.makedirs(configs["paths"][path][log_path], exist_ok=True)
+                else:
+                    raise TypeError(f"Not recognized entry {log_path} in configs")
+        else:
+            raise TypeError(f"Not recognized entry {path} in configs")
+
+
+def check_folders(configs):
+    """Check if all the folders spelled up in configs exist.
```
felixhekhorn, line 35:
```suggestion
    """Check if all the folders spelled out in configs exist.
```

src/pineko/configs.py
```diff
@@ -11,6 +11,16 @@
 configs = {}
 "Holds loaded configurations"
 
+needed_keys = [
+    "ymldb",
+    "operator_cards",
+    "grids",
+    "operator_card_template",
```
felixhekhorn, line 18: I guess you can already update to `operator_card_template_name`

src/pineko/scaffold.py
```diff
@@ -0,0 +1,59 @@
...
+
+    Parameters
+    ----------
+    configs : dict
+        configs dictionary containing all the paths to be checked
+    """
+    for key in needed_keys:
+        if key not in configs["paths"]:
+            raise ValueError(f"{key} not found in config file")
+        if key == "operator_card_template":
```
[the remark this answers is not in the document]
andreab1997: You mean if I change `operator_card_template` to `operator_card_template_name`? Because in any case I would need the exception: `operator_card_template_name` would be something like `_template.yaml` and I don't want to check if it exists because I can't, it will be stored inside a theory folder

tests/test_configs.py
```diff
@@ -21,6 +21,7 @@ def test_enhance_paths():
     with pytest.raises(ValueError):
         pineko.configs.enhance_paths(test_configs)
     test_configs["paths"]["operator_cards"] = pathlib.Path("my/ope/cards/")
+    test_configs["paths"]["operator_card_template_name"] = "_template.yaml"
```
felixhekhorn, line 24: you should check he leaves this field alone ( a) it's sensible and b) you get another line coverage ;-) )

tests/test_configs.py
```diff
@@ -37,6 +38,7 @@ def test_default():
             "ymldb": pathlib.Path(""),
             "grids": pathlib.Path(""),
             "operator_cards": pathlib.Path("my/ope/cards/"),
+            "operator_card_template_name": "_template",
```
felixhekhorn, line 41:
```suggestion
            "operator_card_template_name": "_template.yaml",
```
let's make it correct ...

src/pineko/scaffold.py
```diff
@@ -0,0 +1,62 @@
+"""Tools related to generation and managing of a pineko project."""
+
+import os
+import pathlib
+
+from .configs import needed_keys
+
+
+def set_up_project(configs):
+    """Set up all the folders spelled up in the configs dictionary.
```
felixhekhorn, line 10:
```suggestion
    """Set up all the folders spelled out in the configs dictionary.
```

src/pineko/configs.py
```diff
@@ -47,16 +57,11 @@ def enhance_paths(configs_):
...
-        "operator_cards",
-        "grids",
-        "theory_cards",
-        "fktables",
-        "ekos",
-    ]:
+    for key in needed_keys:
         if key not in configs_["paths"]:
             raise ValueError(f"Configuration is missing a 'paths.{key}' key")
+        if key == "operator_card_template_name":
```
felixhekhorn, line 63: the string `operator_card_template_name` is spelled to many times ... can't you refactor? e.g. a list of required dirs and a list of required files?

src/pineko/scaffold.py
```diff
@@ -0,0 +1,62 @@
...
+            raise TypeError(f"Not recognized entry {path} in configs")
+
+
+def check_folders(configs):
+    """Check if all the folders spelled out in configs exist.
+
+    Parameters
+    ----------
+    configs : dict
+        configs dictionary containing all the paths to be checked
```
felixhekhorn, line 38: Please add a `Returns` section

src/pineko/cli/scaffold.py
```diff
@@ -0,0 +1,50 @@
...
+@click.option(
+    "-c",
+    "--configs",
+    "cfg",
+    default=None,
+    type=click.Path(resolve_path=True, path_type=pathlib.Path),
+    help="Explicitly specify config file (it has to be a valid TOML file).",
+)
+def scaffold_(cfg):
+    """Set the configs."""
```
felixhekhorn, line 21: well not really ... "Manage folder for assets." or something similar?

benchmarks/bench_cli.py
```diff
@@ -90,3 +90,10 @@ def benchmark_compare_cli(lhapdf_path, test_files, test_pdf):
             ["compare", str(grid_path), str(fk_path), "2", "0", "NNPDF40_nlo_as_01180"],
         )
     assert "yll left" in result.output
+
+
+def benchmark_scaffold_cli(test_files):
+    runner = CliRunner()
+    conf_file = test_files / "pineko.toml"
+    res = runner.invoke(command, ["scaffold", "-c", str(conf_file), "check"])
+    assert "All the folders are correctly configured" in res.output
```
felixhekhorn, line 99: you should run the `check` command right after; then destroy the setup and check `check` to fail

tests/test_scaffold.py
```diff
@@ -0,0 +1,80 @@
...
+    with pytest.raises(TypeError):
+        scaffold.set_up_project(wrong_fake_configs)
+    fake_configs_incomplete = construct_fake_configs_incomplete(tmp_path)
+    scaffold.set_up_project(fake_configs_incomplete)
+    assert (tmp_path / "data/ymldb").exists()
+    assert (tmp_path / "logs/eko").exists()
+
+
+def test_check_folder(tmp_path):
+    fake_configs_incomplete = construct_fake_configs_incomplete(tmp_path)
```
felixhekhorn, line 58:
```suggestion
    # we may fail because we use a wrong config ...
    fake_configs_incomplete = construct_fake_configs_incomplete(tmp_path)
```

tests/test_scaffold.py
```diff
@@ -0,0 +1,80 @@
...
+    assert success == False
+    assert [
+        "operator_cards",
+        "grids",
+        "operator_card_template_name",
+        "theory_cards",
+        "fktables",
+        "ekos",
+    ] == wrong_confs
+    fake_configs = construct_fake_configs(tmp_path)
```
felixhekhorn, line 72:
```suggestion
    # or because we didn't setup up properly ...
    fake_configs = construct_fake_configs(tmp_path)
```

tests/test_scaffold.py
```diff
@@ -0,0 +1,80 @@
...
+    success, wrong_confs, wrong_folders = scaffold.check_folders(fake_configs)
+    assert success == False
+    assert len(wrong_confs) == 0
+    for key in wrong_folders:
+        if not isinstance(wrong_folders[key], dict):
+            assert key in configs.needed_keys
+    scaffold.set_up_project(fake_configs)
```
felixhekhorn, line 79:
```suggestion
    # but if we use our function we have to be safe.
    scaffold.set_up_project(fake_configs)
```

tests/test_scaffold.py
```diff
@@ -0,0 +1,80 @@
+import pathlib
+
+import pytest
+
+from pineko import configs, scaffold
+
+
+def construct_wrong_fake_configs(tmp_path):
```
felixhekhorn, line 8: you could make those even fixtures (then, I think, you don't run in the risk of mutability)

felixhekhorn (changes requested):
more minor stuff - actual, I think, only the refactor comment above is maybe crucial

src/pineko/configs.py
andreab1997: You mean, we should have something similar to what we have for needed folders but for files, right? We can have but at the moment the needed file list would only contain one entry

AleCandido: In general, any string should appear in the package exactly once. If you use many times, assign to a constant (same for any other value).

tests/test_scaffold.py
AleCandido: You can make a session-wise fixture

AleCandido: https://docs.pytest.org/en/6.2.x/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session

src/pineko/configs.py
felixhekhorn: the list is maybe one element at this point in time, but it might grow ...

andreab1997: Ok I agree, done :)

tests/test_scaffold.py
andreab1997: Thanks, I don't think this to be exactly what I need but I managed to find a solution :)

src/pineko/scaffold.py
```diff
@@ -36,14 +36,21 @@ def check_folders(configs):
     ----------
     configs : dict
         configs dictionary containing all the paths to be checked
+    Returns
+    -------
+    : tuple(bool, list, dict)
```
felixhekhorn, line 41: please check the generated sphinx explicitly ... however, I think this should be the usual syntax:
```
success : bool
    True if the configuration is correct
missing_key : list
    ...
```

benchmarks/bench_cli.py
felixhekhorn: the important thing is to test `check` and see it succeed once and fail once

felixhekhorn (approved):

andreab1997:
At the moment this lacks only the possibility of downloading the resources and the optional commands that @AleCandido proposed (update and repair). I would say that we need to wait to implement the download of the resources because at the moment we don't even have an official place where to store them. So if you are satisfied with this I would merge and do the rest in a future PR.

AleCandido:
> I would say that we need to wait to implement the download of the resources because at the moment we don't even have an official place where to store them.

I agree, but the URL should never be fixed, also in the final version (we can provide a default, but the value should be configured to allow different sources).

> So if you are satisfied with this I would merge and do the rest in a future PR.

I just finished to debug EKO, I'll try to review the PR today :)

AleCandido:
Sorry for having taken so long: I'm now reviewing, and already the help has problems.
In particular, if no configuration is detected is raising an error, but it is a bit absurd since I'm just asking for the help (that should not require any configs).

```
❯ poe pineko scaffold new -h
Poe => pineko scaffold new -h
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/alessandro/.cache/pypoetry/virtualenvs/pineko-11j1S7bC-py3.10/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/home/alessandro/.cache/pypoetry/virtualenvs/pineko-11j1S7bC-py3.10/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/home/alessandro/.cache/pypoetry/virtualenvs/pineko-11j1S7bC-py3.10/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/alessandro/.cache/pypoetry/virtualenvs/pineko-11j1S7bC-py3.10/lib/python3.10/site-packages/click/core.py", line 1654, in invoke
    super().invoke(ctx)
  File "/home/alessandro/.cache/pypoetry/virtualenvs/pineko-11j1S7bC-py3.10/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/alessandro/.cache/pypoetry/virtualenvs/pineko-11j1S7bC-py3.10/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/mnt/cave/Projects/N3PDF/pineko/src/pineko/cli/scaffold.py", line 22, in scaffold_
    path = configs.detect(cfg)
  File "/mnt/cave/Projects/N3PDF/pineko/src/pineko/configs.py", line 125, in detect
    raise FileNotFoundError("No configurations file detected.")
FileNotFoundError: No configurations file detected.
```

This I'm pretty sure is partially my/@felixhekhorn's fault, since it is more configs related, but it might be nice to solve the problem, since we are updating the CLI.
Maybe we could do in a separate PR.

Expected behavior:
- if not asking for help (`-h/--help`, or args-less invocations) it should always read configs as the first thing
- if configs not found (but also any other foreseen possible exception), a click exception should be raised https://click.palletsprojects.com/en/8.1.x/exceptions/ (that most likely won't splat the whole traceback on the user's face)
- if asking for help, just print it without attempting to do anything else

By the way, to account for unforeseen errors (something like "configs broken in unknown way" as opposed to "configs not found) we might want to install `rich` [traceback renderer](https://rich.readthedocs.io/en/stable/traceback.html), that makes them a bit more elegant :)

src/pineko/cli/scaffold.py
```diff
@@ -0,0 +1,50 @@
...
+    "cfg",
+    default=None,
+    type=click.Path(resolve_path=True, path_type=pathlib.Path),
+    help="Explicitly specify config file (it has to be a valid TOML file).",
+)
+def scaffold_(cfg):
+    """Manage folders needed for the project as spelled out in the configuration file."""
+    path = configs.detect(cfg)
+    base_configs = configs.load(path)
+    configs.configs = configs.defaults(base_configs)
```
AleCandido, line 24: Ah wait, the problem *is* here, since this code is run always, also when the user is just asking for help.

I propose this should be solved in this PR, at least for this subcommand.

AleCandido (changes requested):

andreab1997:
> Sorry for having taken so long: I'm now reviewing, and already the help has problems. In particular, if no configuration is detected is raising an error, but it is a bit absurd since I'm just asking for the help (that should not require any configs).
> 
> ```
> ❯ poe pineko scaffold new -h
> Poe => pineko scaffold new -h
> Traceback (most recent call last):
>   File "<string>", line 1, in <module>
>   File "/home/alessandro/.cache/pypoetry/virtualenvs/pineko-11j1S7bC-py3.10/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
>     return self.main(*args, **kwargs)
>   File "/home/alessandro/.cache/pypoetry/virtualenvs/pineko-11j1S7bC-py3.10/lib/python3.10/site-packages/click/core.py", line 1055, in main
>     rv = self.invoke(ctx)
>   File "/home/alessandro/.cache/pypoetry/virtualenvs/pineko-11j1S7bC-py3.10/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
>     return _process_result(sub_ctx.command.invoke(sub_ctx))
>   File "/home/alessandro/.cache/pypoetry/virtualenvs/pineko-11j1S7bC-py3.10/lib/python3.10/site-packages/click/core.py", line 1654, in invoke
>     super().invoke(ctx)
>   File "/home/alessandro/.cache/pypoetry/virtualenvs/pineko-11j1S7bC-py3.10/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
>     return ctx.invoke(self.callback, **ctx.params)
>   File "/home/alessandro/.cache/pypoetry/virtualenvs/pineko-11j1S7bC-py3.10/lib/python3.10/site-packages/click/core.py", line 760, in invoke
>     return __callback(*args, **kwargs)
>   File "/mnt/cave/Projects/N3PDF/pineko/src/pineko/cli/scaffold.py", line 22, in scaffold_
>     path = configs.detect(cfg)
>   File "/mnt/cave/Projects/N3PDF/pineko/src/pineko/configs.py", line 125, in detect
>     raise FileNotFoundError("No configurations file detected.")
> FileNotFoundError: No configurations file detected.
> ```
> 
> This I'm pretty sure is partially my/@felixhekhorn's fault, since it is more configs related, but it might be nice to solve the problem, since we are updating the CLI. Maybe we could do in a separate PR.
> 
> Expected behavior:
> 
> * if not asking for help (`-h/--help`, or args-less invocations) it should always read configs as the first thing
> * if configs not found (but also any other foreseen possible exception), a click exception should be raised https://click.palletsprojects.com/en/8.1.x/exceptions/ (that most likely won't splat the whole traceback on the user's face)
> * if asking for help, just print it without attempting to do anything else
> 
> By the way, to account for unforeseen errors (something like "configs broken in unknown way" as opposed to "configs not found) we might want to install `rich` [traceback renderer](https://rich.readthedocs.io/en/stable/traceback.html) (only in the CLI, similar to the installation of the logging handler), that makes them a bit more elegant :)

Thanks you for noticing this problem. I agree with you, this is a more general problem and we probably want to solve it in a separate PR. In fact you get exactly the same error if you type `pineko theory ekos -h` for example. The point is that if you type just `pineko scaffold -h` which is the parent command it would work but if you use one of the subcommands, it will try to load the configurations (eventually failing). The same is true for the the `theory` command.

src/pineko/cli/scaffold.py
andreab1997: Yes but, as I was saying in the other comment, this is exactly the same of the `theory` command so, whatever solution we want to apply here, we need to copy it also there.

AleCandido: Yep, that is fine, but at least we close this PR with a working command. Then we open a new one for fixing all the others.
However, if it becomes urgent we can release without (but at the moment is not terribly urgent)

andreab1997: No, I agree to solve it here Anyway I am not sure how. We for sure want the parent command to set the configs. Do we want to use a `try, except`? Or do you have any suggestions?

AleCandido: I'm looking for a solution, but I do not have an optimal one. The best would have been that it were already doing on its own, as it is for the parent command, but I don't know why it isn't...
Have a quick look online, I'm doing as well. In any case, if possible we should try to detect an help request (i.e. when `click` would decide to print help or a prompt for missing arguments/wrong CLI invocation) and avoid execution at all, but otherwise any working solution might be fine for the time being.

AleCandido: https://github.com/pallets/click/issues/295

AleCandido: Ok, it is complicated: here there is a rather generic (and lengthy) solution https://github.com/pallets/click/issues/295#issuecomment-611648993

I'd try to keep it smaller, but I'm failing for the time being:
```py
import click

CONTEXT_SETTINGS = dict(help_option_names=["-h", "--help"])


@click.group(context_settings=CONTEXT_SETTINGS)
@click.option("-h", "--help", is_flag=True)
def command(help):
    """Base command"""
    print("This should not bee seen if asking for help.")


@command.command()
@click.argument("ciao")
def sub(ciao):
    """Subcommand"""
    print(f"Ciao: {ciao} (but not during help)")
```

While this is working:
```py
import sys

import click

CONTEXT = {"help_option_names": ["-h", "--help"]}


def print_cmd_help(ctx, param, value):
    group = ctx.command
    # parse the cmdline and get the command and its arguments
    parser = group.make_parser(ctx)
    global_opts, args, _ = parser.parse_args(args=sys.argv[1:])
    if global_opts.get("help") is True or not args:
        # global help wanted
        click.echo(ctx.get_help())
        ctx.exit()

    # create the command object manually and parse its arguments
    name, cmd, args = group.resolve_command(ctx, args)
    help_names = cmd.get_help_option_names(ctx)

    if set(args) & set(help_names) or not args:
        # return command help when a user wants it or when he does not provide
        # arguments
        # taken from https://github.com/pallets/click/blob/7.1.1/src/click/core.py#L597
        cmd_ctx = click.Context(cmd, info_name=cmd.name, parent=ctx)
        click.echo(cmd_ctx.get_help())
        cmd_ctx.exit()


@click.group(context_settings=CONTEXT)
@click.option(
    "-h",
    "--help",
    is_flag=True,
    callback=print_cmd_help,
    expose_value=False,
    is_eager=True,
)
def command():
    print("This should not bee seen if asking for help.")


@command.command(help="this is the help of test")
@click.argument("ciao")
def sub(ciao):
    print(f"Ciao: {ciao} (but not during help)")
```

AleCandido: I propose to take the second one, put the function in a `library.py` module, and add there an alias to the full decorator as well, such that we can simply add `@library.help` to groups.

We should also improve a bit, since e.g. the `Context` construction has been updated in `click.core` as well https://github.com/pallets/click/blob/5e75fe7f9c1b820d71f482599cd6996523332d06/src/click/core.py#L744

AleCandido: Other pain point is that I'm not sure that `1` in `parser.parse_args(args=sys.argv[1:])` is always correct (you could have a nested subgroup).
At the moment I'd just leave a `TODO:` comment there, since I don't know how to detect how much nested is the current command (it might be also just fine, if `click` manipulates the `sys.argv` list, and the `1` thus is the correct choice...)

AleCandido: Further pain point is if we have a subcommand that should actually be invoked without any argument, in which case the second `not args` check would be wrong.

andreab1997: Yes, for instance `scaffold new` is called without arguments and it is exaclty one of the commands raising the error.

benchmarks/conftest.py
```diff
@@ -39,3 +45,18 @@ def wrapped(newdir):
             lhapdf.setPaths(paths)
 
     return wrapped
+
+
+@pytest.fixture
+def cd():
```
AleCandido, line 51: It seems to me that you're not using this one, correct?

src/pineko/scaffold.py
```diff
@@ -0,0 +1,71 @@
...
+        configs dictionary containing all the paths to be checked
+    Returns
+    -------
+    : bool
+        True if the configuration is correct, False otherwise
+    : list
+        list of missing keys in the configuration file
+    : dict
+        dictionary containing folders that should exists but could
+        not be found
```
AleCandido, line 47: It is simpler to write a very brief dataclass, let's say `CheckResult`, and return this one. Something like:
```py
@dataclasses.dataclass
class CheckResult:
    confs: List[pathlib.Path]
    folders: Dict[str, pathlib.Path]
```

src/pineko/scaffold.py
```diff
@@ -0,0 +1,71 @@
...
+    else:
+        wrong_folders["logs"] = {}
+        for key in configs["paths"]["logs"]:
+            if not configs["paths"]["logs"][key].exists():
+                wrong_folders["logs"][key] = configs["paths"]["logs"][key]
+    success = (
+        True
+        if (len(wrong_confs) == 0 and list(wrong_folders.keys()) == ["logs"])
+        else False
+    )
```
AleCandido, line 70: You are using an `if` to construct a boolean from a condition (that is already a boolean)
```suggestion
    success = len(wrong_confs) == 0 and list(wrong_folders.keys()) == ["logs"]
```

src/pineko/scaffold.py
```diff
@@ -0,0 +1,71 @@
...
+            if not configs["paths"][key].exists():
+                wrong_folders[key] = configs["paths"][key]
+    if "logs" not in configs["paths"]:
+        print("WARNING: logs folder is not spelled out in the config file")
+    else:
+        wrong_folders["logs"] = {}
+        for key in configs["paths"]["logs"]:
+            if not configs["paths"]["logs"][key].exists():
+                wrong_folders["logs"][key] = configs["paths"]["logs"][key]
+    success = (
```
AleCandido, line 66: This can be a property of the result dataclass, since it is just a function of `wrong_confs` and `wrong_folders`

tests/conftest.py
```diff
@@ -0,0 +1,48 @@
+import pytest
+
+
+@pytest.fixture
+def my_tmp_path(tmp_path):
+    return tmp_path
```
AleCandido, line 6: This fixture is not doing anything...

tests/conftest.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+@pytest.fixture
+def construct_wrong_fake_configs(my_tmp_path):
```
AleCandido, line 10: Add a brief comment on why is wrong (put it in a docstring)

tests/conftest.py
```diff
@@ -0,0 +1,48 @@
...
+            "ymldb": my_tmp_path / "data/ymldb",
+            "logs": {"eko": my_tmp_path / "logs/eko", "fk": ["something", "wrong"]},
+        },
+        "root": my_tmp_path,
+    }
+    return wrong_fake_configs
+
+
+@pytest.fixture
+def construct_fake_configs_incomplete(my_tmp_path):
```
AleCandido, line 22: Add a brief comment on why is incomplete (put it in a docstring)

tests/test_scaffold.py
```diff
@@ -0,0 +1,46 @@
...
+from pineko import configs, scaffold
+
+
+def test_set_up_project(
+    my_tmp_path, construct_wrong_fake_configs, construct_fake_configs_incomplete
+):
+    wrong_fake_configs = construct_wrong_fake_configs
+    with pytest.raises(TypeError):
+        scaffold.set_up_project(wrong_fake_configs)
+    fake_configs_incomplete = construct_fake_configs_incomplete
```
AleCandido, line 14: Why do you need this alias?

tests/test_scaffold.py
```diff
@@ -0,0 +1,46 @@
...
+    scaffold.set_up_project(fake_configs_incomplete)
+    assert (my_tmp_path / "data/ymldb").exists()
+    assert (my_tmp_path / "logs/eko").exists()
+
+
+def test_check_folder(construct_fake_configs_incomplete, construct_fake_configs):
+    # we may fail because we use a wrong config ...
+    fake_configs_incomplete = construct_fake_configs_incomplete
```
AleCandido, line 22: Why do you need this alias?

tests/conftest.py
```diff
@@ -0,0 +1,48 @@
...
+            "ymldb": my_tmp_path / "data/ymldb",
+            "logs": {"eko": my_tmp_path / "logs/eko"},
+        },
+        "root": my_tmp_path,
+    }
+    return fake_configs_incomplete
+
+
+@pytest.fixture
+def construct_fake_configs(my_tmp_path):
```
AleCandido, line 34: Here and in those above: in a fixture the action is implicit, also because the function is invoked and replaced with the result (e.g. PyTest provides `tmp_path`, not `construct_tmp_path`).
It is different for factory fixtures (that return functions), but these are not factories https://docs.pytest.org/en/7.2.x/how-to/fixtures.html#factories-as-fixtures

tests/test_scaffold.py
AleCandido: Maybe it is an artifact caused by the name choice above.

tests/test_scaffold.py
```diff
@@ -0,0 +1,46 @@
...
+    assert [
+        "operator_cards",
+        "grids",
+        "operator_card_template_name",
+        "theory_cards",
+        "fktables",
+        "ekos",
+    ] == wrong_confs
+    # or because we didn't setup up properly ...
+    fake_configs = construct_fake_configs
```
AleCandido, line 37: Also this is just an alias

tests/conftest.py
```diff
@@ -0,0 +1,48 @@
...
+@pytest.fixture
+def my_tmp_path(tmp_path):
+    return tmp_path
+
+
+@pytest.fixture
+def construct_wrong_fake_configs(my_tmp_path):
+    wrong_fake_configs = {
+        "paths": {
+            "ymldb": my_tmp_path / "data/ymldb",
```
AleCandido, line 13: Here and everywhere else: the `/` operator of paths is there also for portability, since it works on all supported systems, but the character `"/"` does not.
So, specify each piece of the path on its own:
```suggestion
            "ymldb": my_tmp_path / "data" / "ymldb",
```

src/pineko/configs.py
```diff
@@ -11,6 +11,18 @@
 configs = {}
 "Holds loaded configurations"
 
+needed_keys = [
```
AleCandido, line 14: if this is supposed to be a constant make it upper case
```suggestion
NEEDED_KEYS = [
```

src/pineko/configs.py
```diff
@@ -11,6 +11,18 @@
...
+    "ymldb",
+    "operator_cards",
+    "grids",
+    "operator_card_template_name",
+    "theory_cards",
+    "fktables",
+    "ekos",
+]
+
+needed_files = ["operator_card_template_name"]
```
AleCandido, line 24: same as above:
```suggestion
NEEDED_FILES = ["operator_card_template_name"]
```
(Python has little to no actual types, so it relies on conventions)

src/pineko/cli/scaffold.py
```diff
@@ -0,0 +1,50 @@
...
+@scaffold_.command()
+def check():
+    """Check if all the configurations are correct."""
+    success, wrong_confs, wrong_folders = scaffold.check_folders(configs.configs)
+    if success:
+        rich.print("[green]Success:[/] All the folders are correctly configured.")
+    else:
+        rich.print("[red]Error:[/] Project is not correctly configured.")
+        for conf in wrong_confs:
+            rich.print(f"[red]Missing entry in conf file: \t{conf}")
```
AleCandido, line 44: To make it more explicit that is a keyword, prefer quotes over tab
```suggestion
            rich.print(f"[red]Missing entry in conf file: '{conf}'")
```
(I've just seen it more frequently used, so it should be clearer for users)

src/pineko/cli/scaffold.py
```diff
@@ -0,0 +1,50 @@
...
+        for key in wrong_folders:
+            if not isinstance(wrong_folders[key], dict):
+                rich.print(f"[red]Missing folder: \t{wrong_folders[key]}")
```
AleCandido, line 47: Since this might be an object, it might have a lengthy representation, so you may want to print it on a separate line.
```suggestion
                rich.print(f"[red]Missing folder:\n{wrong_folders[key]}")
```

src/pineko/cli/scaffold.py
```diff
@@ -0,0 +1,50 @@
...
+            else:
+                for log_key in wrong_folders[key]:
+                    rich.print(f"[red]Missing folder: \t{wrong_folders[key][log_key]}")
```
AleCandido, line 50: About this one I'm not sure, apply one of the two suggestions above

src/pineko/cli/scaffold.py
```diff
@@ -0,0 +1,50 @@
...
+def check():
+    """Check if all the configurations are correct."""
+    success, wrong_confs, wrong_folders = scaffold.check_folders(configs.configs)
+    if success:
+        rich.print("[green]Success:[/] All the folders are correctly configured.")
+    else:
+        rich.print("[red]Error:[/] Project is not correctly configured.")
+        for conf in wrong_confs:
+            rich.print(f"[red]Missing entry in conf file: \t{conf}")
+        for key in wrong_folders:
```
AleCandido, line 45: Since you are accessing over and over the value, please use the `.items()` iterator:
```suggestion
        for key, folder in wrong_folders.items():
```

AleCandido (changes requested):

tests/conftest.py
andreab1997: It is. I need it because I want the tmp_path given to the other fixtures and to the test itself to be the same. This solution seems to me the cleanest.

tests/test_scaffold.py
andreab1997: I actually don't need it :)

tests/conftest.py
AleCandido: I see your point, but, in order to improve clarity, I'd do the following three things:
1. the easiest: comment why you are defining it in a docstring
2. the most complicate: choose a telling name (e.g. `shared_tmp_path`, but it is still not optimal, according to me)

However, it seems weird to me that you need it, since it should be the same out-of-the-box:
- *This means we can request fixtures in multiple fixtures that are dependent on them (and even again in the test itself) without those fixtures being executed more than once.* https://docs.pytest.org/en/7.2.x/how-to/fixtures.html#fixtures-can-be-requested-more-than-once-per-test-return-values-are-cached
- your fixture is really doing nothing, since if `tmp_path` invoked multiple times were yielding different values, I see no reason why `my_tmp_path` should yield a single one: if multiple values per function are allowed, then multiple `my_tmp_path` invocations would result in multiple `tmp_path` invocations

In general, you can achieve what you want by using a [scope](https://docs.pytest.org/en/7.2.x/how-to/fixtures.html#scope-sharing-fixtures-across-classes-modules-packages-or-session), and to declare a `"module"` or `"class"` scope you would require what you put there (plus the `scope` kwarg, of course). But `"function"` is already the default scope.

AleCandido: In general, if you need multiple values per-function, you should pick a factory instead.

src/pineko/scaffold.py
```diff
@@ -0,0 +1,67 @@
...
+
+def set_up_project(configs):
+    """Set up all the folders spelled out in the configs dictionary.
+
+    Parameters
+    ----------
+    configs : dict
+        configs dictionary containing all the paths to be set up
+    """
+    for path in configs["paths"]:
```
AleCandido, line 27: Here as well, you are using `configs["paths"][path]` many times.
For readability (but in general also for performance) use the `.items()` iterator (i.e. assign it to a variable, but this way is more idiomatic).

AleCandido (changes requested):

src/pineko/scaffold.py
```diff
@@ -0,0 +1,82 @@
...
+        else:
+            if key in NEEDED_FILES:
+                continue
+            if not configs["paths"][key].exists():
+                wrong_folders[key] = configs["paths"][key]
+    if "logs" not in configs["paths"]:
+        print("WARNING: logs folder is not spelled out in the config file")
+    else:
+        wrong_folders["logs"] = {}
+        for key in configs["paths"]["logs"]:
```
AleCandido, line 79: I know I start being annoying, but since I already notified many of these...

AleCandido (changes requested):

src/pineko/scaffold.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    for path_key, path in configs["paths"].items():
+        if path_key == "root" or path_key in NEEDED_FILES:
+            continue
+        if isinstance(path, pathlib.Path):
+            path.mkdir(parents=True, exist_ok=True)
+        elif isinstance(path, dict):
+            for log_path in path:
+                if isinstance(path[log_path], pathlib.Path):
+                    os.makedirs(path[log_path], exist_ok=True)
```
felixhekhorn, line 46: move to `mkdir` as above

tests/conftest.py
felixhekhorn: actually, in this case you could also just use different subdirectories inside `tmp_path`

AleCandido: @felixhekhorn why? `tmp_path` is already different for each test

andreab1997:
@AleCandido @felixhekhorn If you are ok with this, I would merge and then open an issue for the problem with the cli (in any case it is a problem also for the `theory` command so even solving here is not enough. And we don't have a proper solution yet). Do you agree?

AleCandido:
> @AleCandido @felixhekhorn If you are ok with this, I would merge and then open an issue for the problem with the cli (in any case it is a problem also for the `theory` command so even solving here is not enough. And we don't have a proper solution yet). Do you agree?

image

There are other two conversations opened (other than the CLI):
- solve those
- open the issue, and reference here
- only then merge

andreab1997:
The tests are now failing just because I updated the benchmarks files for #73 but here the tests are not yet updated. Should I rebase?

andreab1997:
Note that the problem with the CLI that has been discovered here is now explained here #74

scarlehoff:
> The tests are now failing just because I updated the benchmarks files for #73 but here the tests are not yet updated. Should I rebase?

Yes, please.

If understood correctly from @AleCandido msg yesterday this is ready to merge right?

AleCandido:
> If understood correctly from @AleCandido msg yesterday this is ready to merge right?

Just waiting on an EKO release, sorry ^^"

src/pineko/cli/scaffold.py
AleCandido: Moved into its own issue #78

andreab1997:
@felixhekhorn @AleCandido I am asking again your review but everything should already be ok. I have rebased on top of main and tests are now passing. Let me know if we can merge this. :)

AleCandido:
I will have a look today

andreab1997:
Can we merge this? (Or do you want to review it again?)

src/pineko/evolve.py
```diff
@@ -5,32 +5,50 @@
...
+def sv_scheme(tcard):
+    """Infere the factorization scale_variation scheme to be used from the theory card.
+
+    Parameters
+    ----------
+    tcard : dict
+        theory card
+
+    """
+    modsv_list = ["expanded", "exponentiated"]
```
AleCandido, line 28: https://github.com/NNPDF/eko/blob/873dc2d3cf68c12058db80f067e5d4ef700fcf2c/src/eko/io/types.py#L160-L164

Do not repeat existing values, this is:
```suggestion
    modsv_list = {a.value for a in ScaleVariationsMethod}
```
(and I put braces since you need a set, not a list)

src/pineko/evolve.py
```diff
@@ -5,32 +5,50 @@
...
+    xif = tcard["XIF"]
+    modsv = tcard["ModSV"]
+    if np.isclose(xif, 1.0):
+        if modsv in modsv_list:
+            raise ValueError("ModSv is not None but xif is 1.0")
+        else:
+            return "None"
```
AleCandido, line 35:
```suggestion
            return None
```

src/pineko/evolve.py
```diff
@@ -5,32 +5,50 @@
...
+    else:
+        # scheme C case
+        if modsv not in modsv_list:
+            return "None"
```
AleCandido, line 39:
```suggestion
            return None
```

AleCandido (approved):
This PR has become large for some obscure reason, and it is not only doing scaffolding, since it touches many other files.

Moreover, there are some quite lenghty and not very clean implementations (e.g. `evolve.evolve_grid`, `evolve.write_operator_card`, and `theory.fk`) partially resulting from previously accumulated technical debt.

We should soon stop and refactor the code in Pineko, since it has grown and it really changed size. However, if this PR is required for something and it is working, we can also merge it and postpone the refactor to a later (but not too late) stage

andreab1997:
> This PR has become large for some obscure reason, and it is not only doing scaffolding, since it touches many other files.
> 

I don't think the reason to be obscure, it has just been rebased on top of main after that the eko compatibility PR was merged.

felixhekhorn (approved):
Let's close this asap

AleCandido:
> I don't think the reason to be obscure, it has just been rebased on top of main after that the eko compatibility PR was merged.

Since this is pointing on `main`, if it were rebased on `main` the changes should have not appeared in the diff. They should have been in both branches.
That's why I'm calling it _obscure_, since it is not clear to me.

PR #71, opened 2022-12-16
Fix evolve result metadata

### LINKED ISSUE #70: Missing metadata?
The following lines look a bit fishy:

https://github.com/NNPDF/pineko/blob/60bb3440a97bc9625b7a99e661fe227188398deb/src/pineko/evolve.py#L186-L194

You add two key-value-pairs, but the FK table is written before, and that's the only call to `write_lz4` in the project. That probably means that the metadata is never propagated to the file on disk.

felixhekhorn:
Closes #70 

The added assert fails in current master but passes here

thanks to @cschwan


andreab1997:
Oh I see. So is this the actual problem for what @scarlehoff found?

PR #73, opened 2023-01-13
Eko 0.12 compatibility 

andreab1997:
Here we should fix what is needed to make pineko work with the new eko implementation (in particular with the develop version). @felixhekhorn @AleCandido 

- [x] restore EKO dep to semver


AleCandido:
I would say just try to run, and you will find broken things. Hopefully not that many.

In particular, you can learn how to use the new methods from the EKO unit tests. In particular, an EKO is now opened with `EKO.open(..., "r"/"w"/"a")`, or one of the aliases `EKO.read(...)`/`EKO.write(...)`/`EKO.edit(...)`.

andreab1997:
> I would say just try to run, and you will find broken things. Hopefully not that many.
> 
> In particular, you can learn how to use the new methods from the EKO unit tests. In particular, an EKO is now opened with `EKO.open(..., "r"/"w"/"a")`, or one of the aliases `EKO.read(...)`/`EKO.write(...)`/`EKO.edit(...)`.

Ok so as far as I understood I don't need anymore the legacy part: `EKO` will take care of loading the legacy ekos automatically if needed.

AleCandido:
> Ok so as far as I understood I don't need anymore the legacy part: `EKO` will take care of loading the legacy ekos automatically if needed.

Not really, from that point of view nothing should have changed...

AleCandido:
I temporarily moved the EKO dependency from the usual semver one to Git, such that unit tests are meaningful, please remember to restore the usual before merging.

AleCandido:
Here there is a good point where to start from:
https://github.com/NNPDF/pineko/actions/runs/3911199573/jobs/6684311767#step:10:13

src/pineko/theory.py
```diff
@@ -224,7 +224,7 @@ def opcard(self, name, grid, xif, tcard_path):
                 return
         _x_grid, q2_grid = evolve.write_operator_card_from_file(
             grid,
-            self.operator_cards_path / configs.configs["paths"]["operator_card_template_name"],
+            configs.configs["paths"]["operator_card_template"],
```
felixhekhorn, line 227: huh? why did you revert to the old style?

andreab1997: Because scaffolding is not yet merged so I would say let's keep the old style and then let's change once scaffolding is merged. Of course I can even do it now

src/pineko/cli/convolute.py
```diff
@@ -51,15 +50,15 @@ def subcommand(
     PDF is an optional PDF set compatible with the EKO to compare grid and FK table.
     """
     grid = pineappl.grid.Grid.read(grid_path)
-    operators = eko.io.legacy.load_tar(op_path)
+    operators = eko.EKO.edit(op_path)
```
AleCandido, line 53: Use the context manager!

AleCandido (changes requested):

src/pineko/cli/convolute.py
```diff
@@ -51,13 +50,26 @@ def subcommand(
...
         theory_card = yaml.safe_load(f)
-    new_tcard = eko.compatibility.update_theory(theory_card)
-    astrong = sc.Couplings.from_dict(new_tcard)
+    legacy_class = eko.io.runcards.Legacy(theory_card, {})
+    new_tcard = legacy_class.new_theory
+    evmod = eko.io.types.CouplingEvolutionMethod.EXACT
+    if theory_card["ModEv"] == "TRN":
+        evmod = eko.io.types.CouplingEvolutionMethod.EXPANDED
+    quark_masses = [(x.value) ** 2 for x in new_tcard.quark_masses]
+    sc = eko.couplings.Couplings(
```
felixhekhorn, line 64: you should move this stuff down to evolve else it is repeated here and in theory - because now we can move it down, because now the EKO object contains everything - same holds for the preparation of operators (see also Alessandro's comment above)

src/pineko/evolve.py
```diff
@@ -145,13 +132,29 @@ def evolve_grid(
...
     order_mask = pineappl.grid.Order.create_mask(grid.orders(), max_as, max_al)
     muf2_grid = operators.mu2grid
     # PineAPPL wants alpha_s = 4*pi*a_s
     # remember that we already accounted for xif in the opcard generation
+    # legacy_class = eko.io.runcards.Legacy(operators.theory_card, {})
+    # new_tcard = legacy_class.new_theory
+    tcard = operators.theory_card
+    opcard = operators.operator_card
+    evmod = eko.io.types.CouplingEvolutionMethod.EXACT
+    if opcard.configs.evolution_method is EvolutionMethod.TRUNCATED:
```
felixhekhorn, line 147: Please use https://github.com/NNPDF/eko/blob/e1bd2ae470695d27df641e5c9211699898f364cb/src/eko/couplings.py#L32

src/pineko/evolve.py
```diff
@@ -145,13 +132,29 @@ def evolve_grid(
...
-    elif not np.allclose(operators.rotations.inputpids, br.evol_basis_pids):
+    elif not np.allclose(operators.rotations.inputpids, br.rotate_flavor_to_evolution):
         raise ValueError("The EKO is neither in flavor nor in evolution basis.")
     # do it
     order_mask = pineappl.grid.Order.create_mask(grid.orders(), max_as, max_al)
     muf2_grid = operators.mu2grid
     # PineAPPL wants alpha_s = 4*pi*a_s
     # remember that we already accounted for xif in the opcard generation
+    # legacy_class = eko.io.runcards.Legacy(operators.theory_card, {})
+    # new_tcard = legacy_class.new_theory
```
felixhekhorn, line 143: Remember to remove void code before merging :upside_down_face:

andreab1997: Yes, I still need to polish everything :)

src/pineko/theory.py
```diff
@@ -389,49 +385,38 @@ def fk(self, name, grid_path, tcard, pdf):
...
-            4.0 * np.pi * astrong.a_s(xir * xir * muf2 / xif / xif)
-            for muf2 in muf2_grid
-        ]
-        # Obtain the assumptions hash
-        assumptions = theory_card.construct_assumptions(tcard)
-        # do it!
-        logger.info("Start computation of %s", name)
-        logger.info("max_as=%d, max_al=%d, xir=%f, xif=%f", max_as, max_al, xir, xif)
-        start_time = time.perf_counter()
+        with eko.EKO.edit(eko_filename) as operators:
```
felixhekhorn, line 389: Actually, why do we need `edit` here? can't we use just `open` and rotate every time? (Remember that `edit` allows to write back to disk)

andreab1997:
@AleCandido @felixhekhorn Just a question: [here](https://github.com/NNPDF/pineko/blob/5e7c272aabf85adb11aa8f8463759f1bdb69588f/src/pineko/evolve.py#L161) I am using the argument `fact_scale` and I am setting it to muf2 (which at this point is exactly the factorization scale for scheme C but it is not for scheme B). In the past we never used that argument. What should we do? Because it is not clear to me how this can be correct for both scheme B and C (and even if I need to use that argument at all).

andreab1997:
Anyway I am asking again your review because now everything should work and most of the corners have been polished. I am adding also @giacomomagni since we did part of this together.

AleCandido:
> What should we do? Because it is not clear to me how this can be correct for both scheme B and C (and even if I need to use that argument at all

We might consider to change name to it, because that argument is only used to computed the additional displacement terms used in scheme A, cf.
https://github.com/NNPDF/eko/blob/e3466fae35fd245c7e1a99158d8582f590276555/src/eko/couplings.py#L664-L666

src/pineko/evolve.py
```diff
@@ -158,18 +175,46 @@ def evolve_grid(
...
+        sv_method = operators.operator_card.configs.scvar_method.name
+    # To compute the alphas values we are first reverting the factorization scale shift
+    # and then obtaining the renormalization scale using xir.
+    alphas_values = [
+        4.0
+        * np.pi
+        * sc.a_s(
+            xir * xir * muf2 / xif / xif,
+            fact_scale=muf2 if sv_method == "EXPONENTIATED" else None,
+        )
```
giacomomagni, line 214: Just to mention that here we should pass also `nf_to`. This most likely will make no differences, but you never know in the future... however it requires having and `nf_grid` as well

andreab1997:
I believe the best strategy is to merge this as soon as possible and then rebase #69 in order to make its tests working again. So, do you believe there is something blocking here?

AleCandido:
> I believe the best strategy is to merge this as soon as possible and then rebase #69 in order to make its tests working again. So, do you believe there is something blocking here?

I agree on the strategy, unfortunately, we are blocked by:
https://github.com/NNPDF/pineko/blob/ac2898b4e202336b7c524657adf2140674917199/pyproject.toml#L27

I.e. https://github.com/NNPDF/eko/pull/190

andreab1997:
> I agree on the strategy, unfortunately, we are blocked by:
> 
> https://github.com/NNPDF/pineko/blob/ac2898b4e202336b7c524657adf2140674917199/pyproject.toml#L27
> 
> I.e. [NNPDF/eko#190](https://github.com/NNPDF/eko/pull/190)

Wait but I don't understand why that is blocking this PR

AleCandido:
> Wait but I don't understand why that is blocking this PR

Because we do not want to depend on a Git branch, it is not reproducible: `develop` is not a fixed point in time, is a moving target.
If you release this, and at a later stage we fix a bug in EKO, you will never be able to reproduce your results.

EKO is just missing tests, and without them is not fully reliable. If anyone is blocked because of that, it should consider contribute to testing EKO.

andreab1997:
> > Wait but I don't understand why that is blocking this PR
> 
> Because we do not want to depend on a Git branch, it is not reproducible: `develop` is not a fixed point in time, is a moving target. If you release this, and at a later stage we fix a bug in EKO, you will never be able to reproduce your results.
> 
> EKO is just missing tests, and without them is not fully reliable. If anyone is blocked because of that, it should consider contributing to testing EKO.

Nono I understand. But I don't want to realease this, just merge to main

AleCandido:
> Nono I understand. But I don't want to realease this, just merge to main

In principle that might be fine, however this means that we will not be able to make any other release.
So, if we are still using somewhere the older version of EKO, we should not merge immediately this. Moreover, this branch should not be used to produce immediately results, for as long as we are not confident that EKO is alright.
And we already know it is not, since @felixhekhorn already found a [bug](https://github.com/NNPDF/eko/pull/199/commits/d671cc0dfa3ed71b2497e126422d96da674dd043), so we should at least wait for https://github.com/NNPDF/eko/pull/199

TL;DR: we need this branch, but we need more an EKO tested release -> this sets the priorities

andreab1997:
> > Nono I understand. But I don't want to realease this, just merge to main
> 
> In principle that might be fine, however this means that we will not be able to make any other release. So, if we are still using somewhere the older version of EKO, we should not merge immediately this. Moreover, this branch should not be used to produce immediately results, for as long as we are not confident that EKO is alright. And we already know it is not, since @felixhekhorn already found a [bug](https://github.com/NNPDF/eko/pull/199/commits/d671cc0dfa3ed71b2497e126422d96da674dd043), so we should at least wait for [NNPDF/eko#199](https://github.com/NNPDF/eko/pull/199)
> 
> TL;DR: we need this branch, but we need more an EKO tested release -> this sets the priorities

Ok I agree. Then, since I cannot compute stuff as long as eko is not tested, I will help in testing it :)

src/pineko/evolve.py
```diff
@@ -148,28 +151,60 @@ def evolve_grid(
...
     xif : float
         factorization scale variation
-    alphas_values : None or list
-        values of strong coupling used to collapse grids
     assumptions : str
         assumptions on the flavor dimension
     comparison_pdf : None or str
         if given, a comparison table (with / without evolution) will be printed
     """
     _x_grid, _pids, mur2_grid, _muf2_grid = grid.axes()
```
felixhekhorn, line 159:
```suggestion
    x_grid, _pids, mur2_grid, _muf2_grid = grid.axes()
```
and accordingly below (now it is used)

andreab1997:
Ok now that the tests are succeding and the dependecy to eko has been restored, can we merge this?

andreab1997:
Ok nevermind, it seems that in eko something else is changed and the tests are failing again...

AleCandido:
> Ok nevermind, it seems that in eko something else is changed and the tests are failing again...

The problem is clearly with [`polarized`](https://github.com/NNPDF/pineko/actions/runs/4112825192/jobs/7098214764#step:11:497)

AleCandido:
I would revert 72f6cf9: notice that tests were actually passing before, so they have been broken by it.

The problem was only in the benchmarks, and just concerning `polarized`.

andreab1997:
> Details

No, note that the tests are still passing, the problem is the usual with codecov

AleCandido:
> No, note that the tests are still passing, the problem is the usual with codecov

Ok, then I change my statement: that commit simply didn't change the status. Benchmarks are still failing because of [`polarized`](https://github.com/NNPDF/pineko/actions/runs/4113508649/jobs/7099715336#step:11:498)

andreab1997:
> > No, note that the tests are still passing, the problem is the usual with codecov
> 
> Ok, then I change my statement: that commit simply didn't change the status. Benchmarks are still failing because of [`polarized`](https://github.com/NNPDF/pineko/actions/runs/4113508649/jobs/7099715336#step:11:498)

I know, I fixed it now :)

src/pineko/evolve.py
```diff
@@ -148,28 +151,60 @@ def evolve_grid(
...
+    evmod = eko.couplings.couplings_mod_ev(opcard.configs.evolution_method)
+    # Couplings ask for the square of the masses
+    quark_masses = [(x.value) ** 2 for x in tcard.quark_masses]
+    sc = eko.couplings.Couplings(
+        tcard.couplings,
+        tcard.order,
+        evmod,
+        quark_masses,
+        hqm_scheme=tcard.quark_masses_scheme,
+        thresholds_ratios=tuple(tcard.matching.raw.values()),
```
felixhekhorn, line 192: no! this *is* the bug fixed in https://github.com/NNPDF/eko/pull/199
```suggestion
        thresholds_ratios=np.power(list(iter(tcard.matching)), 2.0),
```
see also https://github.com/NNPDF/eko/blob/2a1a26d05b6a36454b8d1e1489e44b983a81a96d/src/eko/runner/legacy.py#L81

AleCandido:
There is one comment from @giacomomagni still open:
https://github.com/NNPDF/pineko/pull/73/files#r1073591854

After that, I believe we can just merge.

src/pineko/evolve.py
AleCandido: @giacomomagni are you sure we need an explicit `nf_to`? I believe it might be useful for matching scales, but for as long as we do not solve https://github.com/NNPDF/eko/issues/149 I believe it should not be possible to retrieve this information.

Would you mind split your comment into a dedicated issue? (possibly tracking EKO ones)

giacomomagni: okay I can move the comment to issues

AleCandido (approved):
The only drawback of this PR is that it is bigger than I would like. But I'm fine with merging and keep going.

andreab1997:
@felixhekhorn @giacomomagni Do you agree that we can merge?

giacomomagni (approved):

felixhekhorn (approved):

PR #75, opened 2023-01-25
Automatic scale variations

andreab1997:
We want a new module called `scale_variations` that should be able to generate automatically renormalization and factorization scale variations starting from a pre-existing grid. It should also be able to generate a new grid filling the scale variations orders. The roadmap of this PR should be approximately the following 

- [x] Be sure that `pineappl_py` is able to extract a subgrid, rescale it and construct a new grid (Note that for factorization scale variations we will also need convolutions)
- [x] Implement renormalization sv (easier)
- [x] Implement a `cli` (maybe taking advantage of the `theory` command) that can upgrade all the necessary grids for a dataset
- [x] Implement a module to include k_factors in a grid 

@felixhekhorn @cschwan @AleCandido This is just a draft, suggestions are welcomed :)


src/pineko/scale_variations.py
```diff
@@ -6,55 +6,149 @@
...
-
+    nec_orders = {}
+    for delt in range(delta):
+        nec_orders[(m + delta, 0, delt + 1, 0)] = [
+            (m + de, 0, 0, 0) for de in range(delta - delt)
+        ]
+    return nec_orders
+
+
+def create_svonly_grid(grid, order, new_order, scalefactor):
```
AleCandido, line 93:
```suggestion
def create_sv_only(grid, order, new_order, scalefactor):
```
that is a grid you know from _the return type_ (in principle, if you wish you can also specify it explicitly in a type hint)

src/pineko/scale_variations.py
```diff
@@ -6,55 +6,149 @@
...
+    for lumi_index in range(len(lumi_grid)):
+        for bin_index in range(grid.raw.bins()):
+            extracted_subgrid = grid.subgrid(order_index, bin_index, lumi_index)
+            extracted_subgrid.scale(scalefactor)
+            # Set this subgrid inside the new grid
+            new_grid.set_subgrid(0, bin_index, lumi_index, extracted_subgrid)
+    return new_grid
+
+
+def create_all_necessary_grids(gridpath, order, kR, nf):
```
AleCandido, line 118:
```suggestion
def create_grids(gridpath, order, kR, nf):
```
if you do not specify which, the only sensible meaning is "all of them". But `create` alone would sound ambiguous, so here I would keep the `_grids`

src/pineko/scale_variations.py
```diff
@@ -6,55 +6,149 @@
...
-
+    delta = to_construct_order[0] - m
+    logpart = to_construct_order[2]
+    return (kR**logpart) * ren_sv_contrib(m, delta, logpart, nec_order[0] - m, nf)
 
-def NLO_ren_var(m, k_R, nf):
-    """NLO ren scale variations depending on the first non vanishing order m.
 
-    k_R is defined as the log(mu_{R}^{2}/ Q^2).
+def compute_map_of_necessary_orders(m, delta):
```
AleCandido, line 70:
```suggestion
def compute_orders_map(m, delta):
```
`necessary` is just a bit redundant :)
(and in any case, you are in the `scale_variations.py` module, so the scope is already telling you something you don't have to repeat)

src/pineko/scale_variations.py
```diff
@@ -6,55 +6,149 @@
...
+        if which_part == 0:
+            if logpart == 1:
+                return -m * beta.beta_qcd((3, 0), nf)
+            else:
+                return 0.5 * m * (m + 1) * (beta.beta_qcd((2, 0), nf) ** 2)
+        else:
+            return -(m + 1) * beta.beta_qcd((2, 0), nf)
+
+
+def compute_scalefactor(m, nec_order, to_construct_order, nf, kR):
```
AleCandido, line 44:
```suggestion
def compute_scale_factor(m, nec_order, to_construct_order, nf, kR):
```

src/pineko/scale_variations.py
```diff
@@ -6,55 +6,149 @@
 from . import check
 
 
-def construct_ren_sv_grid(grid, max_as, nf):
-    """Generate renormalization scale variation terms for the given grid, according to the max_as.
+def ren_sv_contrib(m, delta, logpart, which_part, nf):
```
AleCandido, line 9:
```suggestion
def ren_sv_coeffs(m, delta, logpart, which_part, nf):
```
`contrib` has usually a different meaning in sw.

src/pineko/scale_variations.py
```diff
@@ -6,55 +6,149 @@
...
+            grid_list.append(
+                create_svonly_grid(grid, nec_order, to_construct_order, scalefactor)
+            )
+    return grid_list
 
-def NNLO_ren_var(m, k_R, nf):
-    """NNLO ren scale variations depending on the first non vanishing order m.
 
-    k_R is defined as the log(mu_{R}^{2}/ Q^2).
+def construct_ren_sv_grid(grid, max_as, nf):
```
AleCandido, line 138:
```suggestion
def build_ren_sv(grid, max_as, nf):
```
build is more frequent, but to be more consistent with the other functions
```suggestion
def compute_ren_sv(grid, max_as, nf):
```

src/pineko/scale_variations.py
```diff
@@ -6,55 +6,149 @@
...
+    ]  # +1 because I don't know
+    lumi_grid = [pineappl.lumi.LumiEntry(mylum) for mylum in grid.raw.lumi()]
+    subgrid_params = pineappl.subgrid.SubgridParams()
+    new_order = [pineappl.grid.Order(*new_order)]
+    # create new_grid with same lumi and bin_limits of the original grid but with new_order
+    new_grid = pineappl.grid.Grid.create(
+        lumi_grid, new_order, bin_limits, subgrid_params
+    )
+    # extract the relevant order to rescale from the grid for each lumi and bin
+    grid_orders = [orde.as_tuple() for orde in grid.orders()]
```
AleCandido, line 107: It is fine to shadow in a more internal scope:
```suggestion
    grid_orders = [order.as_tuple() for order in grid.orders()]
```
otherwise you can use `order_`, if you really don't want to

src/pineko/scale_variations.py
```diff
@@ -6,55 +6,149 @@
...
+    nec_orders = compute_map_of_necessary_orders(m_value, deltaorder)
+    grid_list = []
+    for to_construct_order in nec_orders:
+        for nec_order in nec_orders[to_construct_order]:
+            scalefactor = compute_scalefactor(
+                m_value, nec_order, to_construct_order, nf, kR
+            )
+            grid_list.append(
+                create_svonly_grid(grid, nec_order, to_construct_order, scalefactor)
+            )
```
AleCandido, line 134: Instead of appending to the list, you can either:
- move these loops inside the other function, or
- merge the various grids obtained

In the end, you want a single grid output, and appending to a list you have already lost the connection between grids and orders.

src/pineko/scale_variations.py
```diff
@@ -6,55 +6,149 @@
...
+        description of all the needed orders
     """
-    return -k_R * m * beta.beta_qcd((2, 0), nf)
-
+    nec_orders = {}
+    for delt in range(delta):
+        nec_orders[(m + delta, 0, delt + 1, 0)] = [
+            (m + de, 0, 0, 0) for de in range(delta - delt)
+        ]
+    return nec_orders
```
AleCandido, line 90:
```suggestion
    orders = {}
    for delt in range(delta):
        orders[(m + delta, 0, delt + 1, 0)] = [
            (m + de, 0, 0, 0) for de in range(delta - delt)
        ]
    return orders
```

src/pineko/scale_variations.py
```diff
@@ -6,55 +6,149 @@
...
 
-    k_R is defined as the log(mu_{R}^{2}/ Q^2).
+    Parameters
+    ----------
+    m : int
+        first non zero perturbative order
+    nec_order : tuple(int)
+        tuple of the order for which the sv contribution is asked for
+    to_contruct_order : tuple(int)
+        tuple of the sv order to be constructed
```
AleCandido, line 54: Just reading this, I do not understand the difference between the two. Provide at least an example.

AleCandido (changes requested):
Mainly suggestions about simplifying names, and clarifying the meaning. Nothing substantial on the task itself (but *readability matters*!)

src/pineko/scale_variations.py
```diff
@@ -0,0 +1,154 @@
...
+        ]
+    return nec_orders
+
+
+def create_svonly_grid(grid, order, new_order, scalefactor):
+    """Create a grid containing only the renormalization scale variations at a given order for a grid."""
+    # Retrieve parameters to create new grid
+    bin_limits = [
+        float(bin) for bin in range(grid.raw.bins() + 1)
+    ]  # +1 because I don't know
```
AleCandido, line 98: This comment is not reassuring...
(maybe leave at least a `TODO` there)

andreab1997:
> Mainly suggestions about simplifying names, and clarifying the meaning. Nothing substantial on the task itself (but _readability counts_!)

Yes I know :) This is really only a draft. I just want to write a proof of concept example as soon as possible and I will improve docs and readability later. The main point is that I want to know if doing this is possible with what we already have or not. Anyway thanks for the comments :) 

Not strictly related to this comment but I take the opportunity to tag @cschwan which may want to have a look.

AleCandido:
> Yes I know :) This is really only a draft. I just want to write a proof of concept example as soon as possible and I will improve docs and readability later. The main point is that I want to know if doing this is possible with what we already have or not. Anyway thanks for the comments :)

I could see that it was a draft (apart from the draft mode, there were some missing implementations), but you wrote:

> @felixhekhorn @cschwan @AleCandido This is just a draft, suggestions are welcomed :)

so I acted consequently :D

src/pineko/scale_variations.py
cschwan: The `+1` explanation is that `n` bins have `n+1` bin limits, and `range` generates numbers from a half-open interval (`range(n)` generates `n` numbers).

AleCandido: Ok, this makes perfect sense, and it can be a better comment :)

src/pineko/scale_variations.py
andreab1997: Yes indeed, I need yet another function to merge the grids. It is the next step :)

AleCandido:
@andreab1997 whenever you actually need another review, either remove draft mode or tag me again :)

src/pineko/scale_variations.py
```diff
@@ -0,0 +1,194 @@
...
+        number of active flavors
+    """
+    # First let's check if the ren_sv are already there
+    grid = pineappl.grid.Grid.read(grid_path)
+    sv_as, sv_al = check.contains_ren(grid, max_as, max_al=0)
+    if sv_as:
+        rich.print(f"[green]Renormalization scale variations are already in the grid")
+    # Creating all the necessary grids
+    order = (max_as, 0, 0, 0)
+    grid_list = create_grids(grid_path, order, nf)
```
AleCandido, line 190: Pass directly `max_as`:
```suggestion
    grid_list = create_grids(grid_path, max_as, nf)
```

src/pineko/scale_variations.py
```diff
@@ -139,10 +143,18 @@ def write_sv_grids(gridpath, grid_list):
...
     grid_paths = []
-    for num, grid in zip(range(len(grid_list)), grid_list):
-        new_grid_path = gridpath.parent / (base_name + "_" + str(num) + final_part)
+    for order in grid_list:
+        if len(grid_list[order]) > 1:
+            for grid in grid_list[order][1:]:
+                tmp_path = gridpath.parent / ("tmp" + final_part)
+                grid.raw.write_lz4(tmp_path)
+                grid_list[order][0].raw.merge_from_file(tmp_path)
+                os.remove(tmp_path)
```
felixhekhorn, line 165: [No need](https://docs.python.org/3/library/pathlib.html?highlight=pathlib#pathlib.Path.unlink) for `os` (as usual :upside_down_face: )

andreab1997: Yes, thanks for the comment. Anyway this is just a draft, it is going to change a lot in the near future

andreab1997:
The benchmark is not passing now because locally I am using a version that has not been published yet. @cschwan Do you think we can have a new pineappl tag with the small improvements that you made (i.e. the `bins` function) together with the small modifications I have locally (i.e. I removed a couple of `todo`)?

cschwan:
> The benchmark is not passing now because locally I am using a version that has not been published yet. @cschwan Do you think we can have a new pineappl tag with the small improvements that you made (i.e. the `bins` function) together with the small modifications I have locally (i.e. I removed a couple of `todo`)?

I already made some breaking changes in PineAPPL's master branch, so I can't release another `v0.5.x`, would a prelease of `v0.6.0` be OK (that'd be an alpha or beta release)? The `todo!()` would have to fixed or circumvented, I still have to implement this.

andreab1997:
> I already made some breaking changes in PineAPPL's master branch, so I can't release another `v0.5.x`, would a prelease of `v0.6.0` be OK (that'd be an alpha or beta release)? The `todo!()` would have to fixed or circumvented, I still have to implement this.

Yes, that would be fine. I don't believe merging this is top priority because I am the only one using it and for the time being I can use my local environment. Still this will be for sure required in the (next) future.

src/pineko/scale_variations.py
```diff
@@ -192,16 +194,21 @@ def compute_ren_sv_grid(grid_path, max_as, nf):
...
+    target_path: str
+        path where store the new grid (optional)
     """
     # First let's check if the ren_sv are already there
+    grid_path = pathlib.Path(grid_path)
     grid = pineappl.grid.Grid.read(grid_path)
     sv_as, sv_al = check.contains_ren(grid, max_as, max_al=0)
     if sv_as:
         rich.print(f"[green]Renormalization scale variations are already in the grid")
+        return 0
```
felixhekhorn, line 206: Why `0`? `None` is just fine (too much c++? :upside_down_face: )

src/pineko/scale_variations.py
```diff
@@ -0,0 +1,214 @@
...
+
+    Returns
+    -------
+    float
+        renormalization scale variation contribution
+    """
+    if max_as == 0:
+        return 0.0
+    elif max_as == 1:
+        return -m * beta.beta_qcd((2, 0), nf) * (-1.0 / (4.0 * np.pi))
```
felixhekhorn, line 36: Factorize $1/(4\pi)$

src/pineko/scale_variations.py
andreab1997: Definetely :)

AleCandido:
> I already made some breaking changes in PineAPPL's master branch, so I can't release another `v0.5.x`, would a prelease of `v0.6.0` be OK (that'd be an alpha or beta release)? The `todo!()` would have to fixed or circumvented, I still have to implement this.

Since, as @andreab1997 is saying, it is not a top-priority, I guess a pre-release would be perfect!

We will release before a version of Pineko, and then, with a PineAPPL pre-release, we can also merge this branch on `main` and, in case it will be required, make also a pre-release of Pineko.
Usually I don't like to follow too close the release of a package in another package, but in this case the feature is being implemented in the two sides, hopefully there will be soon enough features in PineAPPL that we do not need to touch it any more to implement this.

AleCandido:
@andreab1997 if you are only missing a PineAPPL version to merge, mark this as "Ready for review", and at least we will start improving the code in this PR through reviews :)

src/pineko/scale_variations.py
```diff
@@ -0,0 +1,210 @@
...
+    target_path: str
+        path where store the new grid (optional)
+    """
+    # First let's check if the ren_sv are already there
+    grid_path = pathlib.Path(grid_path)
+    grid = pineappl.grid.Grid.read(grid_path)
+    sv_as, sv_al = check.contains_ren(grid, max_as, max_al=0)
+    if sv_as:
+        rich.print(f"[green]Renormalization scale variations are already in the grid")
+        return None
```
felixhekhorn, line 202:
```suggestion
        return
```

tests/test_evolve.py
```diff
@@ -16,4 +16,3 @@ def test_sv_scheme():
     assert pineko.evolve.sv_scheme(schemeA_tcard) == "exponentiated"
     assert pineko.evolve.sv_scheme(schemeB_tcard) == "expanded"
     assert pineko.evolve.sv_scheme(schemeC_tcard) == None
```
AleCandido, line 18:
```suggestion
    assert pineko.evolve.sv_scheme(schemeC_tcard) is None
```

I know it's not this PR fault, but since you touched the file I saw it :P

pyproject.toml
```diff
@@ -25,7 +25,7 @@ packages = [{ include = "pineko", from = "src" }]
 [tool.poetry.dependencies]
 python = ">=3.8,<3.11"
 eko = "^0.12.0"
-pineappl = "^0.5.9"
+pineappl = "0.6.0a1"
```
AleCandido, line 28: I believe you have to allow prereleases here https://python-poetry.org/docs/dependency-specification/#expanded-dependency-specification-syntax

src/pineko/scale_variations.py
```diff
@@ -200,6 +201,8 @@ def compute_ren_sv_grid(grid_path, max_as, nf, target_path=None):
     if sv_as:
         rich.print(f"[green]Renormalization scale variations are already in the grid")
         return
+    # With respect to the usual convention here max_as is max_as-1
+    max_as = max_as - 1
```
AleCandido, line 205: Why?

pyproject.toml
andreab1997: Ok thank you!

src/pineko/scale_variations.py
andreab1997: As it is written in the comment (`# With respect to the usual convention here max_as is max_as-1`), we are usually adopting the convention such that `max_as = 1` is LO, `max_as = 2` is NLO and so on (in particular the pineappl function `order_mask` is using this). This means that, for instance, the check `check.contains_ren`, that I am doing just before this line,  is built taking this into account. However, all the functions that I have implemented for automatic scale variations are implemented with a different convention and they work decreasing the max_as by 1. If you prefer I can use another parameter instead of changing `max_as`.

AleCandido: > However, all the functions that I have implemented for automatic scale variations are implemented with a different convention and they work decreasing the max_as by 1. If you prefer I can use another parameter instead of changing `max_as`.

But is this something local, is there a reason not to use the same `max_as` also in the functions you wrote?

andreab1997: Of course I could adapt my functions to use the "other" `max_as` but I think that they are more clear as they are, which is the reason why I just changed the max_as value with which I call them. So yes, this is just something local

andreab1997:
Regarding #[d4af2ce](https://github.com/NNPDF/pineko/pull/75/commits/d4af2ce48d2436f020bade904a9dd12112a93378), it is not really related to this PR, however it is related to the fact that we are now using the new pre-release of `pineappl`. So I am sorry for that but I believe that maybe this is the correct PR in which solve that problem. Just to explain, the problem is just that now `grid.axes()` is deprecated and I needed to use the new function `evolve_info`.

src/pineko/scale_variations.py
AleCandido: While I may understand that is clearer as you wrote, it might still be easier to have a single definition of `max_as` around (if possible consistent among all programs, but at least consistent in Pineko).

If we read that code, and there is `max_as` around, but we forget to read first the translation, it will make everything harder to understand.
It is not a bad choice on its own, but consistency is good for maintenance.

AleCandido: I'd just like to understand to which level clearness would be compromised. If it is a "not that much", I'd rather use the original definition of `max_as`

andreab1997: Well basically now all the functions depend on `max_as`. Using instead the other definition all the functions will depend on `max_as - 1`. We will have then a couple of functions in which every time `max_as` is used we need to put `max_as-1`. So at the end instead of having only one place in which we do `max_as - 1` we will have three or four in a couple of functions (that however are not "top-level" functions) .

src/pineko/check.py
```diff
@@ -172,34 +185,50 @@ def contains_ren(grid, max_as, max_al):
...
     for order in order_list:
-        as_orders.append(order[0])
-        al_orders.append(order[1])
-    min_as = min(as_orders)
-    min_al = min(al_orders)
+        if order[1] == 0:
+            as_orders.append(order)
+        else:
+            al_orders.append(order)
+    min_as = min([ord[0] for ord in as_orders]) if len(as_orders) != 0 else 0
```
AleCandido, line 205:
```suggestion
    min_as = min([ord[0] for ord in as_orders], default=0)
```

src/pineko/check.py
```diff
@@ -172,34 +185,50 @@ def contains_ren(grid, max_as, max_al):
...
     order_as_is_present = False
     order_al_is_present = False
     sv_as_present = False
     sv_al_present = False
-    # ren sv starts one order after the first order with as if the first order in as is not as^0
-    add_order = 1 if min_as != 0 else 2
-    for order in order_list:
-        if order[0] == min_as + add_order:
+    # ren sv starts one order after the first order with as (al) if the first order in aa (al) is not as^0 (al^0)
+    add_order_as = 1 + (max_as - 2)
```
AleCandido, line 211: Why not just `-1`?

src/pineko/theory.py
```diff
@@ -378,18 +379,38 @@ def fk(self, name, grid_path, tcard, pdf):
...
-            is_ren_as, is_ren_al = check.contains_ren(grid, max_as, max_al)
+            is_ren_as, is_ren_al, is_cen_as, is_cen_al = check.contains_ren(
+                grid, max_as, max_al
+            )
             if not (is_ren_as and is_ren_al):
                 raise ValueError(
                     "Renormalization scale variations are not available for this grid"
                 )
+            if is_ren_as and not is_cen_as:
+                rich.print(
```
AleCandido, line 390: Prefer logging to printing.

src/pineko/check.py
andreab1997: Of course is the same. However, in this case I wanted to make clear that `+1` means what the comment says above, i.e. that ren sv starts one order later than the first non zero order. Then, because of our convention on `max_as`, I need to do `max_as -2`. If I just write `max_as - 1` it would not be clear (because it won't be consistent with the comment).

src/pineko/theory.py
andreab1997: Yes but in this case there are really 3 different behaviour and I believe that we want to know if `is_ren_as` is `True` because really the ren_sv at that order are there or just because there is no central for that order (which is the case more or less always with NNLO)

AleCandido:
I mentioned a single place, but the same pattern of https://github.com/NNPDF/pineko/pull/75/commits/5b3941b66d40813d1210138fd71340745bde8ca9 was applied in multiple places.

andreab1997:
I am sorry if in 847eb28 I am introducing a whole new package but I believe this is the correct place to do it: I need to include k_factors in a grid in order to be able to use this automatic_scale_varaition package.

src/pineko/kfactor_inclusion.py
```diff
@@ -0,0 +1,291 @@
...
+
+def is_already_in(to_check, list_orders):
+    """Check if the requested order is already in the grid."""
+    is_in = False
+    for order in list_orders:
+        if order[-2] == 0 and order[-1] == 0:
+            if (order[0] == to_check[0]) and (order[1] == to_check[1]):
+                is_in = True
+                return is_in
+    return is_in
```
felixhekhorn, line 217: Why not the short version?
```suggestion
    for order in list_orders:
        if order[-2] == 0 and order[-1] == 0 and (order[0] == to_check[0]) and (order[1] == to_check[1]):
            return True
    return False
```
\+ black of course ;-)

AleCandido (commented):
Extremely basic suggestion about adopting a minimalistic (but clear enough) naming scheme :)

andreab1997:
For the benchmark that is not working at the moment, the reason is that @cschwan removed completely the argument `ignore-bin-limits` but I guess that we do not have a new release yes. Can you confirm @cschwan?

cschwan:
@andreab1997 We can make another prerelease if that's helpful.

src/pineko/kfactor.py
```diff
@@ -0,0 +1,525 @@
...
+    new_grid = pineappl.grid.Grid.create(
+        lumi_grid, new_order, bin_limits, subgrid_params
+    )
+    # extract the relevant order to rescale from the grid for each lumi and bin
+    grid_orders = [order.as_tuple() for order in grid.orders()]
+    order_index = grid_orders.index(order)
+    for lumi_index in range(len(lumi_grid)):
+        for bin_index in range(grid.raw.bins()):
+            extracted_subgrid = grid.subgrid(order_index, bin_index, lumi_index)
+            Q = 10  # Just to see
```
AleCandido, line 119: TODO

src/pineko/kfactor.py
```diff
@@ -0,0 +1,525 @@
+"""Module to include k_factors in grids."""
+import io
+import pathlib
+
+import lhapdf  # pylint: disable=import-error, import-outside-toplevel
```
felixhekhorn, line 5:
```suggestion
import lhapdf  # pylint: disable=import-error
```
this is toplevel :upside_down_face: - then why do you need lhapdf?

andreab1997:
> @andreab1997 We can make another prerelease if that's helpful.

Yes that would be perfect! Thanks

src/pineko/kfactor.py
andreab1997: ops just an error :)

src/pineko/check.py
```diff
@@ -118,36 +123,50 @@ def contains_fact(grid, max_as, max_al):
         is fact scale-variation available for as
     bool
         is fact scale-variation available for al
+    bool
+        is the central as order reached by the grid
+    bool
+        is the central al order reached by the grid
```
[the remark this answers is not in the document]
AleCandido: Better below.

src/pineko/check.py
```diff
@@ -118,36 +123,50 @@ def contains_fact(grid, max_as, max_al):
...
+        if order[1] == min_al + add_order_al:
             order_al_is_present = True
             if order[-1] != 0:
                 sv_al_present = True
     if not order_as_is_present:
         sv_as_present = True
     if not order_al_is_present:
         sv_al_present = True
-    return sv_as_present, sv_al_present
+    return sv_as_present, sv_al_present, order_as_is_present, order_al_is_present
```
AleCandido, line 169: Can you split this function in 4 different functions? Or at least 2, since `as` and `al` are fully decoupled (in particular, they would be the same function, with a switch for one or the other)

They should take as input an `order_list`, and the relevant `max` (call it `max_a`, if you wish).

src/pineko/check.py
```diff
@@ -167,34 +186,50 @@ def contains_ren(grid, max_as, max_al):
...
+    if min_as == 0 and max_as == 2:
+        sv_as_present = True
+    if min_al == 0 and max_al == 2:
+        sv_al_present = True
     if not order_as_is_present:
         sv_as_present = True
     if not order_al_is_present:
         sv_al_present = True
-    return sv_as_present, sv_al_present
+    return sv_as_present, sv_al_present, order_as_is_present, order_al_is_present
```
AleCandido, line 235: Same as `contains_fact`: it is overly complicated for no reason.

Split it, and possibly reuse pieces. Most likely, most of `contains_fact` and `contains_ren` is just the same.

AleCandido (changes requested):
Some comments about `check.py`

cschwan:
> > @andreab1997 We can make another prerelease if that's helpful.
> 
> Yes that would be perfect! Thanks

There you go: <https://pypi.org/project/pineappl/0.6.0a4/>

src/pineko/check.py
andreab1997: The problem is that `as` and `al` are not decoupled. In particular the function that is coupling them is `order_mask = pineappl.grid.Order.create_mask(grid.orders(), max_as, max_al)` and it makes sense because, depending on `max_al`, you need to take into account different orders. Now, it is true that if there is no ren_sv taking `max_al=0`, there won't be for `max_al != 0` as well, but I believe we should not rely on this. This is the reason why they are coupled.

andreab1997: Of course the same applies for `contains_fact`

src/pineko/check.py
```diff
@@ -59,7 +59,12 @@ def check_grid_and_eko_compatible(pineappl_grid, operators, xif):
     ValueError
         If the operators and the grid are not compatible.
     """
-    x_grid, _pids, _mur2_grid, muf2_grid = pineappl_grid.axes()
+    # Note that this is enough but once we consider max_al different
+    # from 0, it will be better to use the actual order_mask
+    mock_order_mask = np.array([True for _ord in pineappl_grid.orders()])
```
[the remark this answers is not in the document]
AleCandido: The function you're calling is actually: https://github.com/NNPDF/pineappl/blob/05c3aadef1c1e5a3e03522fdae6acdad1e5767aa/pineappl_py/src/grid.rs#L557

In Rust there are no optional arguments, but we could have provided in the Python interface:
https://github.com/NNPDF/pineappl/blob/05c3aadef1c1e5a3e03522fdae6acdad1e5767aa/pineappl_py/pineappl/grid.py
That is what that package is made for, but no, looks like no one defined an `.evolve_info()` in Python.

src/pineko/cli/check.py
```diff
@@ -71,13 +71,14 @@ def sub_scvar(grid_path, scale, max_as_order, max_al_order):
     )
     sv_as = None
     central_as = None
-    # TODO: fix this
-    if max_as_effective == max_as_order:
+    # Remember that we have a different definition for the alpha, i.e. we need to sum 1
+    if max_as_effective + 1 == max_as_order:
```
AleCandido, line 75: Exactly, this is the `+ 1` I was thinking about yesterday.

But maybe I'd move on the other side, since it will be the same `- 1` that you have in the other place.

src/pineko/check.py
```diff
@@ -161,40 +165,36 @@ def contains_ren(grid, max_as, max_al):
...
+    max_as_effective = max(ord[0] for ord in as_orders)
+    min_as = min(ord[0] for ord in as_orders)
+    max_as_effective_cen = max(ord[0] for ord in as_orders if ord[index_to_check] == 0)
+    max_as_effective_sv = max(
+        (ord[0] for ord in as_orders if ord[index_to_check] != 0), default=0
+    )
+    if max_as_effective_cen == max_as_effective:
+        if max_as_effective_sv == max_as_effective:
+            checkres = AvailableAtMax.BOTH
+        elif max_as_effective - min_as == 0:
```
AleCandido, line 190:
```suggestion
        elif max_as_effective == min_as:
```
Stupid performance tip: one operation less :)

andreab1997: Fine for me but I believe the other way is more clear

src/pineko/check.py
```diff
@@ -161,40 +165,36 @@ def contains_ren(grid, max_as, max_al):
...
+    max_as_effective = max(ord[0] for ord in as_orders)
+    min_as = min(ord[0] for ord in as_orders)
+    max_as_effective_cen = max(ord[0] for ord in as_orders if ord[index_to_check] == 0)
+    max_as_effective_sv = max(
+        (ord[0] for ord in as_orders if ord[index_to_check] != 0), default=0
+    )
+    if max_as_effective_cen == max_as_effective:
+        if max_as_effective_sv == max_as_effective:
+            checkres = AvailableAtMax.BOTH
+        elif max_as_effective == min_as:
```
AleCandido, line 190: Add a tiny comment also for this case: we exactly know what is happening, but it might be non-trivial for a new reader (a future student, or even our future selves xD)

src/pineko/cli/check.py
```diff
@@ -72,18 +62,41 @@ def sub_scvar(grid_path, scale, max_as_order, max_al_order):
...
-        if condition:
+        if sv_condition and cen_condition:
             to_write += success
+        elif sv_condition and not cen_condition:
+            to_write += warning
         else:
             to_write += error
-        to_write += " " + Scale[scale].value.descr
+        to_write += " " + check.Scale[scale].value.description
         to_write += f" for {coupling.name.lower()}"
```
AleCandido, line 101: Instead of breaking the message in many pieces, and adding them together (that makes the overall message unclear) just use a unique template.
Something like:
```py
message = messages[condition]
descr = check.Scale[scale].value.description
cname = coupling.name.lower()
rich.print(f"{message} {descr} for {cname}")
```

AleCandido (commented):
CLI check review

src/pineko/cli/kfactor.py
```diff
@@ -0,0 +1,41 @@
...
+    compound_folder,
+    target_folder,
+    max_as,
+):
+    """Construct new grid with k_factor included."""
+    grids_folder = pathlib.Path(grids_folder)
+    kfactor_folder = pathlib.Path(kfactor_folder)
+    yamldb_path = pathlib.Path(yamldb_path)
+    target_folder = pathlib.Path(target_folder)
+    comp = pathlib.Path(compound_folder)
```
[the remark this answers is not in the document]
AleCandido: If you have a reason to keep them as they are is fine, but can you write it?

src/pineko/evolve.py
```diff
@@ -99,7 +99,12 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
 
     """
     operators_card = copy.deepcopy(default_card)
-    x_grid, _pids, _mur2_grid, muf2_grid = pineappl_grid.axes()
+    # Note that this is enough but once we consider max_al different
+    # from 0, it will be better to use the actual order_mask
+    mock_order_mask = np.array([True for _ord in pineappl_grid.orders()])
```
AleCandido, line 104: I see that this pattern repeats. If ever, we'll be able save this line https://github.com/NNPDF/pineko/pull/75/files#r1138264088

src/pineko/evolve.py
```diff
@@ -149,7 +154,10 @@ def evolve_grid(
     comparison_pdf : None or str
         if given, a comparison table (with / without evolution) will be printed
     """
-    x_grid, _pids, mur2_grid, _muf2_grid = grid.axes()
+    order_mask = pineappl.grid.Order.create_mask(grid.orders(), max_as, max_al)
+    evolve_info = grid.raw.evolve_info(order_mask)
```
AleCandido, line 158: In general, you should be able to call methods on PineAPPL objects without using `.raw`
```suggestion
    evolve_info = grid.evolve_info(order_mask)
```

If there is any problem with this, please let me know :)

src/pineko/evolve.py
```diff
@@ -99,7 +99,12 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
 
     """
     operators_card = copy.deepcopy(default_card)
-    x_grid, _pids, _mur2_grid, muf2_grid = pineappl_grid.axes()
+    # Note that this is enough but once we consider max_al different
+    # from 0, it will be better to use the actual order_mask
+    mock_order_mask = np.array([True for _ord in pineappl_grid.orders()])
+    evolve_info = pineappl_grid.raw.evolve_info(mock_order_mask)
+    x_grid = evolve_info.x1
+    muf2_grid = evolve_info.fac1
```
AleCandido, line 107: Since the whole pattern is repeated twice in this file, and also somewhere else, you can create your `axes()` function:
```py
def axes(grid):
    mock_order_mask = np.array([True for _ord in grid.orders()])
    evolve_info = pineappl_grid.raw.evolve_info(mock_order_mask)
    return evolve_info.x1, evolve_info.fac1

x_grid, muf2_grid = axes(pineappl_grid)
```

src/pineko/theory.py
```diff
@@ -380,18 +381,29 @@ def fk(self, name, grid_path, tcard, pdf):
         max_al = 0
         # check for sv
         if not np.isclose(xir, 1.0):
-            is_ren_as, is_ren_al = check.contains_ren(grid, max_as, max_al)
-            if not (is_ren_as and is_ren_al):
-                raise ValueError(
-                    "Renormalization scale variations are not available for this grid"
-                )
+            checkres, max_as_effective = check.contains_sv(
```
AleCandido, line 384: I know this function is more or less the central one in Pineko, but it is growing more and more, and start being too long and complex.

We should try to break it down in more atomic pieces.

src/pineko/theory.py
```diff
@@ -380,18 +381,29 @@ def fk(self, name, grid_path, tcard, pdf):
...
+                if max_as == max_as_effective:
+                    if checkres is check.AvailableAtMax.SCVAR:
+                        raise ValueError(
+                            "Central order is not available but sv order is."
+                        )
+                if (
+                    max_as < max_as_effective
+                    and checkres is not check.AvailableAtMax.BOTH
+                ):
+                    raise ValueError("No available central order or sv order.")
```
AleCandido, line 406: If I'm not wrong, this is exactly the same as L384-391, with the only difference of `check.Scale.REN` being traded for `check.Scale.FACT`.

You know how to do it: split a function and call twice ;)

AleCandido (changes requested):
First look at `scale_variations.py`.

You can try to tackle on your own, but we can also discuss together:
- I believe there is an analytic part that can be improved, to be more concise and explicit
- but there is also a verbose usage of the PineAPPLpy API, that might be sometimes improved, and sometimes the improvements might not be possible yet, but the code used here should end up in PineAPPLpy itself

src/pineko/evolve.py
andreab1997: [5b3a12e](https://github.com/NNPDF/pineko/pull/75/commits/5b3a12e7e5f565ddcb5d4511a554868059912c47)

src/pineko/evolve.py
```diff
@@ -11,11 +11,21 @@
 import rich.box
 import rich.panel
 import yaml
-from eko.io.types import ScaleVariationsMethod
+from eko.io.types import EvolutionMethod, ScaleVariationsMethod
 
 from . import check, comparator, ekompatibility, version
 
 
+def axes(grid, max_as=None, max_al=None):
```
AleCandido, line 19: I agree that it makes sense to have the two cases together, but since it doesn't make much sense to have the max for one and not the other (and indeed in that case you're ignoring the passed one), I'd collect them in a single argument:
```suggestion
def axes(grid, maxs: Optional[Tuple[int]]=None):
```

src/pineko/scale_variations.py
```diff
@@ -0,0 +1,237 @@
...
+    Returns
+    -------
+    dict(tuple(int))
+        description of all the needed orders
+    """
+    orders = {}
+    for delt in range(max_as):
+        orders[(m + max_as, min_al, delt + 1, 0)] = [
+            (m + de, min_al, 0, 0) for de in range(max_as - delt)
+        ]
```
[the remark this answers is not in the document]
AleCandido: Re the second point: not a problem from UI point of view, calling once or twice might even be fine.

Simply, I do not see much the point of computing one and not the other, I believe we should compute all the missing ones. Is there a use case for computing only NNLO, if NLO are also missing?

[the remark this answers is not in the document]
andreab1997: > Simply, I do not see much the point of computing one and not the other, I believe we should compute all the missing ones. Is there a use case for computing only NNLO, if NLO are also missing?

No of course but if you want to include all the sv terms up the one the user is asking, it means that we need to be careful and testing that all the centrals are included (I mean the ones used to compute the sv terms). Moreover, if I don't see when, I believe it could be useful to have the possibility of including only one contribution even in relation with the k_factor business that, in principle can also be used to include sv k_factors.

src/pineko/evolve.py
andreab1997: [2cbedfb](https://github.com/NNPDF/pineko/pull/75/commits/2cbedfb06f19413cc18d5b86c41170287267a1da)

RoyStegeman:
As a general question, do you document all the cli scripts or rely on docstrings?

andreab1997:
> As a general question, do you document all the cli scripts or rely on docstrings?

In principle I should document everything but for the moment we are relying on the docstrings :)

AleCandido:
> As a general question, do you document all the cli scripts or rely on docstrings?

There is a single CLI, organized as a package, not a set of scripts.
The CLI framework, `click`, transforms docstrings in help messages. You can invoke them with `-h/--help`.

Sphinx documentation might be useful for tutorials (but they are partly already on https://nnpdf.github.io/pineline), and a brief overview. But I'd not duplicate docstrings information, given that is already well exposed to the user :)

src/pineko/kfactor.py
```diff
@@ -0,0 +1,528 @@
...
+                break
+            descstring.write(line)
+        description = descstring.getvalue()
+        try:
+            data = np.loadtxt(f)
+        except Exception as e:
+            raise TypeError(e) from e
+        data = data.reshape(-1, 2)
+        central_value = data[:, 0]
+        pdf_set = description.split(sep="PDFset:")[-1].split(sep="\n")[0].strip()
```
RoyStegeman, line 33: Not all pdfset names are in lhapdf format, see e.g. seaquest where it's just "NNPDF3.1 NNLO". Also in many cases the pdfs will not be in lhpadf so they will first have to be obtained through vp.

src/pineko/cli/kfactor.py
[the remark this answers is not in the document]
RoyStegeman: I think they're just checks of the input type, they only accept `Path`-like objects but don't return them.

src/pineko/cli/kfactor.py
```diff
@@ -0,0 +1,41 @@
...
+from .. import kfactor
+from ._base import command
+
+
+@command.command("kfactor")
+@click.argument("grids_folder", type=click.Path(exists=True))
+@click.argument("kfactor_folder", type=click.Path(exists=True))
+@click.argument("yamldb_path", type=click.Path(exists=True))
+@click.argument("compound_folder", type=click.Path(exists=True))
+@click.argument("target_folder", type=click.Path(exists=False))
```
RoyStegeman, line 18:
```suggestion
@click.argument("target_folder", type=click.Path(exists=True))
```

Or alternatviely it would make sense for the script to create the folder, e.g. by calling `mkdir` on the `pahtlib.Path` below.

RoyStegeman (commented):
I guess my review is too early? I'm pretty sure that in many cases it doesn't correctly work yet. Anyway, I've left a few comments but this is not a full review.

src/pineko/cli/kfactor.py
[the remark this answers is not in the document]
andreab1997: Yes indeed, in fact I restored these lines

andreab1997:
It seems that pylint was complaining for an actual error only for `python3.8`. Do you know why? @AleCandido @felixhekhorn

andreab1997:
It seems that `pylint` was complaining for an actual error (https://github.com/NNPDF/pineko/actions/runs/4499485302/jobs/8026830359) but only for `python3.8`. Do you know why?

src/pineko/scale_variations.py
```diff
@@ -0,0 +1,237 @@
...
+import pineappl
+import rich
+from eko import beta
+
+from . import check
+
+
+def sort_orders(order):
+    """Define the sorting for an order. In particular the order is sorted according to the as order."""
+    return order[0]
```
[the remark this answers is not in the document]
andreab1997: [cc17832](https://github.com/NNPDF/pineko/pull/75/commits/cc17832e097f73338aa83026895abb6df1ad4150)

src/pineko/scale_variations.py
```diff
@@ -0,0 +1,237 @@
...
+            else:
+                return (
+                    0.5
+                    * m
+                    * (m + 1)
+                    * (beta.beta_qcd((2, 0), nf) ** 2)
+                    * (pifactor**2)
+                )
+        else:
+            return -(m + 1) * beta.beta_qcd((2, 0), nf) * pifactor
```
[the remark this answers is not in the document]
andreab1997: [03ab298](https://github.com/NNPDF/pineko/pull/75/commits/03ab2989ad94147d7e5bd726ba221827d829ac3b)

src/pineko/scale_variations.py
```diff
@@ -0,0 +1,237 @@
...
+    nf : int
+        number of active flavors
+
+    Returns
+    -------
+    float
+        full contribution of ren sv
+    """
+    max_as = to_construct_order[0] - m
+    logpart = to_construct_order[2]
```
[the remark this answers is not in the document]
andreab1997: [7663846](https://github.com/NNPDF/pineko/pull/75/commits/766384657096503de41232e4432862df0fb80cf6)

felixhekhorn:
> It seems that `pylint` was complaining for an actual error (https://github.com/NNPDF/pineko/actions/runs/4499485302/jobs/8026830359) but only for `python3.8`. Do you know why? @AleCandido @felixhekhorn

I believe in py3.8 the type "tuple" is still called "Tuple" i.e. [`typing.Tuple`](https://docs.python.org/3/library/typing.html?highlight=optional#typing.Tuple) - only in a later version the "typing type" and the "actual type" can be used synonymous (if I remember correctly what @AleCandido told me once)

src/pineko/scale_variations.py
```diff
@@ -0,0 +1,237 @@
...
+    bin_limits = [
+        float(bin) for bin in range(grid.raw.bins() + 1)
+    ]  # The +1 explanation is that n bins have n+1 bin limits, and range generates numbers from a half-open interval (range(n) generates n numbers).
+    lumi_grid = [pineappl.lumi.LumiEntry(mylum) for mylum in grid.raw.lumi()]
+    subgrid_params = pineappl.subgrid.SubgridParams()
+    new_order = [pineappl.grid.Order(*new_order)]
+    # create new_grid with same lumi and bin_limits of the original grid but with new_order
+    new_grid = pineappl.grid.Grid.create(
+        lumi_grid, new_order, bin_limits, subgrid_params
+    )
```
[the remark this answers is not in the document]
andreab1997: [d1ac639](https://github.com/NNPDF/pineko/pull/75/commits/d1ac6394884792106b4be166f2e671cf70ba65df)

src/pineko/kfactor.py
```diff
@@ -0,0 +1,528 @@
...
+    # For RATIO dataset the thing is a bit more messy
+    if operation == "RATIO":
+        is_concatenated = True
+        for op_list in yamldict["operands"]:
+            list_op_list = []
+            for op in op_list:
+                list_op_list.append(op)
+            list_grids.append(list_op_list)
+    else:
+        for op in yamldict["operands"][0]:
```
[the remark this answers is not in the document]
andreab1997: The distinction between RATIO and non-RATIO is exactly done for this reason. So, I checked the k-factors and, for what I have seen, `operands` is always a list of lists but, when the dataset is not RATIO, it is just a trivial list of lists, i.e. the external list only contains one single list with the operands (which is the reason why in the non-RATIO case I do `for op in yamldict["operands"][0]`). In the RATIO case instead I need to be careful and take both the lists that are contained in `operands`. However you are right that I would need to do the same for ASY. Thanks!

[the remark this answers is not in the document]
andreab1997: [f79483a](https://github.com/NNPDF/pineko/pull/75/commits/f79483ac47c3061b4d3c2db7cbb405f511117639) this should have solved the problem

src/pineko/kfactor.py
RoyStegeman: I understand what you tried, but I think in this case you'd end up doing `lhapdf.mkAlphaS("NNPDF3.1 NNLO")`, since `pdf_set` here would not be an empty string as you check below. 

My suggestion would be to just remove the if statement check below to really make this function called `read_kfactor` about only reading the cfactor file. Then below you can do a try except when doing `lhapdf.mkAlphaS`.

src/pineko/kfactor.py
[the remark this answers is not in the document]
RoyStegeman: I see. Okay I'll have a closer look tomorrow but I think we can make this more general. If it's not a list of list in all cases we may need to update the yamldb file or perhaps the yaml reader settings since I suspect we can make this more general.

src/pineko/kfactor.py
andreab1997: I agree and I can even do it but, as I said, I believe that just editing the K-factor file is better and it is what I did. I think that in general it does not make much sense that the PDF written there is not an actual PDF so I don't think catching the error  is the best strategy in this case

RoyStegeman: I see, so you want it to raise an error if there is a `PDFset` that is not in lhapdf, but use NNPDF4.0 if there is no `PDFset` at all? That's might be a reasonable enough thing to do, but it would be good if that were reflected more explicitly in the code as well.

andreab1997: Yes you are right and yes, that is the idea

tests/test_kfactor.py
```diff
@@ -0,0 +1,42 @@
+import numpy as np
+
+from pineko import kfactor
+
+
+class fakealpha:
```
AleCandido, line 6: I know it is just for testing, but try to be consistent with naming conventions:
```suggestion
class FakeAlpha:
```

andreab1997: [ee68474](https://github.com/NNPDF/pineko/pull/75/commits/ee684740e87309acb1659b98115ea7d5ea057ab9)

RoyStegeman:
@andreab1997 what is the status of this PR? Is everything working and ready to be merged or are there still problems to be fixed (e.g. `pineappl.optimize()` related)?

src/pineko/check.py
```diff
@@ -106,58 +140,25 @@ def is_fonll_b(fns, lumi):
...
-             max al order
-    Returns
-    -------
-    bool
-        is fact scale-variation available for as
-    bool
-        is fact scale-variation available for al
-    """
+def get_order_list(grid, max_as, max_al):
+    """Return the order list after filtering according to max_as and max_al."""
```
AleCandido, line 144: Just because of my aversion for redundant names (let's save characters, I'll take less to review!) I'd really drop the `get_` in the name, and even the `Return` in the docstring.

```suggestion
def orders(grid, max_as, max_al):
    """Select the relevant orders. 
    
    The orders in the grid are filtered according to `max_as` and `max_al`.
    
    """
```

Moreover, `_list` is just explicating the return type. If that is relevant, I would suggest to use type hints:
```py
def orders(grid, max_as, max_al) -> list:
```

src/pineko/check.py
```diff
@@ -106,58 +140,25 @@ def is_fonll_b(fns, lumi):
...
     order_array = np.array([order.as_tuple() for order in grid.orders()])
-    order_mask = pineappl.grid.Order.create_mask(grid.orders(), max_as, max_al)
+    order_mask = pineappl.grid.Order.create_mask(grid.orders(), max_as, max_al, True)
     order_list = order_array[order_mask]
+    return order_list
+
+
+def pure_qcd_orders(order_list):
+    """Return the pure alpha_s orders appearing in the grid."""
```
AleCandido, line 152:
```suggestion
def pure_qcd(orders):
    """Select the QCD LO and pure QCD corrections to it."""
```

src/pineko/check.py
```diff
@@ -106,58 +140,25 @@ def is_fonll_b(fns, lumi):
...
-        sv_as_present = True
-    if not order_al_is_present:
-        sv_al_present = True
-    return sv_as_present, sv_al_present
-
-
-def contains_ren(grid, max_as, max_al):
+        if order[1] == min_al:
+            as_orders.append(order)
+    return as_orders
```
AleCandido, line 158: For this a list comprehension might be appropriate:
```py
    min_al = min(ord[1] for ord in order_list)
    return [ord for ord in orders if ord[1] == min_al]
```

I know it is just syntactic sugar, but it seems extremely clear, so I'd exploit it :)

src/pineko/check.py
```diff
@@ -168,40 +169,37 @@ def contains_ren(grid, max_as, max_al):
...
     Returns
     -------
-    bool
-        is ren scale-variation available for as
-    bool
-        is ren scale-variation available for al
+    : AvailableAtMax
+        result of the check
+    : int
+        effective max_as in the grid
```
AleCandido, line 180: Since the information in the docstring is not telling much more than what the names are already telling, I'd propose you to move the types in the function type hints, and drop the `Parameters` and `Returns` sections.

src/pineko/check.py
```diff
@@ -168,40 +169,37 @@ def contains_ren(grid, max_as, max_al):
...
-            if order[-2] != 0:
-                sv_al_present = True
-    if not order_as_is_present:
-        sv_as_present = True
-    if not order_al_is_present:
-        sv_al_present = True
-    return sv_as_present, sv_al_present
+    index_to_check = sv_type.value.index
+    order_list = get_order_list(grid, max_as, max_al)
+    as_orders = pure_qcd_orders(order_list)
```
AleCandido, line 184: This function only works for pure QCD, and nothing else. Make it explicit in the docstring.

Because of this, the only relevant `orders` are `pure_qcd` ones, so I would fold these functions:
```suggestion
    ords = pure_qcd(orders(grid, max_as, max_al))
```

And since only pure QCD are considered, `ords` is not ambiguous (`ords` instead of `orders` to avoid shadowing the function, it is clear enough in this context).

src/pineko/check.py
```diff
@@ -168,40 +169,37 @@ def contains_ren(grid, max_as, max_al):
...
+    max_as_effective = max(ord[0] for ord in as_orders)
+    min_as = min(ord[0] for ord in as_orders)
+    max_as_effective_cen = max(ord[0] for ord in as_orders if ord[index_to_check] == 0)
+    max_as_effective_sv = max(
+        (ord[0] for ord in as_orders if ord[index_to_check] != 0), default=0
+    )
```
AleCandido, line 190: I believe that `_effective` is there because of the pure QCD restriction. But as said before, this whole function is only for pure QCD, so the context already clarifies it:
```suggestion
    max_as = max(ord[0] for ord in ords)
    min_as = min(ord[0] for ord in ords)
    max_as_cen = max(ord[0] for ord in ords if ord[svindex] == 0)
    max_as_sv = max(
        (ord[0] for ord in ords if ord[svindex] != 0), default=0
    )
```

src/pineko/check.py
```diff
@@ -168,40 +169,37 @@ def contains_ren(grid, max_as, max_al):
...
-        if order[1] == min_al + 1:
-            order_al_is_present = True
-            if order[-2] != 0:
-                sv_al_present = True
-    if not order_as_is_present:
-        sv_as_present = True
-    if not order_al_is_present:
-        sv_al_present = True
-    return sv_as_present, sv_al_present
+    index_to_check = sv_type.value.index
```
AleCandido, line 182:
```suggestion
    svindex = sv_type.value.index
```
Slightly clearer (`sv` is more specific than `to_check`). Even if `sv` is already in the context, `index` is definitely too generic on its own.

src/pineko/cli/kfactor.py
```diff
@@ -0,0 +1,38 @@
...
+    yamldb_path,
+    target_folder,
+    max_as,
+    order_exists,
+):
+    """Construct new grid with k_factor included."""
+    grids_folder = pathlib.Path(grids_folder)
+    kfactor_folder = pathlib.Path(kfactor_folder)
+    yamldb_path = pathlib.Path(yamldb_path)
+    target_folder = pathlib.Path(target_folder)
```
AleCandido, line 30: You can avoid all of these, since `click` is already making them of type `Path`, since you specified it.
```suggestion
```

src/pineko/evolve.py
```diff
@@ -2,6 +2,7 @@
 import copy
 import os
 import pathlib
+from typing import Optional, Tuple
```
AleCandido, line 5: I believe you're not using this imports anywhere.

src/pineko/evolve.py
```diff
@@ -11,7 +12,7 @@
 import rich.box
 import rich.panel
 import yaml
-from eko.io.types import ScaleVariationsMethod
+from eko.io.types import EvolutionMethod, ScaleVariationsMethod
```
AleCandido, line 15: It seems to me that you're not using even `EvolutionMethod` anywhere...

src/pineko/evolve.py
```diff
@@ -178,6 +181,7 @@ def evolve_grid(
     # Here we are checking if the EKO contains the rotation matrix (flavor to evol)
     elif not np.allclose(operators.rotations.inputpids, br.rotate_flavor_to_evolution):
         raise ValueError("The EKO is neither in flavor nor in evolution basis.")
+    # do it
```
AleCandido, line 184: ?

src/pineko/theory.py
```diff
@@ -15,11 +15,21 @@
...
 import yaml
 
-from . import check, configs, evolve, parser, theory_card
+from . import check, configs, evolve, parser, scale_variations, theory_card
 
 logger = logging.getLogger(__name__)
 
 
+def check_scvar_evolve(grid, max_as, max_al, tocheck):
+    """Check if the scale variations and central orders are to enough to construct the fktable."""
```
AleCandido, line 24:
```suggestion
    """Check scale variations presence and consistency.
    
    In particular, check if the scale variations and central orders are those 	
    required to construct a scale-varied fktable.
    
    """
```

Try to keep the first line as brief as possible (but meaningful)

src/pineko/theory.py
```diff
@@ -15,11 +15,21 @@
 import rich
 import yaml
 
-from . import check, configs, evolve, parser, theory_card
+from . import check, configs, evolve, parser, scale_variations, theory_card
 
 logger = logging.getLogger(__name__)
 
 
+def check_scvar_evolve(grid, max_as, max_al, tocheck):
```
AleCandido, line 23:
```suggestion
def check_scvar_evolve(grid, max_as, max_al, kind: Scale):
```
A slightly clearer variable name, especially if accompanied by the type hint.

(`type` would be good as well, but `type` is a built-in in Python, and it has a specific meaning in almost any programming language, so I try to avoid it).

src/pineko/theory.py
```diff
@@ -15,11 +15,21 @@
...
+    """Check if the scale variations and central orders are to enough to construct the fktable."""
+    checkres, max_as_effective = check.contains_sv(grid, max_as, max_al, tocheck)
```
AleCandido, line 25:
```suggestion
    available, max_as_effective = check.contains_sv(grid, max_as, max_al, kind)
```

`checkres` is a bit generic. In case you want it, just use `res`, since `check` is in the name of the function called, and not ambiguous in this context.

AleCandido (commented):
Still missing `kfactor` and `scale_variations` (the big ones).

Pay attention to the suggestions: they are not always self-contained (so do not commit them from GitHub, unless you're sure by yourself that they are actually self-contained).

src/pineko/scale_variations.py
```diff
@@ -0,0 +1,266 @@
...
+        number of active flavors
+
+    Returns
+    -------
+    float
+        renormalization scale variation contribution
+    """
+    bcoeff = beta.beta_qcd((max_as - logpart - which_part + 2, 0), nf)
+    as_norm_fact = AS_NORM ** (max_as - which_part)
+    m_factor = m
```
AleCandido, line 45: You can also use `m`: the modified variable would be only in this scope, since `float`s are not mutable (to first approximation, you can think that they don't own "dynamic memory").

The only effect of mutating `m` is that you won't have access any longer to the original value. But you don't need it any longer, since the rest of the function only uses `m_factor` and not `m`.
```suggestion
```

src/pineko/scale_variations.py
```diff
@@ -0,0 +1,266 @@
...
+OrderTuple = Tuple[int, int, int, int]
+"""Tuple representing a PineAPPL order."""
+
+
+def sort_qcd_orders(order: OrderTuple) -> int:
+    """Define the sorting for an order. In particular the order is sorted according to the as order."""
+    return order[0]
+
+
+def ren_sv_coeffs(m, max_as, logpart, which_part, nf):
```
AleCandido, line 22: I would try to improve the names.

- `nf` is fine as it is
- `m`, `max_as`, `logpart`, `which_part` are all orders (or order-tuple elements), but from their names they seem very different from each other: I'd try to find more explicit and harmonized names

I will try to propose some names myself.

src/pineko/scale_variations.py
```diff
@@ -0,0 +1,266 @@
...
+    nf : int
+        number of active flavors
+
+    Returns
+    -------
+    float
+        renormalization scale variation contribution
+    """
+    bcoeff = beta.beta_qcd((max_as - logpart - which_part + 2, 0), nf)
+    as_norm_fact = AS_NORM ** (max_as - which_part)
```
AleCandido, line 44:
```suggestion
    as_normalization = AS_NORM ** (max_as - which_part)
```

`_fact` is fully generic, and if you only need a name different from `as_norm`, then much better to have entire words.

Notice that in principle even `as_norm` would be available, since `as_norm != AS_NORM`. But this would be confusing to refer to.

AleCandido: Otherwise, just use a more consistent naming with the previous term, and call it:
```suggestion
    ascoeff = AS_NORM ** (max_as - which_part)
```

src/pineko/scale_variations.py
```diff
@@ -0,0 +1,266 @@
...
+    m_factor = m
+    if max_as == 0:
+        return 0.0
+    elif max_as == 2:
+        if which_part > 0:
+            m_factor += 1
+        else:
+            if logpart > 1:
+                m_factor = 0.5 * m_factor * (m_factor + 1)
```
AleCandido, line 53:
```suggestion
        elif logpart > 1:
            m_factor = 0.5 * m_factor * (m_factor + 1)
```

src/pineko/scale_variations.py
```diff
@@ -0,0 +1,266 @@
...
+    return m_factor * as_norm_fact * bcoeff
+
+
+def compute_orders_map(m, max_as, min_al):
```
AleCandido, line 57:
```suggestion
def compute_orders_map(m, max_as, al):
```

No need to call it `min_`, since it is only the order of interest for this function. The fact that is minimum is external to this function, but only relevant for the usage (i.e. for the caller).

AleCandido: I'd like a better name than `al`, because it seems the value of $\alpha$, rather than its power.

But this would be part of the orders renaming, also needed for the function above. Let's think about them.

AleCandido, line 57:
```suggestion
def compute_orders_map(m, max_as, min_al) -> Dict[OrderTuple, List[OrderTuple]]:
```

src/pineko/scale_variations.py
```diff
@@ -0,0 +1,266 @@
...
+    -------
+    dict(tuple(int))
+        description of all the needed orders
+    """
+    orders = {}
+    for delt in range(max_as):
+        orders[(m + max_as, min_al, delt + 1, 0)] = [
+            (m + de, min_al, 0, 0) for de in range(max_as - delt)
+        ]
+    return orders
```
AleCandido, line 78: In principle even this can be a clean dictionary comprehension:
```suggestion
    return {
        (m + max_as, min_al, delt + 1, 0): [
            (m + de, min_al, 0, 0) for de in range(max_as - delt)
        ]
        for delt in range(max_as)
    }
```

(Most likely Black would reformat it differently...)

AleCandido (commented):
Okay, here are a few suggestions on the already clean enough part of `scale_variations`.
I will address the more critical one in the next review.

src/pineko/check.py
andreab1997: [2857085](https://github.com/NNPDF/pineko/pull/75/commits/2857085498191b8b5a650a2333b81f184ce89271)

src/pineko/check.py
andreab1997: [4a1066b](https://github.com/NNPDF/pineko/pull/75/commits/4a1066bcdbc04b4f3deb420914620b230dc9c38b)

src/pineko/check.py
andreab1997: [6fef36d](https://github.com/NNPDF/pineko/pull/75/commits/6fef36d9bb18081b706428f993776fa827643f2a)

src/pineko/check.py
andreab1997: [d9b063c](https://github.com/NNPDF/pineko/pull/75/commits/d9b063c4be679386dcbe4fae246bf35756772716)

src/pineko/cli/kfactor.py
andreab1997: Actually I am not sure, I tried without and it falied exactly because one of these paths was actually a string. In fact if you look to the commit history you can see that at a certain point I removed this and then added it again.

src/pineko/check.py
andreab1997: [6ee6ddd](https://github.com/NNPDF/pineko/pull/75/commits/6ee6ddd3bfc28bd0ec3882c9970e4923a2502205)

src/pineko/check.py
andreab1997: [6ee6ddd](https://github.com/NNPDF/pineko/pull/75/commits/6ee6ddd3bfc28bd0ec3882c9970e4923a2502205)

src/pineko/evolve.py
andreab1997: [6ee6ddd](https://github.com/NNPDF/pineko/pull/75/commits/6ee6ddd3bfc28bd0ec3882c9970e4923a2502205)

src/pineko/evolve.py
andreab1997: Well after the checks you start "doing it" :). Joke apart I can remove it

andreab1997: [9b34b1d](https://github.com/NNPDF/pineko/pull/75/commits/9b34b1d9e94671b5000428c78f7fb013a20d2913)

src/pineko/theory.py
andreab1997: [bc2c141](https://github.com/NNPDF/pineko/pull/75/commits/bc2c14168987895dd9926d2f2743034829e9d5f6)

src/pineko/theory.py
andreab1997: [459450f](https://github.com/NNPDF/pineko/pull/75/commits/459450f6c5df40023f20d72d1db58623cd164666)

src/pineko/theory.py
andreab1997: [e76bd60](https://github.com/NNPDF/pineko/pull/75/commits/e76bd60d64bb9be1dedd3b4230fd9453401cf10b)

src/pineko/scale_variations.py
```diff
@@ -0,0 +1,266 @@
...
+from eko import beta
+
+from . import check
+
+AS_NORM = 1.0 / (4.0 * np.pi)
+OrderTuple = Tuple[int, int, int, int]
+"""Tuple representing a PineAPPL order."""
+
+
+def sort_qcd_orders(order: OrderTuple) -> int:
```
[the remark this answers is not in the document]
andreab1997: [9705ce8](https://github.com/NNPDF/pineko/pull/75/commits/9705ce82b15580f9dadedf7f0df9b7585ada61ac)

src/pineko/scale_variations.py
```diff
@@ -0,0 +1,266 @@
...
+    """Define the sorting for an order. In particular the order is sorted according to the as order."""
+    return order[0]
+
+
+def ren_sv_coeffs(m, max_as, logpart, which_part, nf):
+    """Return the ren_sv contribution relative to the requested log power and perturbative order contribution (which_part).
```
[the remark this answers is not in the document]
andreab1997: [439af64](https://github.com/NNPDF/pineko/pull/75/commits/439af64ac1bed876d8c5f3327a01ff2ca769d411)

src/pineko/scale_variations.py
andreab1997: [08aade5](https://github.com/NNPDF/pineko/pull/75/commits/08aade5443e9baeb6515e960d8234d0f2a418748)

src/pineko/scale_variations.py
andreab1997: [2aecc15](https://github.com/NNPDF/pineko/pull/75/commits/2aecc1562cc6a710935adb877ca7cc9156dd2fd9)

src/pineko/check.py
```diff
@@ -140,66 +141,47 @@ def is_fonll_b(fns, lumi):
...
     order_mask = pineappl.grid.Order.create_mask(grid.orders(), max_as, max_al, True)
     order_list = order_array[order_mask]
     return order_list
 
 
-def pure_qcd_orders(order_list):
-    """Return the pure alpha_s orders appearing in the grid."""
+def pure_qcd(orders):
+    """Select the QCD LO and pure QCD corrections to it."""
     as_orders = []
```
AleCandido, line 158: Now dead :)

src/pineko/scale_variations.py
andreab1997: [c822719](https://github.com/NNPDF/pineko/pull/75/commits/c822719d67cbef26cc67c418c67a8c44b2876244)

andreab1997: I am not marking this as resolved so we can keep in our mind the naming business

src/pineko/check.py
andreab1997: ?

src/pineko/scale_variations.py
andreab1997: [e36e120](https://github.com/NNPDF/pineko/pull/75/commits/e36e1202e97fecf54effe3dc6c3ccf26246cae78)

src/pineko/check.py
andreab1997: Ah you mean that `as_orders` is not used anymore, right?

src/pineko/scale_variations.py
andreab1997: [4ba538a](https://github.com/NNPDF/pineko/pull/75/commits/4ba538a43a1ed708e52fa1d7294f129107c81f10)

src/pineko/scale_variations.py
andreab1997: [d786b4c](https://github.com/NNPDF/pineko/pull/75/commits/d786b4c6cd4f6cfc384f1284532423a64be60f34)

src/pineko/scale_variations.py
```diff
@@ -0,0 +1,276 @@
...
+        return 0.0
+    if max_as == 2:
+        if which_part > 0:
+            m += 1
+        elif logpart > 1:
+            m = 0.5 * m * (m + 1)
+    return m * as_normalization * bcoeff
+
+
+def compute_orders_map(m, max_as, al) -> Dict[OrderTuple, List[OrderTuple]]:
```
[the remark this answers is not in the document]
andreab1997: [39cc3ca](https://github.com/NNPDF/pineko/pull/75/commits/39cc3ca9a161d9e0a71b219be0f385d0db015090)

src/pineko/scale_variations.py
```diff
@@ -0,0 +1,276 @@
...
+        max as order
+    nf : int
+        number of active flavors
+    target_path: pathlib.Path
+        path where store the new grid (optional)
+    """
+    # First let's check if the ren_sv are already there
+    grid_path = pathlib.Path(grid_path)
+    grid = pineappl.grid.Grid.read(grid_path)
+    checkres, max_as_effective = check.contains_sv(grid, max_as, 0, check.Scale.REN)
```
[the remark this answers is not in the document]
andreab1997: [9948075](https://github.com/NNPDF/pineko/pull/75/commits/99480750fe6e53977066144672f39a0c38a85523)

src/pineko/scale_variations.py
```diff
@@ -0,0 +1,276 @@
...
+        pineappl grid path
+    max_as : int
+        max as order
+    nf : int
+        number of active flavors
+    target_path: pathlib.Path
+        path where store the new grid (optional)
+    """
+    # First let's check if the ren_sv are already there
+    grid_path = pathlib.Path(grid_path)
```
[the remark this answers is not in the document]
andreab1997: [67d24b1](https://github.com/NNPDF/pineko/pull/75/commits/67d24b1b049690593a398043e855929f4fe8af17)

src/pineko/scale_variations.py
```diff
@@ -0,0 +1,276 @@
...
+    checkres, max_as_effective = check.contains_sv(grid, max_as, 0, check.Scale.REN)
+    # Usual different convention with max_as
+    if max_as_effective == max_as and (checkres is not check.AvailableAtMax.CENTRAL):
+        if not order_exists:
+            rich.print(
+                f"[green]Renormalization scale variations are already in the grid"
+            )
+            return
+    else:
+        if order_exists:
```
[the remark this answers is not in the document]
andreab1997: [75f3f47](https://github.com/NNPDF/pineko/pull/75/commits/75f3f472c8bea486754d1c07060d8ba4734e09f9)

src/pineko/scale_variations.py
```diff
@@ -0,0 +1,276 @@
...
+            raise ValueError(
+                "Order_exists is True but the order does not appear to be in the grid"
+            )
+    if max_as_effective < max_as and checkres is check.AvailableAtMax.SCVAR:
+        raise ValueError(
+            "Central order is not high enough to compute requested sv orders"
+        )
+    # With respect to the usual convention here max_as is max_as-1
+    max_as = max_as - 1
```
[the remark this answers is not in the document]
andreab1997: [e6180f4](https://github.com/NNPDF/pineko/pull/75/commits/e6180f4b36079ec9e307986cf132171966c03f9f)

src/pineko/scale_variations.py
```diff
@@ -0,0 +1,276 @@
...
+    m : int
+        first non zero perturbative order of the grid
+    max_as : int
+        max alpha_s order
+    min_al : int
+        al order of leading order
+    Returns
+    -------
+    dict(tuple(int))
+        description of all the needed orders
```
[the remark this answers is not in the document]
andreab1997: [6a24833](https://github.com/NNPDF/pineko/pull/75/commits/6a248336c71de920ce9fd864535a05058c2ab147)

src/pineko/scale_variations.py
```diff
@@ -0,0 +1,276 @@
...
+    """Merge the single grids in the original."""
+    grid = pineappl.grid.Grid.read(gridpath)
+    if target_path is None:
+        target_path = gridpath.parent / gridpath.name
+    else:
+        target_path = target_path / gridpath.name
+    if order_exists:
+        grid = construct_and_dump_order_exists_grid(grid, list(nec_orders.keys())[0])
+    for grid_path in grid_list_path:
+        grid.raw.merge_from_file(grid_path)
```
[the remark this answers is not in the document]
andreab1997: [82c1ed9](https://github.com/NNPDF/pineko/pull/75/commits/82c1ed9bfabe5586e1b20d80fe6a3e3b77d4c305)

src/pineko/scale_variations.py
```diff
@@ -0,0 +1,276 @@
...
+            return
+    else:
+        if order_exists:
+            raise ValueError(
+                "Order_exists is True but the order does not appear to be in the grid"
+            )
+    if max_as_effective < max_as and checkres is check.AvailableAtMax.SCVAR:
+        raise ValueError(
+            "Central order is not high enough to compute requested sv orders"
+        )
```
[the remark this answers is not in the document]
andreab1997: [270682e](https://github.com/NNPDF/pineko/pull/75/commits/270682ed2f77731c8b3133bc664f2213b3c63b1a)

Note that this is just the very first simple idea I had but at least is something.

andreab1997:
Given that this PR is becoming very big and given that the tests are passing I would say that this is a goos moment to merge it. I know that @AleCandido, together with @cschwan, is making an effort  to make the code here simpler but since we are all very busy I believe we can merge this as it is and  we will open another PR once the necessary improvements in `pineappl` have been made. Do you agree?

felixhekhorn (approved):

Issue #81, opened 2023-03-10
Remove `fact_scale` even when using EXPONENTIATED


andreab1997:
Just to be sure, with the new implementation of scale variations methods in EKO we should also avoid the `fact_scale` argument that at the moment is used here https://github.com/NNPDF/pineko/blob/63af934791c9c7794c847513bc7d627f9db3405e/src/pineko/evolve.py#L198

right? @felixhekhorn @AleCandido

felixhekhorn:
yes, but in exchange we should adjust the threshold values https://github.com/NNPDF/pineko/blob/63af934791c9c7794c847513bc7d627f9db3405e/src/pineko/evolve.py#L187
I'm a bit confused by this transformation https://github.com/NNPDF/pineko/blob/63af934791c9c7794c847513bc7d627f9db3405e/src/pineko/evolve.py#L197
because I believe we should do the same on the thresholds - which however is in contradiction with the current implementation in eko https://github.com/NNPDF/eko/blob/935ad228c081fdf8a540c07b6435d5f413c5c9f4/src/eko/runner/legacy.py#L97-L103 ...

Actually, it might be sufficient to swap the condition here:
https://github.com/NNPDF/pineko/blob/63af934791c9c7794c847513bc7d627f9db3405e/src/pineko/evolve.py#L156

andreab1997:
> yes, but in exchange we should adjust the threshold values
> 
> https://github.com/NNPDF/pineko/blob/63af934791c9c7794c847513bc7d627f9db3405e/src/pineko/evolve.py#L187
> 
You mean that if you want to also include resummation scale variations you need to move the thresholds, right? Otherwise they are fine as they are now.

> 
> I'm a bit confused by this transformation
> https://github.com/NNPDF/pineko/blob/63af934791c9c7794c847513bc7d627f9db3405e/src/pineko/evolve.py#L197
> 
> 
> because I believe we should do the same on the thresholds - which however is in contradiction with the current implementation in eko https://github.com/NNPDF/eko/blob/935ad228c081fdf8a540c07b6435d5f413c5c9f4/src/eko/runner/legacy.py#L97-L103 ...
> Actually, it might be sufficient to swap the condition here:
> 
> https://github.com/NNPDF/pineko/blob/63af934791c9c7794c847513bc7d627f9db3405e/src/pineko/evolve.py#L156

For this, 

[...]

PR #82, opened 2023-03-13
Fix sv schemes 

andreab1997:
We want to fix in `pineko` the different sv schemes in such a way it is consistent with `eko`
Solve #81 

CC: @AleCandido @felixhekhorn @giacomomagni

### DIFF
--- a/src/pineko/evolve.py
+++ b/src/pineko/evolve.py
@@ -101,11 +101,7 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
     operators_card = copy.deepcopy(default_card)
     x_grid, _pids, _mur2_grid, muf2_grid = pineappl_grid.axes()
     sv_method = sv_scheme(tcard)
-    xif = (
-        1.0
-        if (sv_method == "expanded" or sv_method == "exponentiated")
-        else tcard["XIF"]
-    )
+    xif = 1.0 if sv_method in ["expanded", "exponentiated"] else tcard["XIF"]
     operators_card["configs"]["scvar_method"] = sv_method
     q2_grid = (xif * xif * muf2_grid).tolist()
     operators_card["_mugrid"] = np.sqrt(q2_grid).tolist()
@@ -157,7 +153,7 @@ def evolve_grid(
     sv_method = None
     if operators.operator_card.configs.scvar_method is not None:
         sv_method = operators.operator_card.configs.scvar_method.name
-    xif = 1.0 if (sv_method == "EXPANDED" or sv_method == "EXPONENTIATED") else xif
+    xif = 1.0 if sv_method in ["EXPANDED", "EXPONENTIATED"] else xif
     tcard = operators.theory_card
     opcard = operators.operator_card
     # rotate the targetgrid
--- a/src/pineko/evolve.py
+++ b/src/pineko/evolve.py
@@ -106,7 +106,7 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
     q2_grid = (xif * xif * muf2_grid).tolist()
     operators_card["_mugrid"] = np.sqrt(q2_grid).tolist()
     if "integrability_version" in pineappl_grid.key_values():
-        x_grid = np.append(x_grid,1.0)
+        x_grid = np.append(x_grid, 1.0)
         operators_card["configs"]["interpolation_polynomial_degree"] = 1
         operators_card["rotations"]["xgrid"] = x_grid.tolist()
 
@@ -158,7 +158,7 @@ def evolve_grid(
     opcard = operators.operator_card
     # rotate the targetgrid
     if "integrability_version" in grid.key_values():
-        x_grid = np.append(x_grid,1.0)
+        x_grid = np.append(x_grid, 1.0)
     eko.io.manipulate.xgrid_reshape(
         operators, targetgrid=eko.interpolation.XGrid(x_grid)
     )
@@ -195,7 +195,6 @@ def evolve_grid(
         * np.pi
         * sc.a_s(
             xir * xir * muf2 / xif / xif,
-            fact_scale=muf2 if sv_method == "EXPONENTIATED" else None,
         )
         for muf2 in muf2_grid
     ]
--- a/src/pineko/evolve.py
+++ b/src/pineko/evolve.py
@@ -101,7 +101,11 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
     operators_card = copy.deepcopy(default_card)
     x_grid, _pids, _mur2_grid, muf2_grid = pineappl_grid.axes()
     sv_method = sv_scheme(tcard)
-    xif = 1.0 if sv_method == "expanded" else tcard["XIF"]
+    xif = (
+        1.0
+        if (sv_method == "expanded" or sv_method == "exponentiated")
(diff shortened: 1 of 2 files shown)


src/pineko/evolve.py
```diff
@@ -101,12 +101,16 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
     operators_card = copy.deepcopy(default_card)
     x_grid, _pids, _mur2_grid, muf2_grid = pineappl_grid.axes()
     sv_method = sv_scheme(tcard)
-    xif = 1.0 if sv_method == "expanded" else tcard["XIF"]
+    xif = (
+        1.0
+        if (sv_method == "expanded" or sv_method == "exponentiated")
```
felixhekhorn, line 106: Pylint is complaining https://github.com/NNPDF/pineko/actions/runs/4405224579/jobs/7715838427#step:10:27

src/pineko/evolve.py
```diff
@@ -157,7 +153,7 @@ def evolve_grid(
     sv_method = None
     if operators.operator_card.configs.scvar_method is not None:
         sv_method = operators.operator_card.configs.scvar_method.name
-    xif = 1.0 if (sv_method == "EXPANDED" or sv_method == "EXPONENTIATED") else xif
+    xif = 1.0 if sv_method in ["EXPANDED", "EXPONENTIATED"] else xif
```
AleCandido, line 156:
```suggestion
    xif = 1.0 if operators.operator_card.configs.scvar_method is not None else xif
```

The check with `None` is definitely sufficient, because an `Optional[ScVarMethod]` can only be three values:
- `None`
- `ScVarMethod.expanded`
- `ScVarMethod.exponentiated`

You can also drop completely the `sv_method` variable from this function.

felixhekhorn (approved):

src/pineko/evolve.py
```diff
@@ -101,11 +101,7 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
     operators_card = copy.deepcopy(default_card)
     x_grid, _pids, _mur2_grid, muf2_grid = pineappl_grid.axes()
     sv_method = sv_scheme(tcard)
-    xif = (
-        1.0
-        if (sv_method == "expanded" or sv_method == "exponentiated")
-        else tcard["XIF"]
-    )
+    xif = 1.0 if sv_method in ["expanded", "exponentiated"] else tcard["XIF"]
```
AleCandido, line 104:
```suggestion
    xif = 1.0 if sv_method is not None else tcard["XIF"]
```

You can assume the runcard has been already validated, so the type is faithful.

However, in general, with enum you can always load them from strings, e.g. with `ScVarMethod["expanded"]`, if it fails is not a member of the enum :)
(but in this case assume is one of the allowed values, and thus check against `None`)

felixhekhorn:
> @felixhekhorn are we officially dropping support for 3.7? (that will be EOL in June 2023, so perfectly fine)
> 
> I just want to know if I can use [new things in 3.8](https://docs.python.org/3.8/whatsnew/3.8.html) safely (e.g. in [`typing`](https://docs.python.org/3.8/whatsnew/3.8.html#typing), of course, since it is the most "unstable" thing I'm using...)

I don't know why @andreab1997 did it here - but [in `eko` we're not running in 3.7 already for a while ...](https://github.com/NNPDF/eko/blob/a9813dc749655858b22fbf01049de56955f00643/.github/workflows/unittests.yml#L9) (and I don't recall the reason there either ...)

andreab1997:
> > @felixhekhorn are we officially dropping support for 3.7? (that will be EOL in June 2023, so perfectly fine)
> > I just want to know if I can use [new things in 3.8](https://docs.python.org/3.8/whatsnew/3.8.html) safely (e.g. in [`typing`](https://docs.python.org/3.8/whatsnew/3.8.html#typing), of course, since it is the most "unstable" thing I'm using...)
> 
> I don't know why @andreab1997 did it here - but [in `eko` we're not running in 3.7 already for a while ...](https://github.com/NNPDF/eko/blob/a9813dc749655858b22fbf01049de56955f00643/.github/workflows/unittests.yml#L9) (and I don't recall the reason there either ...)

I did it here because tests were failing. I believe we should drop 3.7 anyway.

felixhekhorn:
actually, also another slightly off-topic question: why did github run only the tests for 3.10 here? the others are [cancelled](https://github.com/NNPDF/pineko/actions/runs/4417686221/jobs/7744726285#step:10:18) - but by whom? by someone of you, @AleCandido @andreab1997 ?

felixhekhorn:
> I did it here because tests were failing. I believe we should drop 3.7 anyway.

Indeed, it seems we dropped (inconsistently) long ago:
https://github.com/NNPDF/pineko/blob/63af934791c9c7794c847513bc7d627f9db3405e/pyproject.toml#L26

andreab1997:
> > I did it here because tests were failing. I believe we should drop 3.7 anyway.
> 
> Indeed, it seems we dropped (inconsistently) long ago:
> 
> https://github.com/NNPDF/pineko/blob/63af934791c9c7794c847513bc7d627f9db3405e/pyproject.toml#L26

probably my fault but I don't think we should keep it.

andreab1997:
> actually, also another slightly off-topic question: why did github run only the tests for 3.10 here? the others are [cancelled](https://github.com/NNPDF/pineko/actions/runs/4417686221/jobs/7744726285#step:10:18) - but by whom? by someone of you, @AleCandido @andreab1997 ?

not intentionally I guess but let me rerun them

AleCandido:
> actually, also another slightly off-topic question: why did github run only the tests for 3.10 here? the others are [cancelled](https://github.com/NNPDF/pineko/actions/runs/4417686221/jobs/7744726285#step:10:18) - but by whom? by someone of you, @AleCandido @andreab1997 ?

They were failed because of Codecov upload, and I clicked "rerun" I just asked to rerun failed ones, by chance (most likely since I was looking at the one who failed, and from that window I had the option to rerun just that)
image

AleCandido:
> I don't know why @andreab1997 did it here - but [in `eko` we're not running in 3.7 already for a while ...](https://github.com/NNPDF/eko/blob/a9813dc749655858b22fbf01049de56955f00643/.github/workflows/unittests.yml#L9) (and I don't recall the reason there either ...)

I confused myself: I knew it was Pineko, but my mind was keeping thinking about EKO...

felixhekhorn:
> > actually, also another slightly off-topic question: why did github run only the tests for 3.10 here? the others are [cancelled](https://github.com/NNPDF/pineko/actions/runs/4417686221/jobs/7744726285#step:10:18) - but by whom? by someone of you, @AleCandido @andreab1997 ?
> 
> not intentionally I guess but let me rerun them

Mmm, now we are failing with CodeCov, but before we were not - and that is the strange thing! I guess there is maybe a hook in the workflows, if one fails to kill all (although that might be dangerous with CodeCov ...)

> They were failed because of Codecov upload, and I clicked "rerun" I just asked to rerun failed ones, by chance (most likely since I was looking at the one who failed, and from that window I had the option to rerun just that)

fine, also in "Attempt \#1" we failed so you reran - but in "Attempt \#2" we have two successes and two cancellations - this I don't understand

andreab1997:
Can we merge this now?

AleCandido (approved):

PR #86, opened 2023-03-24
Adding regression tests to Pineko

scarlehoff:
This test will serve two purposes:

1. Ensure that our "default" theory is always working (which is theory 400) in the way the user is supposed to use it 
2. Ensure that the theories repository is in sync with this repository.

(rn just testing how long it takes in the CI / whether it works at all)


scarlehoff:
Ok, this is ready for review now. It takes ~5 minutes which I think it is reasonable. It can be made to take a bit less or more by playing with the x-grid.

I've chosen a double-hadronic dataset which contains 3 fktables so that it is not too trivial as a check but also all three grids are fairly small, so it is not a very heavy test.

I can add other datasets (the trade-off is that each will be tested to a lower-precision because I will have to reduce the size of the grid). Any other considerations or things that I should add to the test?

At the moment the check is checking that the grid to -> fktable conversion generates an object that, upon PDF-convolution, produces the exact same result.

tests/test_regression.py
```diff
@@ -0,0 +1,123 @@
...
+        pids[6] = 21
+
+        alphas = np.linspace(1.2, 1.8, len(pids))
+        betas = np.linspace(1.2, 3.8, len(pids))
+
+        self._alphas = dict(zip(pids, alphas))
+        self._betas = dict(zip(pids, betas))
+
+    def xfxQ2(self, pid, x, q2):
+        """Compute x^alpha*(1-x)^alpha"""
```
felixhekhorn, line 60:
```suggestion
        """Compute x^alpha*(1-x)^beta"""
```

felixhekhorn:
- as already said, I would add a test to this branching https://github.com/NNPDF/pineko/blob/1664a9121115f16b5da276fc132818297109d81a/src/pineko/evolve.py#L108 because this already caused problems 
- else, this is a good start and for sure better than what we have so far (=0) - in the future we may want to add also the other options/commands (`inherit-grids` or specifying several datasets at the same time)

scarlehoff:
> as already said, I would add a test to this branching

You mean, to test it also with a integrability dataset? That should be "for free" actually because they are so small.

tests/test_regression.py
```diff
@@ -0,0 +1,124 @@
...
+THEORYID = 400
+
+
+def _download_resources(filename, tmp_path):
+    """Download resources (filename) from the theories repo and put it in the same path
+    relative to THEORIES_REPO
+    """
+    output_file = tmp_path / filename
+    output_file.parent.mkdir(exist_ok=True)
+    urlretrieve(f"{THEORIES_REPO}/{filename}", output_file)
```
AleCandido, line 28: Fine by me, I appreciate you tried to stick to stdlib. But there would be nothing wrong to add a `requests` dependency as the rest of the world (especially if you're planning a `pineko download`, so it won't even be just a test dep).

tests/test_regression.py
```diff
@@ -0,0 +1,124 @@
...
+from pathlib import Path
+from subprocess import run
+from urllib.request import urlretrieve
+
+import numpy as np
+import pytest
+from pineappl.fk_table import FkTable
+from yaml import dump, safe_load
+
+THEORIES_REPO = "https://raw.githubusercontent.com/NNPDF/theories/main"
```
AleCandido, line 18: What is the current plan about this repo? We gave up on storing EKOs, so what are we storing there?

tests/test_regression.py
AleCandido, line 28: In principle, URLs are like paths, at least the way you're using they are actually `PurePosixPath`s, since they are locations inside `THEORIES_REPO` (they are pure because you can not operate on them, e.g. you can not ask if a file exists, or create a directory).

If you wish, you could use that, as a better abstraction over strings.

tests/test_regression.py
```diff
@@ -0,0 +1,124 @@
...
+    urlretrieve(f"{THEORIES_REPO}/{filename}", output_file)
+    return output_file
+
+
+def _download_dataset(dataset, theoryid, tmp_path):
+    """Download both the yaml file and all grids for a given dataset for a given theory"""
+    yaml_file = f"data/yamldb/{theoryid}/{dataset}.yaml"
+    _download_resources(yaml_file, tmp_path)
+    # However, the yaml file goes into the "root" of the yaml directory, move it there!
+    right_yaml = tmp_path / "data" / "yamldb" / f"{dataset}.yaml"
```
AleCandido, line 37: If `yaml_file` were a path, this could have been:
```suggestion
    right_yaml = tmp_path / yaml_file.parents[1] / yaml_file.name
```

tests/test_regression.py
```diff
@@ -0,0 +1,124 @@
...
+    (tmp_path / yaml_file).rename(right_yaml)
+    # Download the relevant grids for this dataset
+    res = safe_load(right_yaml.open("r", encoding="utf-8"))
```
AleCandido, line 40:
```suggestion
    res = safe_load(right_yaml.read_text(encoding="utf-8"))
```

tests/test_regression.py
```diff
@@ -0,0 +1,124 @@
...
+    grids = [i for j in res["operands"] for i in j]
```
AleCandido, line 41:
```suggestion
    grids = list(itertools.chain(*res["operands"]))
```
Personally, I prefer more explicit words, rather than more basic ones. But I guess it is a matter of taste...

tests/test_regression.py
```diff
@@ -0,0 +1,124 @@
...
+    for grid_name in grids:
+        _download_resources(f"data/grids/400/{grid_name}.pineappl.lz4", tmp_path)
+    return grids
+
+
+class _FakePDF:
+    """A Fake lhapdf-like PDF to evolve the grids"""
```
AleCandido, line 48: We have the same object in EKO's tests (not the same function, but same idea). I wonder if we should start having a `pdftest` library...

But I know I should not overengineer things...

tests/test_regression.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def __init__(self):
+        pids = np.arange(-6, 7, dtype=int)
+        pids[6] = 21
```
AleCandido, line 52: You might have forgotten the photon. Maybe in this case is not relevant, but since we know it might be used, I'd put it.
What do you think?

tests/test_regression.py
```diff
@@ -0,0 +1,124 @@
...
+    for grid_name in grids:
+        _download_resources(f"data/grids/400/{grid_name}.pineappl.lz4", tmp_path)
+    return grids
+
+
+class _FakePDF:
+    """A Fake lhapdf-like PDF to evolve the grids"""
+
+    def __init__(self):
+        pids = np.arange(-6, 7, dtype=int)
```
AleCandido, line 51:
```suggestion
        pids = np.arange(-6, 7)
```
Nothing wrong, but maybe you're just too used to tf and Numba :)

tests/test_regression.py
```diff
@@ -0,0 +1,124 @@
...
+    def xfxQ2(self, pid, x, q2):
+        """Compute x^alpha*(1-x)^beta"""
+        alpha = self._alphas[pid]
+        beta = self._betas[pid]
+        return np.power(x, alpha) * np.power(1 - x, beta)
+
+
+def _trim_template(template_card, take_points=10):
+    """Trim the template card so that the number of x-values to compute is much smaller"""
+    card_info = safe_load(template_card.open("r", encoding="utf-8"))
```
AleCandido, line 69:
```suggestion
    card_info = safe_load(template_card.read_text(encoding="utf-8"))
```

AleCandido: If you can, I'd suggest using the native EKO's runcards object, in this case `eko.io.runcards.OperatorCard`.
 ```suggestion
     card_info = OperatorCard(**safe_load(template_card.read_text(encoding="utf-8")))
```

It automatically validates the runcard, and gives you an explicit interface (useful also for autocompletion xD).

File: tests/test_regression.py [hunk 11]
```diff
@@ -0,0 +1,124 @@
...
+    original_x = np.array(card_info["rotations"]["xgrid"])
+    skip = int(len(original_x) / take_points)
+    idxs = list(range(0, len(original_x), skip))
+    card_info["rotations"]["xgrid"] = original_x[idxs].tolist()
```
AleCandido, line 73:
```suggestion
    size = len(original_x)
    card_info["rotations"]["xgrid"] = original_x[:size:skip].tolist()
```
I find it a bit cleaner, but of course it's a detail.

tests/test_regression.py
```diff
@@ -0,0 +1,124 @@
...
+    dump(card_info, template_card.open("w", encoding="utf-8"))
```
AleCandido, line 74:
```suggestion
    template_card.write_text(dump(card_info), encoding="utf-8")
```

tests/test_regression.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+@pytest.mark.parametrize("dataset", ["LHCBWZMU8TEV", "INTEGXT3"])
+def test_regression(tmp_path, dataset, rebuild=False):
```
AleCandido, line 78: What is `rebuild` used for?

Not being a fixture, nor a parameter, how can you use it? (Most likely I'm just ignorant myself...)

tests/test_regression.py
```diff
@@ -0,0 +1,124 @@
...
+    """Run pineko through subprocess to ensure that the shell scripts are working exactly
+    as intended.
+    """
+    # We start by downloading pineko.toml in order to generate the folder structure
+    _download_resources("pineko.toml", tmp_path)
+    # Which we create... now!
+    run(["pineko", "scaffold", "new"], cwd=tmp_path, check=True)
```
AleCandido, line 85: Pineko functions can be run programmatically, without the need of `subprocess`es.
The CLI has been designed to be a thin wrapper over functions (as all the others in the pineline, starting with Pinefarm), such that you are actually able to script them, within Python.

If this is not possible, I'd ask you the favor of filing an issue!

tests/test_regression.py
```diff
@@ -0,0 +1,124 @@
...
+    # Now go, first with eko creation
+    run(
+        ["pineko", "theory", "opcards", str(THEORYID), dataset],
+        cwd=tmp_path,
+        check=True,
+    )
+    run(["pineko", "theory", "ekos", str(THEORYID), dataset], cwd=tmp_path, check=True)
+
+    # Then FK Table production!
+    run(["pineko", "theory", "fks", str(THEORYID), dataset], cwd=tmp_path, check=True)
```
AleCandido, line 106: See above :)

I'm asking this also because it makes easier refactoring. If ever we will change something, the static analyzer knows functions, and can help us to update them consistently.
These instead are just strings, and it makes harder to discover them ("pineko", "theory", "ekos", and "fks" are used all over the place...).

tests/test_regression.py
```diff
@@ -0,0 +1,124 @@
...
+
+    # Now loop over the grids and check the results of the convolution with the PDF
+    pdf = _FakePDF()
+    regression_path = Path(__file__).parent / "regression_data" / f"{dataset}.npy"
```
AleCandido, line 110: Can you please put the folder path in a top-level constant? Thanks :)

tests/test_regression.py
```diff
@@ -0,0 +1,124 @@
...
+        fkt = FkTable.read(
+            tmp_path / "data" / "fktables" / str(THEORYID) / f"{grid_name}.pineappl.lz4"
+        )
+        result.append(fkt.convolute_with_one(2212, pdf.xfxQ2))
+    result = np.concatenate(result)
+
+    if rebuild:
+        np.save(regression_path, result)
+
+    regression_data = np.load(regression_path)
```
AleCandido, line 123: I see, but then I'd split the content of the test in a separate function, instead of using the test itself to create new metadata, and here stick to computed data.

I fear that the argument is non-standard, and if you forget that default toggled, we might inadvertently run meaningless regressions, until we will find out.

AleCandido (commented):
Mostly fine, most of the suggestions are stylistics and a matter of taste.
The only one I'm actually more concerned about is the `subprocess` one. 

However, from my point of view, this is almost good to go. Thank you very much! :)

tests/test_regression.py
scarlehoff: It stores anything necessary to run pineko and create new theories. So basically the template cards (and also the cards, because they are cheap). The theory cards, the yaml... and the grids. Only the grids will be "a reference to the grid" as we discussed to avoid storing big files in github but we know we will never do it.

tests/test_regression.py
```diff
@@ -0,0 +1,124 @@
+"""
+    Suite of tests that go through the entire process of creating a new fktable
+    from a empty folder.
+
+    The target theory is 400 and the relevant `.toml`, theory runcard and eko template
+    are downloaded from https://github.com/NNPDF/theories during this test so this tests
+    has the double effect of ensuring compatibility between both repositories.
+"""
+from pathlib import Path
```
scarlehoff, line 9:
```suggestion
import itertools
from pathlib import Path
```

tests/test_regression.py
AleCandido: > but we know we will never do it.

I actually agree xD

But I believe that even the grids will be inconsistently stored (since you don't want to do it for any throwaway work-in-progress theory).

Concerning the others:
- we are approaching the point in which the yaml-db will be inside the new Commondata (slowly approaching...)
- the theory cards at some point should be an interface to the theory db, because we definitely want to be consistent with it (we just need a dedicated effort to make the interface, and maybe even improving the DB itself)

So, I feel like NNPDF/theories is a temporary solution (that might last quite a while, because it is not a priority to drop it), but that's fine :)

tests/test_regression.py
AleCandido: ?

tests/test_regression.py
scarlehoff: I had indeed. I'll add it. Should change nothing.

tests/test_regression.py
```diff
@@ -0,0 +1,124 @@
...
+from subprocess import run
+from urllib.request import urlretrieve
+
+import numpy as np
```
scarlehoff, line 13:
```suggestion
from eko.io.runcards import OperatorCard
import numpy as np
```

tests/test_regression.py
scarlehoff: This is exactly what I want. If that happens then I know that all the documentation / my own bash scripts / etc are going to break!

tests/test_regression.py
```diff
@@ -0,0 +1,126 @@
...
+        return np.power(x, alpha) * np.power(1 - x, beta)
+
+
+def _trim_template(template_card, take_points=10):
+    """Trim the template card so that the number of x-values to compute is much smaller"""
+    card_info = OperatorCard(**safe_load(template_card.read_text(encoding="utf-8")))
+    original_x = np.array(card_info["rotations"]["xgrid"])
+    skip = int(len(original_x) / take_points)
+    size = len(original_x)
+    card_info["rotations"]["xgrid"] = original_x[:size:skip].tolist()
```
AleCandido, line 75:
```suggestion
    card = OperatorCard(**safe_load(template_card.read_text(encoding="utf-8")))
    original_x = np.array(card.rotations.xgrid)
    skip = int(len(original_x) / take_points)
    size = len(original_x)
    card.rotations.xgrid = original_x[:size:skip].tolist()
```

tests/test_regression.py
AleCandido: Not a fan of doing two tasks in a single shot: you're testing numbers here, so I'd test the CLI somewhere else.

But if you believe to be better like this, because it reproduces a full user run, I'll accept your point of view :)

scarlehoff: I'm testing numbers because is the only thing that came to mind, but they are not the point. It is an "user run test" indeed.

AleCandido (approved):
At some point, I would consider moving this out of unit tests (since it is definitely not *unit*).

However, there is no hurry, we can do at a later stage :)

PR #88, opened 2023-04-19
Eko v0.13.0 compatibility

giacomomagni:
This PR is for eko compatibility with ~~`^0.13`~~ `0.13.0`


src/pineko/evolve.py
```diff
@@ -113,11 +116,21 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
     xif = 1.0 if sv_method is not None else tcard["XIF"]
     operators_card["configs"]["scvar_method"] = sv_method
     q2_grid = (xif * xif * muf2_grid).tolist()
-    operators_card["_mugrid"] = np.sqrt(q2_grid).tolist()
+    atlas = ThresholdsAtlas(
+        masses=np.array([tcard["mc"], tcard["mb"], tcard["mt"]]) ** 2,
+        q2_ref=tcard["Q0"],
```
giacomomagni, line 121:
```suggestion
        q2_ref=tcard["Q0"] ** 2,
```
@andreab1997 I'm not sure if this means I screwed some ekos...

andreab1997: Yes, but I believe this affects everything I fear

andreab1997: Ah no, I am sorry, you are right. It should only affect the points between `1.65` and `1.65**2` which we do not fit.

AleCandido:
Since we have been bad[*] at stability with EKO, we decided to start from the very basics, and to limit the public API to runcards and the `solve` function.

The result is that Pineko depends on EKO internals, so there might be (small) breaking changes even in patch releases.
If we declare immediately public all what is needed for Pineko, the risk is to never achieve stability, since we'd limit a lot ourselves, just before a round of major refactoring. However, since we know Pineko, we care about breaking little, and we will try to prioritize the publication of the API required. While, in Pineko, we will try to limit EKO API usage as much as possible.

[*]: because of fast development, bugs fixing, and lack of resources, not because we don't care

andreab1997:
What is missing here is mainly fixing the tests and the benchmarks. However, since fixing the benchmarks requires changing the benchmark files and since the files can be correct either for this PR or for #75, not for both, I would like first to merge #75 (in such a way tests and benchmarks there work) and only then rebase this on `main` and break the benchmarks again. 
So, just to remember, I am switching this to draft. @felixhekhorn @AleCandido @giacomomagni

AleCandido:
Just to write it here for the benefit of everyone: I proposed to actually fix the benchmarks for this PR, since it has a greater priority and it is partially blocking some other tasks (or the completion of this might simplify something else).

PR #99, opened 2023-06-09
Numerical FONLL

RoyStegeman:
Based on our discussion added the functionality that if pineko is called to produce opcards for a theory of which the flavor number scheme is FONLL, pineko will produce the necessary new theory cards by appending 1,2,3 to the end of the "input" theory id. 

This may create problems with storing the intermediate grids since scale variations uses the same numbering, though I'm not sure how important that really is? Based on https://github.com/NNPDF/theories/tree/main/data/grids you did save them for scale variations, but since this will only be for DIS maybe that's not recommended? Alternatively, I could of course just append different numbers but that requires some awareness/coordination of which are already in use.


src/pineko/check.py
```diff
@@ -131,12 +153,7 @@ def is_fonll_b(fns, lumi):
     bool
         true if the fktable is a FONLL-B DIS fktable
     """
-    for lists in lumi:
-        for el in lists:
-            if (not islepton(el[0])) and (not islepton(el[1])):
-                # in this case we are sure it is not DIS so for sure it is not FONLL-B
-                return False
-    if fns == "FONLL-B":
+    if is_dis(lumi) and fns == "FONLL-B":
```
AleCandido, line 156: I wonder if we still need an explicit check for FONLL-B @felixhekhorn 

For sure, we need to map FONLL-A/B/C on the correct massive (FO) and massless (resummed) order, but we might just need something like:
```py
fonll_schemes = {"A": (1, 1), "B": (1, 2), "C": (2, 2)}
```

src/pineko/theory.py
```diff
@@ -224,22 +225,62 @@ def opcard(self, name, grid, tcard):
...
+                    {**tcard, "ID": theoryid + "1", "FNS": "FFNS", "NfFF": 3},
+                    {**tcard, "ID": theoryid + "2", "FNS": "FFN0", "NfFF": 3},
+                    {**tcard, "ID": theoryid + "3", "FNS": "FFNS", "NfFF": 4},
+                ]
+            else:
+                tcards = [
+                    {**tcard, "ID": theoryid + "1", "FNS": "FFNS", "NfFF": 3},
+                    {**tcard, "ID": theoryid + "2", "FNS": "FFN0", "NfFF": 3},
+                    {**tcard, "ID": theoryid + "3", "FNS": "FFNS", "NfFF": 4},
+                ]
```
AleCandido, line 254: At the moment, these two branches seem defining the exact same value.

RoyStegeman: They do, they're just FONLL with the PTO of the input theory as placeholder for the different FONLL schemes that I did not yet want to think about. In particular I still need to look at how I define the difference between the pto for evolution and the coefficients since that's relevant for FONLL-B. This distinction is no longer needed in yadism but I need a call to eko with a different pto for evolution than was used for the calculation is yadism, which is something I still need to figure out how (haven't really looked at pinefarm/eko yet). This is also what the last `TODO` in above is referencing.

RoyStegeman:
Since these are only for internal use, you can also make a tuple: `(theoryid, 1)`, `(theoryid, 2)`, ...
They will be consumed before coming back to the NNPDF world ^^

If we just use numbering we still need to coordinate though, I don't see so much the benefit from doing `(theoryid,1)` instead of `f"{theoryid}1". Anyway, I'd rather use integers since most/all codes expect integer theoryIDs even if they are just used for internal steps and I don't want to risk unexpected behaviours by allowing non-integer inputs.

RoyStegeman:
I converted it to draft since I opened the PR do discuss about some details first, not to merge immediatly.

src/pineko/theory.py
AleCandido: I would suggest to start directly producing an EKO theory for it, that will require moving the current translator from EKO in here.
https://github.com/NNPDF/eko/blob/3db8876f26de5c22bd59a90163f47eb0b232955a/src/eko/io/runcards.py#L168-L209

At that point, you will really have two different theories for EKO and pinefarm.

AleCandido:
> I don't see so much the benefit from doing `(theoryid,1)` instead of `f"{theoryid}1".

I see it: `f"{theoryid}1" could actually be a different theory (Andrea and Giacomo are using 4-digits theories already, and maybe even 5). Moreover, `id[1]` on a pair is a bit cleaner than `id[-1]` on a string, since:
- it will be already an integer (rather than a char)
- in the (currently?) absurd case of >10 theories generated, it would still work

RoyStegeman:
Ah, I think I misunderstood your proposal. Where would this tuple be stored and read and what would be the `ID` key in the theory card? 

All code assumes that theoryIDs are integers, so are you proposing to use `id[1]` as the integer passed to identify? Then we would get e.g. the folder `NNPDF/theories/data/id[1]` which is of course also not unique (I'm sure this is not what you have in mind, but I don't understand what you do ;) ).

src/pineko/theory.py
RoyStegeman: Thanks, that would make sense.

RoyStegeman:
> What I was arguing is that the only one who cares about IDs is Pineko itself. In the pineline, no one else is reading IDs (and in case Pinefarm does, it should just propagate them straight, but we could check).

Ah okay this I didn't realise, but that might make things a lot simpler. The original reason for my question though, was that if it is useful to store not only the final FKtables but also the grids that are produced during the intermediate step I need a unique folder name to not clash with Andrea (and maybe Giacomo), which I don't think your suggestion solves? Since it's only pineko that deals with theoryid's I could even add a more descriptive string or subfolder structure without needing to worry.

Note that even with the current system as I suggested we can do more than 9 theories since there is nothing stopping us from creating a theorid of 40011 - we don't read the id from the string using `id[-1]` but 40011 would be the full id. Again, my worry was purely about preventing clashes if the intermediate grids should be saved for the long term.

AleCandido:
> Note that even with the current system as I suggested we can do more than 9 theories since there is nothing stopping us from creating a theorid of 40011 - we don't read the id from the string using `id[-1]` but 40011 would be the full id. Again, my worry was purely about preventing clashes if the intermediate grids should be saved for the long term.

In principle, you'd be right, we might want to store the grids we computed.

In practice, this is not yet a problem, since only FONLL requires multiple grids, and FONLL is only applied to DIS, for which recomputing is potentially easier than storing.

However, even with a more complex structure it would still work, because you could save `(400, 1)`, `(400, 2)`, ... all as part of `400` (in case we stored grids by theories, that does not work in the long term).
Moreover, if you wish you can distinguish with a different key, you don't even have to use `ID:`. Something like `fragment:` or similar. In the end, it will be wiped out when you construct the theory for fitting (since only one joint FK table will survive).

RoyStegeman:
> In practice, this is not yet a problem, since only FONLL requires multiple grids, and FONLL is only applied to DIS, for which recomputing is potentially easier than storing.

There already exists grids for theoryids 4001 and 4005, not exactly sure why since sv should correspond to different theories for nnpdf as well but regardless, it's not just fonll. 

> Moreover, if you wish you can distinguish with a different key, you don't even have to use `ID:`. Something like `fragment:` or similar. In the end, it will be wiped out when you construct the theory for fitting (since only one joint FK table will survive).

If you agree (also @andreab1997) I'll only save intermediate results locally for now (also the ekos) since it's DIS that shouldn't cause too much trouble if we ever need to reproduce them. What we could do if we do want to save is just add subfolders for fonll-a/b/c with inside the intermediate theorycards and grids. In this way it doesn't matter so much if we use theoryid 4001 or (400,1) since for fonll we won't need more than 9, but I do think (400,1) is a bit cleaner and also immediately signals that it's not a theory that directly corresponds to an fktable or "nnpdf theory".

felixhekhorn:
> but 40011 would be the full id

I think, creating new "NNPDF theories" could work - of course the naming has to be more complicated than just adding a number for the reason @AleCandido said ... so e.g. $10^7 + \mathrm{theoryid}*10^4 + 1$ might work. Those theories are viable physical theories after all (e.g. FFNS3) ... the complication then is how to rejoin them together since you need them when merging ...

> which recomputing is potentially easier than storing.

mmm CHORUS ekos are (very) expensive ... so we want to save intermediate results

RoyStegeman:
> I think, creating new "NNPDF theories" could work - of course the naming has to be more complicated than just adding a number for the reason @AleCandido said ... so e.g. $10^7 + \mathrm{theoryid}*10^2 + 1$ might work. 

Sure, but there would still need to be some sort of system to it. Keeping a database might be a bit excessive but we do need to ensure that I'm not occupying the same identifiers for fonll (to remain abstract as to how exactly we'll do this) as Andrea and Giacomo do for their stuff.

> Those theories are viable physical theories after all (e.g. FFNS3) ... the complication then is how to rejoin them together since you need them when merging ...

FFN0 is less viable, but I see your point. The rejoining and damping I plan to do at the level of fktables. I haven't thought out the exact details yet, but as long as we know which theory/grids is which I don't foresee any problems at this step, do you?

src/pineko/fonll.py
```diff
@@ -90,9 +90,8 @@ def produce_combined_fk(
     fonll_info = FONLLInfo(ffns3, ffn03, ffns4til, ffns4bar, ffn04, ffns5til, ffns5bar)
 
     theorycard_constituent_fks = fonll_info.theorycard_no_fns_pto
-    if theorycard_constituent_fks["DAMP"] == 0:
+    if theorycard_constituent_fks["DAMP"] == "False":
```
AleCandido, line 93: This is weird... why a string?

andreab1997: You are right, fixed here [3ca875c](https://github.com/NNPDF/pineko/pull/99/commits/3ca875cd48f5b0e72befed2991924f1d45ac0b78)

RoyStegeman:
As I'm now trying this as a user I'll add anything I notice (maybe it will just be one thing):

- Please add an info message where the tcards are stored when running `pineko fonll_tcards [theoryID]`

src/pineko/fonll.py
```diff
@@ -182,6 +182,8 @@ def produce_combined_fk(
     "FONLL-A": [1 for _ in range(7)],
     "FONLL-B": [2, 2, 1, 2, 2, 1, 2],
     "FONLL-C": [2 for _ in range(7)],
+    "FONLL-D": [3, 3, 2, 3, 3, 2, 3],
+    "FONLL-E": [3 for _ in range(7)],
```
AleCandido, line 186: All the lists above could be generated from a single tuple:

```py
def components_orders(fo, res):
    return [fo, fo, resum, fo, fo, resum, fo]
```

But I'm not sure about the meaning of `"full"` (because I suspect you do not even need the 7 elements).

andreab1997: Why not?

AleCandido: Because there is a pattern, and I guess I know where this pattern is coming from.

andreab1997: No wait my question was, why do you think we do not even need the 7 elements?

AleCandido: Because there is a pattern, and in that case you might be able to make a function respecting that pattern, without making a list.

src/pineko/cli/fonll.py
```diff
@@ -0,0 +1,154 @@
...
+        if new_fk_path.exists():
+            if not overwrite:
+                rich.print(
+                    f"[green]Success:[/] skipping existing FK Table {new_fk_path}"
+                )
+                return
+        fonll.produce_combined_fk(
+            configs.configs["paths"]["fktables"] / str(ffns3) / grid
+            if (configs.configs["paths"]["fktables"] / str(ffns3) / grid).exists()
+            else None,
```
felixhekhorn, line 112: definitely too much repetition, I'd say
```py
def load_cfg(name):
    return configs.configs["paths"]["fktables"] / name / grid            if (configs.configs["paths"]["fktables"] / name / grid).exists()            else None
```
and then
```py
*(load_cfg(str(var)) for var in (ffns3, ffn03))
```

src/pineko/check.py
```diff
@@ -131,12 +153,14 @@ def is_fonll_b(fns, lumi):
...
+        return True
+    return False
+
+
+def issimFONLL(FNS):
+    """Check if the FNS is a simFONLL FNS."""
+    if FNS in ["FONLL-FFNS", "FONLL-FFN0"]:
         return True
     return False
```
felixhekhorn, line 165:
```suggestion
    return FNS in ["FONLL-FFNS", "FONLL-FFN0"]
```

src/pineko/check.py
felixhekhorn: I guess with the new strategy of @andreab1997 the answer is no (since now there is an explicit spellout)

src/pineko/evolve.py
```diff
@@ -106,6 +106,7 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
     """
     # Add a +1 to the orders for the difference in convention between nnpdf and pineappl
     is_fns = int(check.is_fonll_b(tcard["FNS"], pineappl_grid.lumi()))
+    # NB: This would not happen for simFONLL
```
felixhekhorn, line 109: since the name `simFONLL` never went outside maybe let's revert to `nFONLL`?

src/pineko/cli/fonll.py
AleCandido: Even better:
```py
def load_cfg(name):
    path = configs.configs["paths"]["fktables"] / name / grid
    return  path if path.exists() else None
```

src/pineko/fonll.py
andreab1997: Okok I agree about this of course. I believe I misunderstood your first comment.

src/pineko/evolve.py
andreab1997: [c718a1f](https://github.com/NNPDF/pineko/pull/99/commits/c718a1f4c7b3d352c7ed365c7cac52196b00fe31)

src/pineko/cli/fonll.py
andreab1997: [c718a1f](https://github.com/NNPDF/pineko/pull/99/commits/c718a1f4c7b3d352c7ed365c7cac52196b00fe31)

src/pineko/check.py
```diff
@@ -131,16 +153,16 @@ def is_fonll_b(fns, lumi):
...
-            if (not islepton(el[0])) and (not islepton(el[1])):
-                # in this case we are sure it is not DIS so for sure it is not FONLL-B
-                return False
-    if fns == "FONLL-B":
+    if is_dis(lumi) and fns == "FONLL-B":
         return True
     return False
 
 
+def issimFONLL(FNS):
```
felixhekhorn, line 161: there is still lot's of `simFONLL` around (Firefox counts 18 occurences) - while changing, please make `snake_case`

src/pineko/theory.py
```diff
@@ -388,6 +388,7 @@ def fk(self, name, grid_path, tcard, pdf):
             grid.lumi(),
         ):
             max_as += 1
+        # NB: This would not happen for simFONLL
```
AleCandido, line 391: Should we also use *numerical FONLL* everywhere, to avoid the double nomenclature?

andreab1997: Yes I am changing all the simFONLL to num_fonll now

AleCandido (commented):
Some comments

src/pineko/check.py
```diff
@@ -116,6 +116,28 @@ def check_grid_and_eko_compatible(pineappl_grid, operators, xif, max_as, max_al)
...
+    lumi : list(list(tuple))
+           luminosity info
+
+    Returns
+    -------
+    bool
+        true if the fktable is a DIS fktable
+    """
+    for lists in lumi:
+        for el in lists:
```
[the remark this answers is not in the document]
andreab1997: [1b310eb](https://github.com/NNPDF/pineko/pull/99/commits/1b310eb9f06215365a6e31efc59ef13be22d7c37)

src/pineko/evolve.py
```diff
@@ -128,9 +129,14 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
         matching_scales=heavy_quarks.MatchingScales(masses * thresholds_ratios),
         origin=(tcard["Q0"] ** 2, tcard["nf0"]),
     )
-    operators_card["mugrid"] = [
-        (float(np.sqrt(q2)), int(nf_default(q2, atlas))) for q2 in q2_grid
-    ]
+    # If we are producing simFONLL FKs we need to look to NfFF...
+    if check.issimFONLL(tcard["FNS"]):
+        nfff = tcard["NfFF"]
+        operators_card["mugrid"] = [(float(np.sqrt(q2)), int(nfff)) for q2 in q2_grid]
```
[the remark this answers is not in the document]
andreab1997: [1b310eb](https://github.com/NNPDF/pineko/pull/99/commits/1b310eb9f06215365a6e31efc59ef13be22d7c37)

src/pineko/check.py
```diff
@@ -131,16 +153,16 @@ def is_fonll_b(fns, lumi):
...
-                return False
-    if fns == "FONLL-B":
+    if is_dis(lumi) and fns == "FONLL-B":
         return True
     return False
 
 
+def issimFONLL(FNS):
+    """Check if the FNS is a simFONLL FNS."""
+    return FNS in ["FONLL-FFNS", "FONLL-FFN0"]
```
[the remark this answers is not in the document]
andreab1997: I don't know if I got your point but the prefix `FONLL` is used by yadism and that is why it is in the theory card (which is what this test is checking). Of course we could drop it but we would need to do it consistently in yadism.

src/pineko/evolve.py
```diff
@@ -128,9 +129,14 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
...
+    else:
+        operators_card["mugrid"] = [
+            (float(np.sqrt(q2)), int(nf_default(q2, atlas))) for q2 in q2_grid
```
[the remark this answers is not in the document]
andreab1997: Yes you are right, https://github.com/NNPDF/eko/blob/3e6246d425affc03a44f5fb4544354b1cc0e7baa/src/eko/matchings.py#L188

src/pineko/check.py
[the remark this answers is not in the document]
AleCandido: I know, the proposal was dropping in both :)

The question is whether we need to distinguish (in both) `"FFNS"` from `"FONLL-FFNS"`, and why :)

src/pineko/cli/fonll.py
```diff
@@ -0,0 +1,163 @@
...
+    default=None,
+    type=click.Path(resolve_path=True, path_type=pathlib.Path),
+    help="Explicitly specify config file (it has to be a valid TOML file).",
+)
+
+
+class TheoryCardError(Exception):
+    """Raised when asked for FONLL theory cards with an original tcard as input that is not asking for FONLL."""
+
+    pass
```
AleCandido, line 23:
```suggestion
```

Stupid remark: you do not need `pass`, it's enough to have a docstring :)
(there has to be something in, and a docstring is something)

src/pineko/cli/fonll.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+class InconsistentInputsError(Exception):
+    """Raised if the inputs are not consistent with FONLL."""
+
+    pass
```
AleCandido, line 29:
```suggestion
```

src/pineko/cli/fonll.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+def load_cfg(name, grid):
```
AleCandido, line 32:
```suggestion
def cfgpath(name, grid):
```

src/pineko/cli/fonll.py
```diff
@@ -0,0 +1,163 @@
...
+    """Path of the fktable in 'name' called 'grid' if it exists, else None."""
+    path = configs.configs["paths"]["fktables"] / name / grid
+    return path if path.exists() else None
+
+
+def get_list_grids_name(yaml_file):
```
AleCandido, line 38:
```suggestion
def grids_names(yaml_file):
```

src/pineko/cli/fonll.py
```diff
@@ -0,0 +1,163 @@
...
+@click.argument("dataset", type=str)
+@click.option("--FFNS3", type=int, help="theoryID containing the ffns3 fktable")
+@click.option("--FFN03", type=int, help="theoryID containing the ffn03 fktable")
+@click.option("--FFNS4", type=int, help="theoryID containing the ffns4 fktable")
+@click.option("--FFNS4til", type=int, help="theoryID containing the ffns4til fktable")
+@click.option("--FFNS4bar", type=int, help="theoryID containing the ffns4bar fktable")
+@click.option("--FFN04", type=int, help="theoryID containing the ffn04 fktable")
+@click.option("--FFNS5", type=int, help="theoryID containing the ffns5 fktable")
+@click.option("--FFNS5til", type=int, help="theoryID containing the ffns5til fktable")
+@click.option("--FFNS5bar", type=int, help="theoryID containing the ffns5bar fktable")
```
AleCandido, line 60: Instead of a thousand options, couldn't we just use a list of integers with some encoding?

src/pineko/cli/fonll.py
```diff
@@ -0,0 +1,163 @@
...
+
+@command.command("fonll_tcards")
+@click.argument("theoryID", type=int)
+@config_setting
+def fonll_tcards(theoryid, cfg):
+    """Produce the FONLL tcards starting from the original tcard given by the theoryID."""
+    path = configs.detect(cfg)
+    base_configs = configs.load(path)
+    configs.configs = configs.defaults(base_configs)
+    if cfg is not None:
```
AleCandido, line 157: This print only checks a function parameter: shouldn't it depend on `configs.configs`?

src/pineko/cli/fonll.py
```diff
@@ -0,0 +1,163 @@
...
+        print(f"Configurations loaded from '{path}'")
+    tcard = theory_card.load(theoryid)
+    tcard_parent_path = theory_card.path(theoryid).parent
+    if "FONLL" not in tcard["FNS"]:
+        raise TheoryCardError
```
AleCandido, line 162: Do you really need this check here?

To avoid proliferating code it is useful to check as late as possible (or even better: never - just pretend it's working, until it will raise an error on its own)

AleCandido (commented):
Just missing a single file (the big one, of course) :)

src/pineko/check.py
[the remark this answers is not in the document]
andreab1997: Ok then I agree with you: we definetely can drop `FONLL` from both (and we probably should). Do you agree @RoyStegeman?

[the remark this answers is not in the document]
RoyStegeman: > The question is whether we need to distinguish (in both) `"FFNS"` from `"FONLL-FFNS"`, and why.

We do, unless we can drop FFNS in favor of what is not FONLL-FFNS. The difference is that regular FFNS includes all massive heavy contributions while FONLL-FFNS only has nf light + 1 heavy flavour. This way we prevent double counting e.g. the top contribution when we sum the different flavour number contributions.

src/pineko/fonll.py
```diff
@@ -0,0 +1,274 @@
...
+        """Initialize fonll info."""
+        self.ffns3 = ffns3
+        self.ffn03 = ffn03
+        self.ffns4 = ffns4
+        self.ffns4til = ffns4til
+        self.ffns4bar = ffns4bar
+        self.ffn04 = ffn04
+        self.ffns5 = ffns5
+        self.ffns5til = ffns5til
+        self.ffns5bar = ffns5bar
```
AleCandido, line 31: Even in this case, a list might be a suitable option.

Or even better: a list of tuples, i.e. `[(m, til, bar, m0), ...]` (at least it's easier to process automatically)

src/pineko/fonll.py
```diff
@@ -0,0 +1,274 @@
...
+            "ffn03": self.ffn03,
+            "ffns4": self.ffns4,
+            "ffns4til": self.ffns4til,
+            "ffns4bar": self.ffns4bar,
+            "ffn04": self.ffn04,
+            "ffns5": self.ffns5,
+            "ffns5til": self.ffns5til,
+            "ffns5bar": self.ffns5bar,
+        }
+        return {p: Path(paths[p]) for p in paths if paths[p] is not None}
```
AleCandido, line 47: In the case of a list of tuples above, you do not need the whole spell out, but just to index with two indices, e.g. `self.elems[nf, M0]` (with `M0 = 3`)

src/pineko/fonll.py
```diff
@@ -0,0 +1,274 @@
...
+
+    @property
+    def fks(self):
+        """Returns the pineappl.Grid objects reading the paths provided by self.fk_paths."""
+        return {fk: pineappl.grid.Grid.read(self.fk_paths[fk]) for fk in self.fk_paths}
```
AleCandido, line 52: Even this one could be a simpler function (it is effectively a precomputed function, the argument being the index of the dictionary).

In any case, always avoid indexing the thing you are iterating, since you have better iterators for this
```suggestion
        return {fk: pineappl.grid.Grid.read(path) for fk, path in self.fk_paths.items()}
```

src/pineko/fonll.py
```diff
@@ -0,0 +1,274 @@
...
+
+    @property
+    def dataset_name(self):
+        """Return the name of the dataset given by the fktables name, if all the fktables have the same name."""
+        if len({self.fk_paths[p].name for p in self.fk_paths}) == 1:
+            return self.fk_paths[list(self.fk_paths)[0]].name
+        else:
```
AleCandido, line 59: No need for `else`, since you're returning in the `if` (this is a warning for Pylint)

src/pineko/fonll.py
```diff
@@ -0,0 +1,274 @@
...
+    @property
+    def fks(self):
+        """Returns the pineappl.Grid objects reading the paths provided by self.fk_paths."""
+        return {fk: pineappl.grid.Grid.read(self.fk_paths[fk]) for fk in self.fk_paths}
+
+    @property
+    def dataset_name(self):
+        """Return the name of the dataset given by the fktables name, if all the fktables have the same name."""
+        if len({self.fk_paths[p].name for p in self.fk_paths}) == 1:
+            return self.fk_paths[list(self.fk_paths)[0]].name
```
AleCandido, line 58:
```suggestion
        names = {self.fk_paths[p].name for p in self.fk_paths}
        if len(names) == 1:
            return names[0]
```

Not only more efficient, but more readable.

src/pineko/fonll.py
```diff
@@ -0,0 +1,274 @@
...
+            del card["FNS"]
+            del card["PTO"]
+            del card["NfFF"]
+            del card["ID"]
+            del card["FONLLParts"]
+            del card["Comments"]
+        if len(theorycards) > 1 and not all(
+            [theorycards[0] == card in theorycards[1:]]
+        ):
+            raise ValueError("theorycards not the same")
```
AleCandido, line 77:
```suggestion
            raise ValueError("Theory cards are not compatible")
```

Since they do not have to be the same.

src/pineko/cli/fonll.py
andreab1997: [e116783](https://github.com/NNPDF/pineko/pull/99/commits/e116783b5df5242da03554e8cfa859d89f551a52)

src/pineko/cli/fonll.py
andreab1997: [e116783](https://github.com/NNPDF/pineko/pull/99/commits/e116783b5df5242da03554e8cfa859d89f551a52)

src/pineko/cli/fonll.py
andreab1997: Maybe yes but they are all options because they are all optional. I believe that following your suggestion we may lose that flexibility or at least in clarity.

src/pineko/cli/fonll.py
andreab1997: Yes but, according to what is `tcard["FNS"]`, the error raised could be rather difficult to understand. For this reason I believe it is better to have an explicit error at least for this fundamental configuration, i.e. that the `FNS` is of `FONLL` kind.

src/pineko/fonll.py
```diff
@@ -0,0 +1,274 @@
...
+    def Q2grid(self):
+        """Return the Q2grid of the fktables given by self.fks ."""
+        aa = json.loads(self.fks[list(self.fks)[0]].key_values()["runcard"])[
+            "observables"
+        ]
+        bb = list(aa.values())[
+            0
+        ]  # there is only a single obseravble because it's a dis fktable
+        cc = np.array([i["Q2"] for i in bb])
+        return cc
```
AleCandido, line 90:
```suggestion
        return np.array([i["Q2"] for i in bb])
```

To save one more random name

src/pineko/fonll.py
```diff
@@ -0,0 +1,274 @@
...
+
+    @property
+    def Q2grid(self):
+        """Return the Q2grid of the fktables given by self.fks ."""
+        aa = json.loads(self.fks[list(self.fks)[0]].key_values()["runcard"])[
+            "observables"
+        ]
+        bb = list(aa.values())[
+            0
+        ]  # there is only a single obseravble because it's a dis fktable
```
AleCandido, line 88:
```suggestion
        obs = json.loads(self.fks[list(self.fks)[0]].key_values()["runcard"])[
            "observables"
        ]
        bb = list(obs.values())[
            0
        ]  # there is only a single obseravble because it's a dis fktable
```

src/pineko/fonll.py
```diff
@@ -0,0 +1,274 @@
...
+        cc = np.array([i["Q2"] for i in bb])
```
AleCandido, line 89:
```suggestion
        kins = list(aa.values())[
            0
        ]  # there is only a single obseravble because it's a dis fktable
        cc = np.array([i["Q2"] for i in kins])
```

src/pineko/fonll.py
```diff
@@ -252,6 +252,12 @@ def produce_fonll_recipe(fonll_fns, damp):
                 "FONLLParts": part,
             }
         )
+        # In a mixed FONLL scheme we only subract the resumt terms that are
+        # present in the FFNS scheme at nf+1. E.g. for FONLL-B in FFN03 we
+        # only subract up to NLL since there is no NNLL in FFNS4
+        if is_mixed and fns == "FONLL-FFN0":
+            fonll_recipe[-1]["PTODIS"] = po
```
andreab1997, line 259: Do I understand correctly that, in this way, `yadism` wil only use `PTODIS` while `eko` will only use `PTO`?

RoyStegeman: No yadism uses both - PTODIS for the coefficient and PTO for the resummed logarithms.

andreab1997: Ah ok thanks

src/pineko/fonll.py
andreab1997: [c8b8f03](https://github.com/NNPDF/pineko/pull/99/commits/c8b8f039400e6f666f87f38c6da8bd5d7cafc7a3)

src/pineko/cli/fonll.py
andreab1997: I believe we can just remove this

andreab1997: Thanks for spotting it

andreab1997: [11d8f84](https://github.com/NNPDF/pineko/pull/99/commits/11d8f84ce6378824967c16879180bc08e1f173d1)

src/pineko/cli/fonll.py
AleCandido: Actually, is this really the case? I believe you can choose to not have a flavor, or some part of it, but not all the options are independent.

We may have a scheme to always interpret the list.

RoyStegeman (commented):
Some minor comments. I don't have much of a strong opinion regarding @AleCandido's point for inputs. Yes, it's ugly, but in practice you only type it once since you call the functions from a bash script, and it does make that script very readable.

giacomomagni (commented):
Since points which vanish in the DIS grid are not included in the EKO `mugrid`, 
some EKO tables contains different points or are even empty. 
This cause some FKtables to fail. It might be good to redirect such errors to a nicer output as: 

```
Massive part is 0, skipping the FKtable.
```

src/pineko/cli/fonll.py
andreab1997: From the code point of view you have full flexibility. Of course from the physics point of view there are some choices that just do not make sense. The code would not fail though

andreab1997:
> Some minor comments. I don't have much of a strong opinion regarding @AleCandido's point for inputs. Yes, it's ugly, but in practice you only type it once since you call the functions from a bash script, and it does make that script very readable.

Indeed this was one of the original reasons to implement it like it is.

src/pineko/evolve.py
```diff
@@ -128,9 +129,14 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
         matching_scales=heavy_quarks.MatchingScales(masses * thresholds_ratios),
         origin=(tcard["Q0"] ** 2, tcard["nf0"]),
     )
-    operators_card["mugrid"] = [
-        (float(np.sqrt(q2)), int(nf_default(q2, atlas))) for q2 in q2_grid
-    ]
+    # If we are producing num_fonll FKs we need to look to NfFF...
+    if check.is_num_fonll(tcard["FNS"]):
```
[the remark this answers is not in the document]
andreab1997: True, but with `pineko`, before nFONLL happened, we never produced FK tables with FFNS. As you see in the code, if we are not computing `num_fonll` we just use `nf_default` so basically we do VFNS as we used to do before with  aFONLL

benchmarks/bench_evolve.py
```diff
@@ -13,6 +13,41 @@
 import pineko.theory_card
 
 
+@pytest.mark.parametrize("theoryid,is_mixed", [(400, False), (208, True)])
```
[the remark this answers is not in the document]
andreab1997: Good point, I did not realize. I usually compute 7 but for production would be good to be aware of this @giacomomagni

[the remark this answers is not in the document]
RoyStegeman: I'm not sure how you generate eko runcards from the DIS grids, or how the convolution of the two is done under the hood. The one thing to consider is that (as Giacomo pointed out) points that vanish in the DIS grid don't are not included in the eko. So to have a general eko that can be used for all fktables at a given nf you'd probably want to include the `mugrid` based on the dataset rather than the DIS grid - that, or you'd somehow need to merge the DIS grids to get the union of all non-zero points. 

Either way - if this sounds like something that could be worth it maybe open an issue and keep it for a separate PR.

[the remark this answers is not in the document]
andreab1997: Yes I agree, for the moment we can generate one eko for each subtheory and that's it. Later on we can surely improve both this and the inputs issue @AleCandido pointed out

andreab1997:
So, I believe the only discussions that are still open are about things that we agreed to solve in a later PR. Then, if I am not wrong, would you all agree to open a couple of issues for these discussions and merge this now?

@giacomomagni note that what you asked (or something similar) has been implemented in [cc66bb4](https://github.com/NNPDF/pineko/pull/99/commits/cc66bb4dc56bd0c3f6763856d6c26cbe84b68cce).

giacomomagni:
> @giacomomagni note that what you asked (or something similar) has been implemented in [cc66bb4](https://github.com/NNPDF/pineko/pull/99/commits/cc66bb4dc56bd0c3f6763856d6c26cbe84b68cce).

Thanks for implementing this log. 
Would it be possible to redirect the `fks` error to a message as well:

```
   File "/project/theorie/gmagni/miniconda3/envs/nnpdf/bin/pineko", line 5, in <module>
    command()
  File "/project/theorie/gmagni/miniconda3/envs/nnpdf/lib/python3.9/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/project/theorie/gmagni/miniconda3/envs/nnpdf/lib/python3.9/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/project/theorie/gmagni/miniconda3/envs/nnpdf/lib/python3.9/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/project/theorie/gmagni/miniconda3/envs/nnpdf/lib/python3.9/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/project/theorie/gmagni/miniconda3/envs/nnpdf/lib/python3.9/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/project/theorie/gmagni/miniconda3/envs/nnpdf/lib/python3.9/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/data/theorie/gmagni/N3PDF/pineko/src/pineko/cli/theory_.py", line 93, in fks
    theory.TheoryBuilder(
  File "/data/theorie/gmagni/N3PDF/pineko/src/pineko/theory.py", line 466, in fks
    self.iterate(self.fk, tcard=tcard, pdf=pdf)
  File "/data/theorie/gmagni/N3PDF/pineko/src/pineko/theory.py", line 212, in iterate
    f(name, grid, **kwargs)
  File "/data/theorie/gmagni/N3PDF/pineko/src/pineko/theory.py", line 432, in fk
    _grid, _fk, comparison = evolve.evolve_grid(
  File "/data/theorie/gmagni/N3PDF/pineko/src/pineko/evolve.py", line 212, in evolve_grid
    operators, targetgrid=eko.interpolation.XGrid(x_grid)
  File "/data/theorie/gmagni/N3PDF/eko_fhmv_private/src/eko/interpolation.py", line 430, in __init__
    raise ValueError(f"xgrid needs at least 2 points, received {len(xgrid)}")
ValueError: xgrid needs at least 2 points, received 0
```

andreab1997:
> > @giacomomagni note that what you asked (or something similar) has been implemented in [cc66bb4](https://github.com/NNPDF/pineko/pull/99/commits/cc66bb4dc56bd0c3f6763856d6c26cbe84b68cce).
> 
> Thanks for implementing this log. Would it be possible to redirect the `fks` error to a message as well:
> 
> ```
>    File "/project/theorie/gmagni/miniconda3/envs/nnpdf/bin/pineko", line 5, in <module>
>     command()
>   File "/project/theorie/gmagni/miniconda3/envs/nnpdf/lib/python3.9/site-packages/click/core.py", line 1130, in __call__
>     return self.main(*args, **kwargs)
>   File "/project/theorie/gmagni/miniconda3/envs/nnpdf/lib/python3.9/site-packages/click/core.py", line 1055, in main
>     rv = self.invoke(ctx)
>   File "/project/theorie/gmagni/miniconda3/envs/nnpdf/lib/python3.9/site-packages/click/core.py", line 1657, in invoke
>     return _process_result(sub_ctx.command.invoke(sub_ctx))
>   File "/project/theorie/gmagni/miniconda3/envs/nnpdf/lib/python3.9/site-packages/click/core.py", line 1657, in invoke
>     return _process_result(sub_ctx.command.invoke(sub_ctx))
>   File "/project/theorie/gmagni/miniconda3/envs/nnpdf/lib/python3.9/site-packages/click/core.py", line 1404, in invoke
>     return ctx.invoke(self.callback, **ctx.params)
>   File "/project/theorie/gmagni/miniconda3/envs/nnpdf/lib/python3.9/site-packages/click/core.py", line 760, in invoke
>     return __callback(*args, **kwargs)
>   File "/data/theorie/gmagni/N3PDF/pineko/src/pineko/cli/theory_.py", line 93, in fks
>     theory.TheoryBuilder(
>   File "/data/theorie/gmagni/N3PDF/pineko/src/pineko/theory.py", line 466, in fks
>     self.iterate(self.fk, tcard=tcard, pdf=pdf)
>   File "/data/theorie/gmagni/N3PDF/pineko/src/pineko/theory.py", line 212, in iterate
>     f(name, grid, **kwargs)
>   File "/data/theorie/gmagni/N3PDF/pineko/src/pineko/theory.py", line 432, in fk
>     _grid, _fk, comparison = evolve.evolve_grid(
>   File "/data/theorie/gmagni/N3PDF/pineko/src/pineko/evolve.py", line 212, in evolve_grid
>     operators, targetgrid=eko.interpolation.XGrid(x_grid)
>   File "/data/theorie/gmagni/N3PDF/eko_fhmv_private/src/eko/interpolation.py", line 430, in __init__
>     raise ValueError(f"xgrid needs at least 2 points, received {len(xgrid)}")
> ValueError: xgrid needs at least 2 points, received 0
> ```

Yes, I guess I can do this

benchmarks/bench_evolve.py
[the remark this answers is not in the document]
giacomomagni: Yes you are right, the ekos do not need to be splatted... 
But I prefer having something automatised as possible, so a solution would be that if you have 7 theories, *02 and *03 (*05, *06) share the same ekos.

RoyStegeman:
> would you all agree to open a couple of issues for these discussions and merge this now?

I can't approve since I opened this PR myself, but once you've implemented the warning @giacomomagni suggested and resolved the conflicts I'm happy for it to be merged.

giacomomagni:
> Actually, this error, when computing fktables for non nFONLL theories, is an actual error. Also, for nFONLL theories, it is not wrong: it is an error, we just know where it comes from and so we ignore it. So, I would not redirect this to a message also in non nFONLL cases to be honest.

I disagree, if the grid is all zeros we can simply skip the computation of the fks table and exit, there is no need to raise this error. It should be an easy fix. As it is now:

1. It's not clear if there is a problem or this originates from the empty grid because you have a pure massive theory
2. When you iterate over datasets with `fks` the program will crash and not continue to the next dataset, which is kind of inconvenient.

andreab1997:
> > Actually, this error, when computing fktables for non nFONLL theories, is an actual error. Also, for nFONLL theories, it is not wrong: it is an error, we just know where it comes from and so we ignore it. So, I would not redirect this to a message also in non nFONLL cases to be honest.
> 
> I disagree, if the grid is all zeros we can simply skip the computation of the fk table and exit, there is no need to raise this error. It should be an easy fix. As it is now:
> 
> 1. It's not clear if there is a problem or this originates from the empty grid because you have a pure massive theory
> 2. When you iterate over datasets with `fks` the program will crash and not continue to the next dataset, which is kind of inconvenient.

I would completely agree if we only computed nFONLL fktables but this is not the case. As it is now, redirecting that error would affect all the fktables and while I agree it is inconvenient for nFONLL FKs, it is instead better to have the error for non nFONLL FKs

giacomomagni:
> I would completely agree if we only computed nFONLL fktables but this is not the case. As it is now, redirecting that error would affect all the fktables and while I agree it is inconvenient for nFONLL FKs, it is instead better to have the error for non nFONLL FKs

Sorry to be insistent: but it's possible to detect if you are doing nFONLL, there is `is_num_fonll` available.

felixhekhorn:
but this `ValueError: xgrid needs at least 2 points, received 0` clearly sounds like a true error - either we should never reach that point or there is something wrong ... don't silence it, solve it

giacomomagni:
> but this `ValueError: xgrid needs at least 2 points, received 0` clearly sounds like a true error - either we should never reach that point or there is something wrong ... don't silence it, solve it

Indeed that is exactly my point, we should not arrive there, otherwise it's really confusing!

andreab1997:
> but this `ValueError: xgrid needs at least 2 points, received 0` clearly sounds like a true error - either we should never reach that point or there is something wrong ... don't silence it, solve it

The error is very confusing and I agree but the reason is very simple: the correspoding eko is empty

RoyStegeman:
Just to give my two cents - I agree with @giacomomagni. The error is confusing and should never be reached, the big practical risk is that if there are any real/different errors they will be buried in these errors we can ignore.

andreab1997:
Ok given that I am the only one thinking that redirecting that error is not needed, I will do it (just not today most likely)

AleCandido:
Just to stress one of the suggested points: isn't it possible to avoid raising this error?

You could check before if the EKO is empty. And in case it is, just skip without generating the error.

Usually I'm huge fan of raising and catching exceptions in Python, because it allows you to avoid a whole lot of checks, frequently more than the `try... except...` blocks you have to write.
However, in this case you would have to check anyhow, even after catching. So let's do it before, if possible.

andreab1997:
@giacomomagni (and all) please let me know if the check implemented in [eb73599](https://github.com/NNPDF/pineko/pull/99/commits/eb735991760bf5ab32221ae8aa4174180f110eba) is what you had in mind.

src/pineko/theory.py
```diff
@@ -397,6 +408,9 @@ def fk(self, name, grid_path, tcard, pdf):
                 check_scvar_evolve(grid, max_as, max_al, check.Scale.FACT)
         # loading ekos to produce a tmp copy
         with eko.EKO.read(eko_filename) as operators:
+            # Skip the computation of the fktable if the eko is empty
+            if len(operators.mu2grid) == 0:
```
giacomomagni, line 412:
```suggestion
            if len(operators.mu2grid) == 0 and check.is_num_fonll(tcard['FNS']) :
```

giacomomagni: Sure but if is nFONLL I don't want to crash, just continue. 
In the other cases you should crash

src/pineko/evolve.py
```diff
@@ -105,6 +105,7 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
 
     """
     # Add a +1 to the orders for the difference in convention between nnpdf and pineappl
+    # NB: This would not happen for nFONLL
     is_fns = int(check.is_fonll_b(tcard["FNS"], pineappl_grid.lumi()))
```
felixhekhorn, line 109: is the `tcard["PTO"]` for nFONLL already corrected? i.e. are we correcting here too much? (which can be an issue for N3LO) - so do you need to have a `is_num_fonll` in the condition?

src/pineko/theory.py
andreab1997: Wait, continue with what? If it cannot compute the fktable, that is it, there are no other things to do when you ask for fktables

src/pineko/evolve.py
andreab1997: Indeed they are already corrected and in fact this is the meaning of the comment. The check `check.is_fonll_b` would return `False` for a nFONLL fk

src/pineko/theory.py
giacomomagni: With the other datasets.... I don't want `fks` to crash

andreab1997: But the other datasets would not have empty ekos so it would not crash. Am I missing something?

giacomomagni: the command will stop and crash, no?

andreab1997: Ok so you mean for the other fk_tables of the same datasets. Ok but then first for DIS only BCDMS and NMCPD have more than one fktable in the yaml file. Second, the fact that an eko is empty is physics related, for instance because for that observable you do not have massive contributions. But then if this is true for one of the two Fktable of NMCPD, for example, it would be true also for the other

felixhekhorn: > But then if this is true for one of the two Fktable of NMCPD, for example, it would be true also for the other

but this might or might not be true for BCDMS, because there you iterate energies and so kinematics

however, what does pineko do when an FK table is empty by physics? can we deal with this situation or do we need an empty grid?

andreab1997: It is basically impossible to create empty fktables because if the grid is empty, then pineappl will not give the mugrid and xgrid, but it will pass empty lists. This means that then the ekos will be empty and the fktables computation would just fail.

AleCandido: Ok, but then at this point I'm back with the `try ... except` pattern: @andreab1997 is raising a very specific error, that could be caught where needed and avoid failure. Isn't it?

giacomomagni: This is really confusing... before we had a command `fks` that looped over all the datasets and create the fk-tables. 
I'd like to have this working again, such that with one single command you loop on all the datasets of a theory, 
without having to call each dataset separated. If this is not possible anymore fine.

andreab1997: But this was never the case... The command `fks` iterates over all the fktables defined in a yamldb of a *single* dataset

andreab1997: Ok I see your point, I am used to call pineko once for each dataset so I did not think about this. Maybe we can just do as @AleCandido is proposing and this should be fine for everyone

src/pineko/evolve.py
giacomomagni: But shouldn't that check include `FONLL-D` as well ?

andreab1997: In principle you are right and I can even add if you like, but we never use the `mixed FONLL` prescriptions beyond NLO

andreab1997: [1b32ca0](https://github.com/NNPDF/pineko/pull/99/commits/1b32ca0ac3478b2a31e874bbfc5d6b967051162b)

src/pineko/theory.py
giacomomagni: [025a578](https://github.com/NNPDF/pineko/pull/99/commits/025a5787a1dff76ccf13e78bd936abc4228bc486) is the easiest solution I can think about and accommodate the what happens for nFONLL

src/pineko/theory.py
```diff
@@ -409,9 +417,12 @@ def fk(self, name, grid_path, tcard, pdf):
...
         with eko.EKO.read(eko_filename) as operators:
             # Skip the computation of the fktable if the eko is empty
-            if len(operators.mu2grid) == 0:
-                raise EmptyEkoError
-            eko_tmp_path = operators.paths.root.parent / "eko-tmp.tar"
+            if len(operators.mu2grid) == 0 and check.is_num_fonll(tcard["FNS"]):
+                rich.print(f"[green] Skipping empty eko for nFONLL.")
+                return
+            eko_tmp_path = (
+                operators.paths.root.parent / f"eko-tmp-{np.random.rand()}.tar"
```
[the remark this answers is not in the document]
felixhekhorn: and if you want to be more diverse (for parallel, I guess), maybe put the grid name or dataset name somewhere

src/pineko/fonll.py
```diff
@@ -0,0 +1,290 @@
...
+            "ffns5til": self.ffns5til,
+            "ffns5bar": self.ffns5bar,
+        }
+        actually_existing_paths = [p for p in paths if paths[p] is not None]
+        for p in paths:
+            if p not in actually_existing_paths:
+                logger.warning(
+                    f"Warning! FK table for {p} does not exist and thus is being skipped."
+                )
+        return {p: Path(paths[p]) for p in actually_existing_paths}
```
RoyStegeman, line 56: I think this warning is a bit much since it's printed each time `fk_paths` is used. 
```suggestion
        return {p: Path(paths[p]) for p in paths if paths[p] is not None}
```

src/pineko/theory.py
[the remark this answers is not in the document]
andreab1997: No it is not for parallel. This is to avoid rotating (in x space) the original eko because, as @giacomomagni noticed, once rotated the eko is much heavier. So we copy the eko here and then we remove it once the fktable is done. For this reason I don't believe the eko name matters so much.

src/pineko/fonll.py
andreab1997: Is [790c0a5](https://github.com/NNPDF/pineko/pull/99/commits/790c0a5f6cb5d8b41eb3e9158662c3a865605c3f) better?

src/pineko/theory.py
andreab1997: For me this is even ok but we are giving up having the same kind of check for not nFONLL fktables

src/pineko/theory.py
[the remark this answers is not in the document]
andreab1997: Just to be precise, this is not for parallel because the FKtable computation is so fast that I don't believe having the same name at the same moment is something possible. Anyway putting the grid name could be a good idea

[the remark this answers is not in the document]
giacomomagni: This was indeed for the rotation memory problem, but the name fix is for parallel. 
In fact if you are computing many `FKtables` at the same time, it's safer/correct to have different temp file names. 
For the actual name any name it's fine for me.

src/pineko/fonll.py
giacomomagni: yes, thanks!

src/pineko/theory.py
giacomomagni: Yes, because in that case (not nFONLL) I don't see why you should end up with a nice empty grid, so you have better to crash!

src/pineko/theory.py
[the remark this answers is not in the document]
andreab1997: So, do you like how it is now?

andreab1997:
is everyone happy with this?

src/pineko/fonll.py
```diff
@@ -23,37 +23,28 @@ def __init__(
...
             if p not in actually_existing_paths:
                 logger.warning(
                     f"Warning! FK table for {p} does not exist and thus is being skipped."
                 )
-        return {p: Path(paths[p]) for p in actually_existing_paths}
+
+    @property
+    def fk_paths(self):
+        """Returns the list of the fk paths needed to produce FONLL predictions."""
+        return {p: Path(self.paths[p]) for p in self.paths if self.paths[p] is not None}
```
felixhekhorn, line 47: These 3 loops could be collapsed to 1 (a minor thing though ...)

src/pineko/fonll.py
```diff
@@ -0,0 +1,281 @@
...
+        # present in the FFNS scheme at nf+1. E.g. for FONLL-B in FFN03 we
+        # only subract up to NLL since there is no NNLL in FFNS4
+        if fonll_fns in MIXED_ORDER_FNS and fns == "FONLL-FFN0":
+            fonll_recipe[-1]["PTODIS"] = po
+            fonll_recipe[-1]["PTO"] = po - 1
+    return fonll_recipe
+
+
+def produce_fonll_tcards(tcard, tcard_parent_path, theoryid):
+    """Produce the seven fonll tcards from an original tcard and dump them in tcard_parent_path with names from '{theoryid}00.yaml' to '{theoryid}06.yaml'."""
```
felixhekhorn, line 266:
```suggestion
    """Produce the seven fonll tcards from an original tcard.
    
     We dump them in tcard_parent_path with names from '{theoryid}00.yaml' to '{theoryid}06.yaml'."""
```
There is nothing wrong with using more then one-line docstrings ... the second part is clearly a detail

src/pineko/fonll.py
```diff
@@ -0,0 +1,281 @@
...
+    fonll_recipe = produce_fonll_recipe(tcard["FNS"], tcard["DAMP"])
+    n_theory = len(fonll_recipe)
+    theorycards = [copy.deepcopy(tcard) for _ in range(n_theory)]
+    paths_list = []
+    for num, (theorycard, recipe) in enumerate(zip(theorycards, fonll_recipe)):
+        # update cards entries
+        theorycard.update(recipe)
+        theorycard["ID"] = int(f"{theoryid}0{num}")
```
felixhekhorn, line 274:
```suggestion
        # TODO: find a better naming scheme
        theorycard["ID"] = int(f"{theoryid}0{num}")
```
the todo is not for this PR, but something to keep in mind for the future

felixhekhorn (changes requested):
Sorry for the long review, but since this was clearly a moving target I never had a detailed look so far ... however, most things are not super complicated

src/pineko/check.py
```diff
@@ -116,31 +116,56 @@ def check_grid_and_eko_compatible(pineappl_grid, operators, xif, max_as, max_al)
...
+           luminosity info
+
+    Returns
+    -------
+    bool
+        true if the fktable is a mixed FONLL DIS fktable
+    """
+    if not is_dis(lumi):
+        return False
+    mixed_FONLL_fns = ["FONLL-B", "FONLL-D", "FONLL-F"]
```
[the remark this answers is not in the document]
felixhekhorn: see also [this](https://github.com/NNPDF/pineko/pull/99#discussion_r1363662237) and [this](https://github.com/NNPDF/pineko/pull/99#discussion_r1363673933) comment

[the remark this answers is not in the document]
felixhekhorn: good - I resolve the relevant comments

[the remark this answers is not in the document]
giacomomagni: I thought we implemented: `D=(3,2), E=(3,3)`.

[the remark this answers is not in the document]
felixhekhorn: Can you please still add some documentation somewhere with the actual mapping?

src/pineko/fonll.py
```diff
@@ -0,0 +1,281 @@
...
+]
+NFFF_LIST = [3, 3, 4, 4, 5]
+PARTS_LIST = ["full" for _ in range(5)]
+
+
+def produce_ptos(fns, is_mixed_or_damp):
+    """Produce the list of PTOs needed for the requested fns."""
+    base_pto = FNS_BASE_PTO[fns]
+    # mixed FONLL
+    if fns in MIXED_ORDER_FNS:
```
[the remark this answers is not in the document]
giacomomagni: This damping stuff is indeed quite confusing. 
The real fix would be to have always the computation of 3 Ekos (no matter of damping). 
And then ekos are inherited to the correct number of theories.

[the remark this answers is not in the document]
felixhekhorn: I agree - let's consider the reuse of ekos in a separate PR #130 else we never finish this one :see_no_evil: the damping however needs to be clarified!

src/pineko/cli/fonll.py
```diff
@@ -0,0 +1,174 @@
...
+def fonll_tcards(theoryid, cfg):
+    """Produce the FONLL tcards starting from the original tcard given by the theoryID."""
+    path = configs.detect(cfg)
+    base_configs = configs.load(path)
+    configs.configs = configs.defaults(base_configs)
+    tcard = theory_card.load(theoryid)
+    tcard_parent_path = theory_card.path(theoryid).parent
+    if "FONLL" not in tcard["FNS"]:
+        raise TheoryCardError("The theorycard does not correspond to an FONLL scheme.")
+    _ = fonll.produce_fonll_tcards(tcard, tcard_parent_path, theoryid)
```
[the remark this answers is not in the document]
andreab1997: It is not but still `produce_fonll_tcards` returns something so I would say putting `_` instead of nothing is clearer.

src/pineko/fonll.py
```diff
@@ -0,0 +1,281 @@
...
+        mc = theorycard_constituent_fks["mc"]
+        mb = theorycard_constituent_fks["mb"]
+        q2grid = fonll_info.Q2grid
+        step_function_charm = mc**2 < q2grid
+        step_function_bottom = mb**2 < q2grid
+        damping_factor_charm = (1 - mc / q2grid) ** damp[1]
+        damping_factor_bottom = (1 - mb / q2grid) ** damp[1]
+        damping_factor_charm *= step_function_charm
+        damping_factor_bottom *= step_function_bottom
+        dampings = {"mc": damping_factor_charm, "mb": damping_factor_bottom}
```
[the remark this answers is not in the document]
andreab1997: [ece695b](https://github.com/NNPDF/pineko/pull/99/commits/ece695bf0bbc5259c863216420448d77f27906a1)

src/pineko/fonll.py
```diff
@@ -0,0 +1,281 @@
...
+        Path(configs.load(configs.detect(cfg))["paths"]["theory_cards"])
+        / f"{theoryid}.yaml"
+    )
+    with open(input_theorycard_path) as f:
+        final_theorycard = yaml.safe_load(f)
+    theorycard = json.loads(combined_fk.key_values()["theory"])
+    theorycard["FNS"] = final_theorycard["FNS"]
+    theorycard["PTO"] = final_theorycard["PTO"]
+    theorycard["NfFF"] = final_theorycard["NfFF"]
+    theorycard["ID"] = final_theorycard["ID"]
```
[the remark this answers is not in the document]
andreab1997: [e406de6](https://github.com/NNPDF/pineko/pull/99/commits/e406de69cc7f3ea813520c346498e0cd05fe786c)

src/pineko/fonll.py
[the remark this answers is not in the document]
andreab1997: [5efe537](https://github.com/NNPDF/pineko/pull/99/commits/5efe53768a865aa7b1569156a4d9e3bea8412033)

src/pineko/fonll.py
```diff
@@ -0,0 +1,281 @@
...
+        if len(theorycards) > 1 and not all(
+            [theorycards[0] == card for card in theorycards[1:]]
+        ):
+            raise ValueError("Theory cards are not compatible")
+        return theorycards[0]
+
+    @property
+    def Q2grid(self):
+        """Return the Q2grid of the fktables given by self.fks ."""
+        obs = json.loads(self.fks[list(self.fks)[0]].key_values()["runcard"])[
```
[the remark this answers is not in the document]
andreab1997: Yes true but maybe @RoyStegeman had a reason to do so. Let's ask him.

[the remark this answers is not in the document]
RoyStegeman: Nope no reason, probably just because I don't know the structure of the `Grid` class in `pineappl`

src/pineko/fonll.py
```diff
@@ -0,0 +1,281 @@
...
+        dampings = {"mc": damping_factor_charm, "mb": damping_factor_bottom}
+        with tempfile.TemporaryDirectory() as tmpdirname:
+            combined_fk = fk_dict[list(fk_dict)[0]]
+            for fk in list(fk_dict)[1:]:
+                tmpfile_path = Path(tmpdirname) / f"{fk}.pineappl.lz4"
+                sign = -1 if fk in FK_WITH_MINUS else 1
+                fk_dict[fk].scale(sign)
+                for mass in FK_TO_DAMP:
+                    if fk in FK_TO_DAMP[mass]:
+                        fk_dict[fk].scale_by_bin(dampings[mass])
```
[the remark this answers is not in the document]
andreab1997: [34fd85d](https://github.com/NNPDF/pineko/pull/99/commits/34fd85d001059a9c3db7a25b7c095b5c5e40eae3)

src/pineko/fonll.py
andreab1997: [34fd85d](https://github.com/NNPDF/pineko/pull/99/commits/34fd85d001059a9c3db7a25b7c095b5c5e40eae3)

src/pineko/fonll.py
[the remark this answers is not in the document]
andreab1997: Ok then let me change that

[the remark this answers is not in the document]
andreab1997: [babe6ae](https://github.com/NNPDF/pineko/pull/99/commits/babe6ae5c91a5ddbb5c4534a1afa9097fb77c3de)

src/pineko/fonll.py
```diff
@@ -0,0 +1,281 @@
...
+MIXED_NFFF_LIST = [3, 3, 4, 4, 4, 5, 5]
+MIXED_PARTS_LIST = [
+    "full",
+    "full",
+    "massless",
+    "massive",
+    "full",
+    "massless",
+    "massive",
+]
```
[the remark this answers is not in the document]
andreab1997: [5a3252e](https://github.com/NNPDF/pineko/pull/99/commits/5a3252e9df3d2a133c5adcffeb8fc57e4d2c0d74)

benchmarks/bench_evolve.py
```diff
@@ -46,7 +46,7 @@ def benchmark_write_operator_card_from_file_num_fonll(
         with open(opcard_path, encoding="utf-8") as f:
             ocard = yaml.safe_load(f)
         for entry in ocard["mugrid"]:
-            assert entry[1] == nfff
+            assert entry[1] == int(nfff)
```
RoyStegeman, line 49: I think this should be saved/read as `int` so no need to force the int here

andreab1997: I believe it is needed because here `nfff` comes from the `numpy` array of `MIXED_FNS_CONFIG` and thus is a string. I tried without and the benchmark was not passing

RoyStegeman: Oh wait you get it from the hardcoded list, but then it should already be coded as int right? Meaning this does nothing

RoyStegeman: I think the benchmark wasn't passing because in the tcard it was saved as a string, but you were comparing it to MIXED_FNS_CONFIG where it is an int

RoyStegeman: See [bc78002](https://github.com/NNPDF/pineko/pull/99/commits/bc7800238693498afb41963025a5550e1c21485e)

andreab1997: Yes but look at [b284bb1](https://github.com/NNPDF/pineko/pull/99/commits/b284bb1f03b8751dd18803b2b2ca7eec61b9bb5d) where another benchmark is failing for the same reason. You see that the string is actually the `'3'` coming from MIXED_FNS_CONFIG

andreab1997: It is not coming from the theory card

RoyStegeman: Does it help if you provide the `dtypes` for each of the components in the tuple?

andreab1997: I don't believe so. I believe that the problem is linked to the `np.array` cast

RoyStegeman: Yes and I believe that can be changed for the individual components of the tuples by doing `MIXED_FNS_CONFIG = np.array(dtype="U10,int,U10"`

andreab1997: If I do like you suggested then the array shape becomes `(7,)` instead of `(7,3)` as it should and the transposition does not work anymore

src/pineko/cli/fonll.py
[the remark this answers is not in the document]
felixhekhorn: I'd say if you want to suppress that return put an explicit `return` here instead

src/pineko/fonll.py
felixhekhorn: It might clash with an already existing theory 70000 (if I got the number correct, i.e. 700 + 00)

src/pineko/fonll.py
```diff
@@ -239,18 +233,30 @@ def produce_fonll_recipe(fonll_fns, damp):
...
+        else np.array(FNS_CONFIG).transpose().tolist()[1]
+    )
+    parts_list = (
+        np.array(MIXED_FNS_CONFIG).transpose().tolist()[2]
+        if is_mixed_or_damp
+        else np.array(FNS_CONFIG).transpose().tolist()[2]
+    )
     for fns, nfff, po, part in zip(
         fns_list, nfff_list, produce_ptos(fonll_fns, is_mixed_or_damp), parts_list
     ):
```
felixhekhorn, line 253: Now you're transposing and zipping here (which are inverse to each other), but this can be simplified: add to your scheme definitions a new field `delta_pto` (which can be 1 or 0) and then you can do a simple iteration here and put `"PTO": base_pto + delta_pto`

We could/should maybe even put a dataclass `FNSSchemeConfig` (or similar) - and we can make factors (i.e. minuses and damping part of it)

felixhekhorn (changes requested):

src/pineko/fonll.py
felixhekhorn: actually, damping depends on the grid so that could be left apart - but minuses are universal and can be attributed to a config

andreab1997: I followed your suggestion (more or less). Let me know if you like it more now :) 
[7e385a3](https://github.com/NNPDF/pineko/pull/99/commits/7e385a36e70932544de4a75b8fd753f2169f7f3b)

src/pineko/cli/fonll.py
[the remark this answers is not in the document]
andreab1997: No but I don't want to suppress the return of `produce_fonll_tcards`. Maybe I am not getting your point.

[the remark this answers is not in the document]
felixhekhorn:
```suggestion
    fonll.produce_fonll_tcards(tcard, tcard_parent_path, theoryid)
    return
```

src/pineko/fonll.py
felixhekhorn: I'm sorry, but no :upside_down_face: (because you transposed again) - however, let me have a shot myself for it

andreab1997: No wait, I never transposed now

felixhekhorn: #131

src/pineko/fonll.py
```diff
@@ -17,13 +17,24 @@
 logger = logging.getLogger(__name__)
 
 FNS_BASE_PTO = {"FONLL-A": 1, "FONLL-B": 1, "FONLL-C": 2, "FONLL-D": 2, "FONLL-E": 3}
+"""Mapping between pertubative orders as in the theory card PTO and the FONLL scheme names.
+
+The explict mapping is the following (evolution, massive parts, massless parts):
+    * A: (1,1,1)
+    * B: (2,2,1)
```
felixhekhorn, line 24: this does not seem correct: B is an NLO scheme and hence evolution is at NLO

src/pineko/fonll.py
```diff
@@ -17,13 +17,24 @@
 logger = logging.getLogger(__name__)
 
 FNS_BASE_PTO = {"FONLL-A": 1, "FONLL-B": 1, "FONLL-C": 2, "FONLL-D": 2, "FONLL-E": 3}
+"""Mapping between pertubative orders as in the theory card PTO and the FONLL scheme names.
+
+The explict mapping is the following (evolution, massive parts, massless parts):
```
AleCandido, line 22: Why do you need to distinguish `evolution` from `massive parts`?  From what I understood the resummation (NLL in FONLL) is just labeled by `massless parts`, so the other two have to be the same...

benchmarks/bench_evolve.py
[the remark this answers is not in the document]
felixhekhorn: #130

src/pineko/fonll.py
AleCandido: It's not so much redundancy on its own, it is actually about the meaning. We have both evolution for massive and massless, and they have two different orders.

Why do you need to specify it separately? Which evolution are you referring to? (I guess the massive one, but I wonder why...)

felixhekhorn: > We have both evolution for massive and massless, and they have two different orders.

no - they have both the same order, which indeed coincides with the massless order

AleCandido: Sorry, I guess massless is asy (*massless* keeps being an ambiguous word, whether it is the *massless limit* or *zero-mass*).

But then they all should have the same order (as I argued back then), otherwise you see it's rather inconsistent - you're definitely taking a weird limit...
(you do not subtract the logs of the extra mass, because they are not in the zero-mass, but you subtract the evolution logs of all the other masses, that are also not in the zero-mass - that's quite arbitrary)

src/pineko/check.py
```diff
@@ -116,31 +116,56 @@ def check_grid_and_eko_compatible(pineappl_grid, operators, xif, max_as, max_al)
...
+    -------
+    bool
+        true if the fktable is a mixed FONLL DIS fktable
+    """
+    if not is_dis(lumi):
+        return False
+    mixed_FONLL_fns = ["FONLL-B", "FONLL-D"]
+    if fns in mixed_FONLL_fns:
         return True
     return False
```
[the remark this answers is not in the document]
felixhekhorn:
```suggestion
    return is_dis(lumi) and fns in ["FONLL-B", "FONLL-D"]
```
even shorter

felixhekhorn (approved):
please apply @RoyStegeman's comments (they are good)

andreab1997 (approved):

src/pineko/check.py
```diff
@@ -116,31 +116,53 @@ def check_grid_and_eko_compatible(pineappl_grid, operators, xif, max_as, max_al)
...
+           luminosity info
+
+    Returns
+    -------
+    bool
+        true if the fktable is a mixed FONLL DIS fktable
+    """
+    if is_dis(lumi) and fns in ["FONLL-B", "FONLL-D"]:
         return True
     return False
```
felixhekhorn, line 158:
```suggestion
    return is_dis(lumi) and fns in ["FONLL-B", "FONLL-D"]
```

PR #109, opened 2023-07-20
Write down the versions for both eko and pineko

### LINKED ISSUE #108: Write the git tag of the programs when version is not available
Just as we do with nnpdf (I can port the code from n3fit to here) so that one can use a development version of eko or pineko and still get that information.

In light of https://github.com/NNPDF/theories_slim/issues/7#issuecomment-1642293486 this might be useful

(and, fair enough, it would be better to create all fktables with tagged versions of the code but everyone knows that's not what is going to happen in reality)

scarlehoff:
Closes #108

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [build-system]
-requires = ["poetry-core>=1.0.0"]
-build-backend = "poetry.core.masonry.api"
+requires = ["poetry-core>=1.0.0", "poetry-dynamic-versioning"]
+build-backend = "poetry_dynamic_versioning.backend"
 
 [tool.poetry]
 name = "pineko"
--- a/src/pineko/evolve.py
+++ b/src/pineko/evolve.py
@@ -140,7 +140,7 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
     # Add the version of eko and pineko to the operator card
     # using importlib.metadata.version to get the correct tag in editable mode
     operators_card["eko_version"] = metadata.version("eko")
-    operators_card["pineko_version"] = metadata.version("pineko")
+    pineko_version = metadata.version("pineko")
 
     # Some safety checks
     if (
@@ -153,6 +153,8 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
 
     with open(card_path, "w", encoding="UTF-8") as f:
         yaml.safe_dump(operators_card, f)
+        f.write(f"# {pineko_version=}")
+
     return operators_card["xgrid"], q2_grid
 
 
--- a/src/pineko/evolve.py
+++ b/src/pineko/evolve.py
@@ -2,6 +2,7 @@
 import copy
 import os
 import pathlib
+from importlib import metadata
 
 import eko
 import eko.basis_rotation as br
@@ -132,6 +133,11 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
         x_grid = np.append(x_grid, 1.0)
         operators_card["configs"]["interpolation_polynomial_degree"] = 1
         operators_card["xgrid"] = x_grid.tolist()
+
+    # Add the version of eko and pineko to the operator card
+    # using importlib.metadata.version to get the correct tag in editable mode
+    operators_card["eko_version"] = metadata.version("eko")
+    operators_card["pineko_version"] = metadata.version("pineko")
 
     with open(card_path, "w", encoding="UTF-8") as f:
         yaml.safe_dump(operators_card, f)
--- a/src/pineko/__init__.py
+++ b/src/pineko/__init__.py
@@ -1,2 +1,3 @@
 """pineko = PineAPPL + EKO."""
 from .cli import command
+from .version import __version__
--- a/src/pineko/theory.py
+++ b/src/pineko/theory.py
@@ -263,9 +263,6 @@ class TheoryBuilder:
         opcard_path = self.operator_cards_path / f"{name}.yaml"
         with open(opcard_path, encoding="utf-8") as f:
             ocard = yaml.safe_load(f)
-        # Drop the pineko_version if it is included as this is metadata for pineko only
-        if "pineko_version" in ocard:
-            ocard.pop("pineko_version")
         return ocard
 
     def activate_logging(self, path, filename, activated_loggers=()):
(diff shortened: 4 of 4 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [build-system]
-requires = ["poetry-core>=1.0.0"]
-build-backend = "poetry.core.masonry.api"
+requires = ["poetry-core>=1.0.0", "poetry-dynamic-versioning"]
+build-backend = "poetry_dynamic_versioning.backend"
```
AleCandido, line 3: Ah, I actually lost that there was also a back-end support. Thanks @scarlehoff!

AleCandido (approved):

scarlehoff:
eko doesn't allow for an unknown key in the runcard, is there a "description" or any other free form field that I can use for the pineko version?

AleCandido:
Nope, there is not...

But EKO should be fully reproducible given its runcards, including the EKO version. Why do you want the Pineko version inside the EKOs?

scarlehoff:
pineko creates the runcard, so we want to know which version of pineko created the runcard

AleCandido:
> pineko creates the runcard, so we I want to know which version of pineko created the runcard

But not in the EKO. That information you could write in the theory.

My point is that EKO is already decoupled from that: whatever version of Pineko created the runcard, if it's the same runcard it will be the same EKO (no matter Pineko).

scarlehoff:
The problem is that in general you might have a theory made with different versions of pineko* and I would like to save that information. The theory, template and eko might be the same but I'd like the operator card to save the pineko version as well.

*because of some ekos being very slow and, eg jets

AleCandido:
Add a log file to the theory?

My point is that I'd like EKO and the EKOs to know nothing about Pineko, for better isolation.

scarlehoff:
But then you need a log file per dataset, is that what you meant?

However, I've done a small change so the information is stripped from the operator card before feeding it to eko.

scarlehoff:
> Not exactly: I meant a single log file, and you log to it adding lines every time (i.e. you cat the log files per-dataset into a single one).

You mean, having a sort of a `operators.yaml` with `<dataset_name>: <pineko_version>` within the `data/operators/<theory number>/` folder?

felixhekhorn:
But the current state is quite useless, no? since the "pineko operator card", which is now different then the "eko operator card", is never saved in the FK and so this information is completely local ... what we should do is open eko at the time of convolution, extract the (eko-) op_card and dump into the metadata ...

AleCandido:
> But the current state is quite useless, no? since the "pineko operator card", which is now different then the "eko operator card", is never saved in the FK and so this information is completely local ... what we should do is open eko at the time of convolution, extract the (eko-) op_card and dump into the metadata ...

The convolution is also done by Pineko, so that is yet a further Pineko version.
I think that storing the information for reproducibility is definitely relevant, but we should also pay attention to store sensible information and not too much, otherwise we might end up wondering about "which Pineko version is this one? the one which generated the runcard? or the one which did the convolution?", or something like that.

About the runcard I really believe there is no need to preserve this information through the EKO. And I would not account for the infinite amount of combinations on how you could generate the runcards: either you used a single version of Pineko for the whole process, or we should add some more free form logs, as the log file proposed above. Keeping structured information is not much more helpful for reproducibility, since you then should propagate it consistently and the propagation process itself might change with the Pineko version...

```
Fully automated = 100% reproducible
Everything else = manually hacked
```

scarlehoff:
> About the runcard I really believe there is no need to preserve this information through the EKO

Indeed! I don't want to preserve the information beyond the operator card itself. If you look at the changes in the code, you can see that I drop it.

The point is that the operator card is a product of pineko, and as a product it should be possible to reproduce it and for that we need to know the pineko version.

The eko is a product of `eko`, and so it only needs to know about its own theory card and the version of `eko` it was produced.

The `fktable` is a product of `pineappl`, `eko` and `pineko` and so it knows all three.

felixhekhorn:
In the case of the eko op_card, I think this is actually useful information - also in the light of sim-FONLL we want to see the grids are convoluted with the correct FNS ...

so I still suggest to add below this line
https://github.com/NNPDF/pineko/blob/3310f4392aa8a649688bced353894bb81ab384c9/src/pineko/evolve.py#L243
`fktable.set_key_value("eko_operator_card", operators.operator_card) `

scarlehoff:
But this issue deals with the operator card, not with the fktable itself.

AleCandido (approved):

scarlehoff:
>  because this creates a new configuration file which is specific to pineko: 

Yes, this is why I was asking for a `description` field or something.
Does `eko` read the card as a yaml? Maybe I can add it as a comment?

felixhekhorn:
> Yes, this is why I was asking for a `description` field or something. Does `eko` read the card as a yaml? Maybe I can add it as a comment?

we currently use this https://github.com/NNPDF/eko/blob/a8ee0a8c90462bfa83a56bcaf3855b7fae461b38/src/eko/io/dictlike.py#L36
so, no - for the moment there is no way to add data beyond the fields of the dataclasses

src/pineko/evolve.py
```diff
@@ -153,6 +153,8 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
 
     with open(card_path, "w", encoding="UTF-8") as f:
         yaml.safe_dump(operators_card, f)
+        f.write(f"# {pineko_version=}")
```
AleCandido, line 156: I guess this was not what you wanted to write, but:
```suggestion
        f.write(f"# pineko_version={pineko_version}")
```
(mainly because of the `=`, that makes little sense without `:` and a width)

scarlehoff: `{pineko_version=}` is shorthand for `pineko_version={pineko_version}`

But I can do `pineko_version: {pineko_version}` instead if you prefer the `:` (it's a comment so it doesn't really matter how is it written)

AleCandido: Ah, great. I forgot about [this option](https://docs.python.org/3/tutorial/inputoutput.html#formatted-string-literals).

The reference to the `:` was only because `=` can be used as a format specifier, but the specifiers should be separated by the identifier with the `:`.

It is fine as it is.

AleCandido: Btw, it was new in py3.8, that by now it is perfectly fine (py3.7 officially died one month ago).
https://docs.python.org/3/whatsnew/3.8.html#bpo-36817-whatsnew

felixhekhorn (approved):

PR #117, opened 2023-09-18
Rotate the eko back to save disk space

andreab1997:
We want to rotate the eko back to the usual grid of 50 points to save disk space. In fact, as noticed by @giacomomagni, depending on the dataset, there is a factor up to ~30 in size between the original eko and the eko after the rotation (for the jets dataset the factor is ~24).

### DIFF
--- a/src/pineko/theory.py
+++ b/src/pineko/theory.py
@@ -395,8 +395,11 @@ class TheoryBuilder:
         if sv_method is None:
             if not np.isclose(xif, 1.0):
                 check_scvar_evolve(grid, max_as, max_al, check.Scale.FACT)
-        # loading ekos
+        # loading ekos to produce a tmp copy
         with eko.EKO.edit(eko_filename) as operators:
+            eko_tmp_path = operators.paths.root.parent / "eko-tmp"
+            operators.deepcopy(eko_tmp_path)
+        with eko.EKO.edit(eko_tmp_path) as operators:
             # Obtain the assumptions hash
             assumptions = theory_card.construct_assumptions(tcard)
             # do it!
@@ -430,6 +433,9 @@ class TheoryBuilder:
                 assumptions=assumptions,
                 comparison_pdf=pdf,
             )
+        # Remove tmp ekos
+        eko_tmp_path.unlink()
+
         logger.info(
             "Finished computation of %s - took %f s",
             name,
--- a/src/pineko/theory.py
+++ b/src/pineko/theory.py
@@ -396,7 +396,7 @@ class TheoryBuilder:
             if not np.isclose(xif, 1.0):
                 check_scvar_evolve(grid, max_as, max_al, check.Scale.FACT)
         # loading ekos to produce a tmp copy
-        with eko.EKO.edit(eko_filename) as operators:
+        with eko.EKO.read(eko_filename) as operators:
             eko_tmp_path = operators.paths.root.parent / "eko-tmp.tar"
             operators.deepcopy(eko_tmp_path)
         with eko.EKO.edit(eko_tmp_path) as operators:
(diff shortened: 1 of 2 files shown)


AleCandido:
Please, avoid rotating back. Just keep the original one (copy/duplicate it before).

Otherwise, you will accumulate more and more numerical error.

AleCandido:
> Yes this was exactly my point. However @giacomomagni pointed out that copying 60gb every time (for the jets ekos) is a bit extreme.

If you do not need the original one, then you can well thrash it, and save space. But if you need it, then instead you should keep.
Notice that there is less computation involved in the copy, rather than going back and forth, so it's not a problem of performances. If it's a problem of disk space, you cancel the rotated EKOs right after. If you have enough space, and you want to save computation, you keep both.

The only case which we can not account for is if you do not have enough space to even perform the operation (copying), and you want to keep the original EKO.
In this case, the optimal solution would be to rotate on the fly, so we should give back an iterator from `manipulate`, that for each value of the scale: loads the corresponding array, rotates it in memory, returns the result without saving it. I.e., if this is a concern, we need a lazy iterator.

felixhekhorn:
> In this case, the optimal solution would be to rotate on the fly, so we should give back an iterator from `manipulate`, that for each value of the scale: loads the corresponding array, rotates it in memory, returns the result without saving it. I.e., if this is a concern, we need a lazy iterator.

Maybe we should start rolling out eko v0.14 with https://github.com/NNPDF/eko/pull/295 and specifically https://github.com/NNPDF/eko/pull/292

giacomomagni:
I'm try to see if this temporary 5016b86 solution is doable also for jets.

giacomomagni (approved):
Thanks @andreab1997 this solves the issue. 
I'd be happy to merge.

src/pineko/theory.py
```diff
@@ -395,8 +395,11 @@ def fk(self, name, grid_path, tcard, pdf):
         if sv_method is None:
             if not np.isclose(xif, 1.0):
                 check_scvar_evolve(grid, max_as, max_al, check.Scale.FACT)
-        # loading ekos
+        # loading ekos to produce a tmp copy
         with eko.EKO.edit(eko_filename) as operators:
+            eko_tmp_path = operators.paths.root.parent / "eko-tmp"
```
felixhekhorn, line 400:
```suggestion
            eko_tmp_path = operators.paths.root.parent / "eko-tmp.tar"
```

src/pineko/theory.py
```diff
@@ -395,8 +395,11 @@ def fk(self, name, grid_path, tcard, pdf):
         if sv_method is None:
             if not np.isclose(xif, 1.0):
                 check_scvar_evolve(grid, max_as, max_al, check.Scale.FACT)
-        # loading ekos
+        # loading ekos to produce a tmp copy
         with eko.EKO.edit(eko_filename) as operators:
```
AleCandido, line 399: Here you do not need `edit` any longer, since you are only deep copying. You can just use `EKO.read`

src/pineko/theory.py
giacomomagni: How come that it worked ?

AleCandido: Most of the programs do not care so much about extensions, they are mostly for human being and editros :)

giacomomagni: Okay it can make sense.

src/pineko/theory.py
andreab1997: True! [8b2a634](https://github.com/NNPDF/pineko/pull/117/commits/8b2a63455585c8873980b958da7c1f3853d46cf2)

AleCandido (approved):

PR #119, opened 2023-09-20
Try to reduce theories form 7 to 5

giacomomagni:
For the FONLL without mixed PTOs (no B,D) and no damping we need to compute just 5 theories not 7.

### DIFF
--- a/src/pineko/cli/fonll.py
+++ b/src/pineko/cli/fonll.py
@@ -45,9 +45,11 @@ def get_list_grids_name(yaml_file):
 @click.argument("dataset", type=str)
 @click.option("--FFNS3", type=int, help="theoryID containing the ffns3 fktable")
 @click.option("--FFN03", type=int, help="theoryID containing the ffn03 fktable")
+@click.option("--FFNS4", type=int, help="theoryID containing the ffns4 fktable")
 @click.option("--FFNS4til", type=int, help="theoryID containing the ffns4til fktable")
 @click.option("--FFNS4bar", type=int, help="theoryID containing the ffns4bar fktable")
 @click.option("--FFN04", type=int, help="theoryID containing the ffn04 fktable")
+@click.option("--FFNS5", type=int, help="theoryID containing the ffns4 fktable")
 @click.option("--FFNS5til", type=int, help="theoryID containing the ffns5til fktable")
 @click.option("--FFNS5bar", type=int, help="theoryID containing the ffns5bar fktable")
 @click.option("--overwrite", is_flag=True, help="Allow files to be overwritten")
@@ -57,9 +59,11 @@ def subcommand(
     dataset,
     ffns3,
     ffn03,
+    ffns4,
     ffns4til,
     ffns4bar,
     ffn04,
+    ffns5,
     ffns5til,
     ffns5bar,
     overwrite,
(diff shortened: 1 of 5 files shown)


src/pineko/fonll.py
```diff
@@ -176,16 +188,33 @@ def produce_combined_fk(
...
-MIXED_ORDER_FNS = ["FONLL-B", "FONLL-D"]
+# plain FONLL schemes
+FNS_LIST = [
+    "FONLL-FFNS",
+    "FONLL-FFN0",
+    "FONLL-FFNS",
+    "FONLL-FFN0",
+    "FONLL-FFNS",
+]
+NFFF_LIST = [3, 3, 4, 4, 5]
```
AleCandido, line 210: @andreab1997 this makes more sense to me, e.g..

But I believe we can even generate this.

andreab1997:
Now that the tests are working again I am even in favor of merging this but I have a couple of comments:

1.  Since the damping (either the theta function or the original FONLL damping) is applied on the tilde and on the zero parts (i.e. for the charm `["ffn03", "ffns4til", "ffn04", "ffns5til"]`), when using the 5 theories (instead of the original 7), where the `tilde` and the `bar` parts are merged, it is not possible to use the damp (as it would also damp the `bar` part while it should not). So one is forced to merge all the fktables at all scales (but it is true that for the moment this is what we are doing by default).
2. There is no check at the moment that checks if the user is providing 7 or 5 theories or even all 9 (which will be just wrong). @giacomomagni

src/pineko/cli/fonll.py
```diff
@@ -51,9 +51,11 @@ def get_list_grids_name(yaml_file):
 @click.argument("dataset", type=str)
 @click.option("--FFNS3", type=int, help="theoryID containing the ffns3 fktable")
 @click.option("--FFN03", type=int, help="theoryID containing the ffn03 fktable")
+@click.option("--FFNS4", type=int, help="theoryID containing the ffns4 fktable")
 @click.option("--FFNS4til", type=int, help="theoryID containing the ffns4til fktable")
 @click.option("--FFNS4bar", type=int, help="theoryID containing the ffns4bar fktable")
 @click.option("--FFN04", type=int, help="theoryID containing the ffn04 fktable")
+@click.option("--FFNS5", type=int, help="theoryID containing the ffns4 fktable")
```
giacomomagni, line 58:
```suggestion
@click.option("--FFNS5", type=int, help="theoryID containing the ffns5 fktable")
```

giacomomagni (commented):
LGTM and it should be working.

> 2. There is no check at the moment that checks if the user is providing 7 or 5 theories or even all 9 (which will be just wrong). @giacomomagni

I'm okay with that as I feel there are quite a lot of settings that can be screwed, but if you believe is necessary I can add it.

andreab1997:
> I'm okay with that as I feel there are quite a lot of settings that can be screwed, so here you always need to pay attention to what you are doing, but if you believe is necessary I can add it.

I agree but exactly for this reason I believe it would be good to have a test for that: at least we are limiting the ways in which we can screw up.
If you can add it, then I will be happy to merge

felixhekhorn:
Actually, for polarized we are using (right now) damping (since we don't have intrinsic), so are we blocking ourself here? right @Radonirinaunimi ?

src/pineko/fonll.py
```diff
@@ -196,14 +225,24 @@ def produce_ptos(fns):
...
+    else:
+        return [base_pto for _ in range(5)]
 
 
-def produce_fonll_recipe(fns):
+def produce_fonll_recipe(fonll_fns, damp):
     """Produce the different theory cards according to which FONLL is asked for."""
     fonll_recipe = []
-    for fns, nfff, po, part in zip(FNS_LIST, NFFF_LIST, produce_ptos(fns), PARTS_LIST):
+    is_mixed = fonll_fns in MIXED_ORDER_FNS or damp != 0
```
giacomomagni, line 239: > Actually, for polarized we are using (right now) damping (since we don't have intrinsic), so are we blocking ourself here? right @Radonirinaunimi ?

DAMP should be working also with this PR. @felixhekhorn ( I mean it's not changed by this PR)

andreab1997: It would not work when using the 5 theories but of course one can use the 7 theories instead

giacomomagni: yes and you will automatically generate 7 theories, so there is no problem.

Radonirinaunimi: @felixhekhorn, just for confirmation, yes exactly! For polarized predictions, DAMP is **necessary** at NLO (especially for FONLL-A), otherwise the matching is not smooth and unstable close to the threshold.

andreab1997:
@felixhekhorn, @giacomomagni Can we merge this?

PR #127, opened 2023-10-10
Move regression test

felixhekhorn:
- Major change is to move the regression test of @scarlehoff to benchmarks (since it is definitely no unit test and takes quite some time)
- I will however add a few other minor changes

### DIFF
--- a/benchmarks/bench_regression.py
+++ b/benchmarks/bench_regression.py
@@ -1,10 +1,10 @@
 """
-    Suite of tests that go through the entire process of creating a new fktable
-    from a empty folder.
+Suite of tests that go through the entire process of creating a new fktable
+from a empty folder.
 
-    The target theory is 400 and the relevant `.toml`, theory runcard and eko template
-    are downloaded from https://github.com/NNPDF/theories during this test so this tests
-    has the double effect of ensuring compatibility between both repositories.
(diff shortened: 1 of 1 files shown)


felixhekhorn:
- @scarlehoff we may want to consider at some point to move the regression test on `theories_slim` instead of `theories`, right?
- @andreab1997 please double check the change in [`merge_grids`](https://github.com/NNPDF/pineko/blob/52096a675f370f3a0e49057fe73fcb4c1a11f517/src/pineko/scale_variations.py#L169) (and by the way: the name `nec_orders` is not the best name ever :see_no_evil: )

andreab1997 (approved):
I had a look everywhere but in particular to `merge_grids` and it looks good @felixhekhorn

andreab1997:
> * @andreab1997 please double check the change in [`merge_grids`](https://github.com/NNPDF/pineko/blob/52096a675f370f3a0e49057fe73fcb4c1a11f517/src/pineko/scale_variations.py#L169) (and by the way: the name `nec_orders` is not the best name ever 🙈 )

I agree it is not the best but I am not good with names. If you have any other suggestions I am happy to apply it :)

felixhekhorn:
> Indeed, it should now look at theories_slim

I leave this to a future PR

PR #128, opened 2023-10-16
Polarized opcards

giacomomagni:
This PR is to allow the computation of polarized Ekos. 

~~As is it now the polarisation is triggered by the name of the DIS observable. 
As **this does not work for Hadronic datasets**, we should refine it.~~

### DIFF
--- a/src/pineko/evolve.py
+++ b/src/pineko/evolve.py
@@ -1,6 +1,5 @@
 """Tools related to evolution/eko."""
 import copy
-import json
 import logging
 import os
 import pathlib
@@ -143,12 +142,10 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
     operators_card["eko_version"] = metadata.version("eko")
     pineko_version = metadata.version("pineko")
 
-    # switch on polarization, only in DIS
-    if "yadism_version" in pineappl_grid.key_values():
-        grid_runcard = pineappl_grid.key_values()["runcard"]
-        # here a grid contains a single obrevable
-        observable = list(json.loads(grid_runcard)["observables"])[0]
-        operators_card["configs"]["polarized"] = observable.startswith("g")
+    # switch on polarization ?
+    kv = pineappl_grid.key_values()
+    if "polarized" in kv:
+        operators_card["configs"]["polarized"] = kv["polarized"] == "True"
 
     # Some safety checks
     if (
--- a/src/pineko/evolve.py
+++ b/src/pineko/evolve.py
@@ -144,7 +144,7 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
     pineko_version = metadata.version("pineko")
 
     # switch on polarization, only in DIS
-    if "runcard" in pineappl_grid.key_values():
+    if "yadism_version" in pineappl_grid.key_values():
         grid_runcard = pineappl_grid.key_values()["runcard"]
         # here a grid contains a single obrevable
         observable = list(json.loads(grid_runcard)["observables"])[0]
--- a/src/pineko/evolve.py
+++ b/src/pineko/evolve.py
@@ -148,7 +148,7 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
         grid_runcard = pineappl_grid.key_values()["runcard"]
         # here a grid contains a single obrevable
         observable = list(json.loads(grid_runcard)["observables"])[0]
-        operators_card["configs"]["polarized"] = "g" in observable.split("_")[0]
+        operators_card["configs"]["polarized"] = observable.startswith("g")
 
     # Some safety checks
     if (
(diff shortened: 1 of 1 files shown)


src/pineko/evolve.py
```diff
@@ -142,6 +143,13 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
     operators_card["eko_version"] = metadata.version("eko")
     pineko_version = metadata.version("pineko")
 
+    # switch on polarization, only in DIS
+    if "runcard" in pineappl_grid.key_values():
+        grid_runcard = pineappl_grid.key_values()["runcard"]
+        # here a grid contains a single obrevable
+        observable = list(json.loads(grid_runcard)["observables"])[0]
+        operators_card["configs"]["polarized"] = "g" in observable.split("_")[0]
```
Radonirinaunimi, line 151: The following is very minor but might be slightly restrictive.

```suggestion
        operators_card["configs"]["polarized"] = observable.startswith("g")
```

Radonirinaunimi (commented):
This looks good as long as we stick to the idea that any term that multiplies the polarised structure functions are accounted as $k$-factors (hence one single observable) $-$ which will always be the case unless we do a simultaneous fit. In addition to this, no polarised observable are given as a linear combination of (polarised) structure functions. So, this LGTM!

giacomomagni:
> This looks good as long as we stick to the idea that any term that multiplies the polarised structure functions are accounted as k-factors (hence one single observable) − which will always be the case unless we do a simultaneous fit. 

This should not be a problem even in that case as the dataset will contain a 2FKtables so 2 Operator cards will be needed.

> In addition to this, no polarised observable are given as a linear combination of (polarised) structure functions. So, this LGTM!

Any new polarised DIS observable not starting with `g` (or viceversa) could be a problem. 

For hadronic datasets we can add and `elif` once such grids will be available.

src/pineko/evolve.py
```diff
@@ -142,6 +143,13 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
     operators_card["eko_version"] = metadata.version("eko")
     pineko_version = metadata.version("pineko")
 
+    # switch on polarization, only in DIS
+    if "yadism_version" in pineappl_grid.key_values():
+        grid_runcard = pineappl_grid.key_values()["runcard"]
```
felixhekhorn, line 148: why don't we do immediately the correct thing? i.e. attribute it to the grid
1. Adjust [`yadbox.export.dump_pineappl_to_file`](https://github.com/NNPDF/yadism/blob/d6366b5e7e078c02193c1cfa5df0639b3f9c19d5/src/yadbox/export.py#L104) by adding
```py
grid.set_key_value("polarized", obsname.startswith("g"))
```
(we could even add a method/property to `ObservableName` since I seem to remember to have seen that test before)
2. do here
```py
kv = pineappl_grid.key_values()
operators_card["configs"]["polarized"] = "polarized" in kv and bool(kv["polarized"])
```

giacomomagni: This is also a solution.

felixhekhorn: and moreover without any additional effort for double hadronic cases on the pineko side

giacomomagni: sure but in principle yadism doesn't need that information and a single grid can contain polarised and unpolarised observables.
And same can hold for the future hadronic polarised tool.

felixhekhorn: > sure but in principle yadism doesn't need that information and a single grid can contain polarised and unpolarised observables. 

yadism (native) output yes, [yadbox.export.dump_pineappl_to_file](https://github.com/NNPDF/yadism/blob/d6366b5e7e078c02193c1cfa5df0639b3f9c19d5/src/yadbox/export.py#L104) no

giacomomagni: 1. [yadbox.export.dump_pineappl_to_file](https://github.com/NNPDF/yadism/blob/d6366b5e7e078c02193c1cfa5df0639b3f9c19d5/src/yadbox/export.py#L104) adjusted in: https://github.com/NNPDF/yadism/commit/508c4feb2d4a91d2c91820a25790043cf62f6faa
2. Added in https://github.com/NNPDF/pineko/pull/128/commits/ca5eca6b91ce1b3efa09ae49bbea8b3afd43aa56

Radonirinaunimi: I was not fully aware of this but I like it very much this way 🙃

andreab1997 (approved):

Radonirinaunimi (approved):
If this could be merged soon (followed by a release) that'd be great 🙃

felixhekhorn (approved):
That's the minimal change I was expecting :upside_down_face:

felixhekhorn:
@Radonirinaunimi feel free to make immediately an release (else #99 will for sure trigger a release, but maybe you want to disentangle :innocent: )

Radonirinaunimi:
> @Radonirinaunimi feel free to make immediately an release (else #99 will for sure trigger a release, but maybe you want to disentangle 😇 )

Thanks @felixhekhorn ! We surely want a release in which this is first disentangle from #99.

PR #131, opened 2023-10-19
Group FONLL configuration by theory

felixhekhorn:
Please tell me and @andreab1997 if you prefer this or the original

### DIFF
--- a/src/pineko/cli/fonll.py
+++ b/src/pineko/cli/fonll.py
@@ -171,4 +171,5 @@ def fonll_tcards(theoryid, cfg):
     tcard_parent_path = theory_card.path(theoryid).parent
     if "FONLL" not in tcard["FNS"]:
         raise TheoryCardError("The theorycard does not correspond to an FONLL scheme.")
-    _ = fonll.produce_fonll_tcards(tcard, tcard_parent_path, theoryid)
+    fonll.dump_tcards(tcard, tcard_parent_path, theoryid)
+    return
--- a/src/pineko/fonll.py
+++ b/src/pineko/fonll.py
@@ -41,7 +41,7 @@ class FonllSchemeConfig:
         """Return the base perturbative order of the fonll fns."""
         return FNS_BASE_PTO[self.fonll_fns]
 
-    def subfks_fns(self) -> list:
+    def subfks_fns(self) -> list[str]:
         """Return the fns of the sub FK table components of the fonll fns."""
         if self.is_mixed() or self.is_damp:
             return [
@@ -62,14 +62,14 @@ class FonllSchemeConfig:
                 "FONLL-FFNS",
             ]
 
-    def subfks_nfff(self) -> list:
+    def subfks_nfff(self) -> list[int]:
         """Return the nfff of the sub FK table components of the fonll fns."""
         if self.is_mixed() or self.is_damp:
             return [3, 3, 4, 4, 4, 5, 5]
         else:
             return [3, 3, 4, 4, 5]
 
-    def subfks_parts(self) -> list:
+    def subfks_parts(self) -> list[str]:
         """Return the parts of the sub FK table components of the fonll fns that need to be computed."""
         if self.is_mixed() or self.is_damp:
             return [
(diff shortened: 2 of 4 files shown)


RoyStegeman:
Thanks, I'd say this is simpler and clearer so I'd prefer a solution along these lines.

src/pineko/cli/fonll.py
```diff
@@ -171,4 +171,5 @@ def fonll_tcards(theoryid, cfg):
     tcard_parent_path = theory_card.path(theoryid).parent
     if "FONLL" not in tcard["FNS"]:
         raise TheoryCardError("The theorycard does not correspond to an FONLL scheme.")
-    _ = fonll.produce_fonll_tcards(tcard, tcard_parent_path, theoryid)
+    fonll.dump_tcards(tcard, tcard_parent_path, theoryid)
+    return
```
AleCandido, line 175: You do not need to `return` explicitly. Empty `return` returns `None`, and the same happens if you reach the end of the function without returning.

felixhekhorn:
```suggestion
```
you're right

src/pineko/fonll.py
```diff
@@ -282,40 +179,77 @@ def produce_combined_fk(
...
+"""Mixed FONLL schemes."""
+
+FNS_CONFIG = [
+    SubTheoryConfig("FONLL-FFNS", 3, "full"),
+    SubTheoryConfig("FONLL-FFN0", 3, "full"),
+    SubTheoryConfig("FONLL-FFNS", 4, "full"),
+    SubTheoryConfig("FONLL-FFN0", 4, "full"),
+    SubTheoryConfig("FONLL-FFNS", 5, "full"),
+]
+"""Plain FONLL schemes."""
```
AleCandido, line 214: It's much tidier than before, and this I like a lot.

But it's still the full spell-out, and this I would avoid.

Leveraging the object, you could start removing some details. E.g. You just need a boolean flag to distinguish `FONLL-FFNS` from `FONLL-FFN0`, and make the string a `@property` (that essentially concatenates the fixed prefix `FONLL-FFN` with a single character, flag-dependent).

Even `full`, `massive`,  `massless`, are one out of 3, so it can be an enum. However, it's not only about making it an enum, but they can also be functions:
- `FNS` -> `full`
- `MIXED` -> `delta_pto = 1`
    - the second for each flavor is always `massive`, and the first `massless`, but for 3 flavors
    - this is the definition, and it'd scale for each number of flavors (that we do not need, but it would make the decision more transparent)

I'm not asking necessarily everything at once. But nFONLL is defined by some rules, and coding the rules in the algorithm is more transparent than the bare result.

AleCandido: With all this, that the result of whatever function (possibly calling sub-functions) is going to be a `SubTheoryConfig` is a great idea. And (as said above) I like it a lot.

felixhekhorn: > Leveraging the object, you could start removing some details. E.g. You just need a boolean flag to distinguish `FONLL-FFNS` from `FONLL-FFN0`, and make the string a `@property` (that essentially concatenates the fixed prefix `FONLL-FFN` with a single character, flag-dependent).

this was easy: b5674d0

the others are a bit more evolved ...

and particularly, I still believe there is something fishy with `is_mixed` and `is_damped` since they are strictly orthogonal ... and to be specific: it might well be that the name `MIXED_FNS_CONFIG` is simply misleading and it should be really `EXTENDED_CONFIG` which happens to be used for either damping or mixed is used (but has nothing to do with FNS)

AleCandido: > it might well be that the name `MIXED_FNS_CONFIG` is simply misleading and it should be really `EXTENDED_CONFIG` which happens to be used for either damping or mixed is used (but has nothing to do with FNS)

Well, I acknowledge that it's not trivial, and I said it should not come necessarily now. But this kind of ambiguity would be almost over if the algorithm were transparent...

felixhekhorn: let's keep further improvements for later then

AleCandido:
You could merge even without approval (it's merge in a branch, not in `main`).

felixhekhorn:
> You could merge even without approval (it's merge in a branch, not in `main`).

Well, I need solid arguments to convince @andreab1997 :upside_down_face:

PR #132, opened 2023-10-23
Generate kfactors for polarized observables

toonhasenack:
Initial pull request for generating the K-factors of the F1 structure functions for polarized DIS data.


giacomomagni:
This PR is mostly to keep track of @toonhasenack work, we might no merge this in the end.

data/kfactors/generate.py
```diff
@@ -0,0 +1,129 @@
...
+
+    date = dt.now().date()
+    string = (
+        f"""********************************************************************************
+SetName: {dataset_name}
+Author: {author_name}
+Date: {date}
+CodesUsed: https://github.com/NNPDF/yadism
+TheoryInput: {theory_name}
+PDFset: NNPDF40_nnlo_pch_as_01180
```
giacomomagni, line 88: This can also be a variable.

data/kfactors/generate.py
```diff
@@ -0,0 +1,129 @@
...
+Warnings: F1 normalization for {dataset_name}
+********************************************************************************
+"""
+        + strf_data
+    )
+
+    os.makedirs(output_name, exist_ok=True)
+    with open(
+        output_name + f"/CF_QCD_{dataset_name}.dat".replace("F1", "G1"), "w"
```
giacomomagni, line 97:
```suggestion
        output_name + f"/CF_NRM_{dataset_name}.dat".replace("F1", "G1"), "w"
```

data/kfactors/generate.py
```diff
@@ -0,0 +1,129 @@
...
+import numpy as np
+import os
+import argparse
+
+
+def get_gpaths(folder):
+    """
+    Get a list of paths to PineAPPL grids in the specified folder.
+
+    Args:
```
felixhekhorn, line 14: Please use [numpy documentation style](https://www.sphinx-doc.org/en/master/usage/extensions/example_numpy.html)

data/kfactors/generate.py
```diff
@@ -0,0 +1,129 @@
+import lhapdf
+import pineappl
+from glob import glob
+from datetime import datetime as dt
+import numpy as np
+import os
+import argparse
+
+
+def get_gpaths(folder):
```
felixhekhorn, line 10: Consider adding [type hints](https://docs.python.org/3/library/typing.html?highlight=any#module-typing)

data/kfactors/generate.py
```diff
@@ -0,0 +1,129 @@
...
+        author_name (str): Name of the author.
+        theory_name (str): Name of the theory.
+        output_name (str): Output folder name.
+    """
+    strf_data = ""
+    for i in range(data.shape[0]):
+        strf_data += f"{data[i]}  0.0\n"
+
+    date = dt.now().date()
+    string = (
```
felixhekhorn, line 81: Make this a module constant

data/kfactors/generate.py
```diff
@@ -0,0 +1,129 @@
...
+CodesUsed: https://github.com/NNPDF/yadism
+TheoryInput: {theory_name}
+PDFset: NNPDF40_nnlo_pch_as_01180
+Warnings: F1 normalization for {dataset_name}
+********************************************************************************
+"""
+        + strf_data
+    )
+
+    os.makedirs(output_name, exist_ok=True)
```
felixhekhorn, line 95: Maybe prefer [`pathlib`](https://docs.python.org/3/library/pathlib.html?highlight=pathlib#module-pathlib) over `os`

felixhekhorn:
> This PR is mostly to keep track of @toonhasenack work, we might no merge this in the end.

Maybe this is not a priority, but I think we do want to merge because this strategy will persists also in the mid-term

toonhasenack:
See latest push for adjustments

felixhekhorn:
> > Maybe this is not a priority, but I think we do want to merge because this strategy will persists also in the mid-term
> 
> Are you sure that this should belong to Pineko? It is quite DIS specific... (even more, just specific to some observable)

I think it is fine to have it here, pineko has the k-factor option - of course the other option is to push it back to yadism

alecandido:
> I think it is fine to have it here, pineko has the k-factor option - of course the other option is to push it back to yadism

Pineko is applying k-factors, but not computing them. We always kept Pineko process agnostic, and I believe that's a good thing (since all the complexity of the processes is delegated to the respective generators, and we avoid coupling to a specific one - in particular DIS, as in APFEL).

giacomomagni:
> * keep the fixes for SV and add (at least) `timelike` (cc @t7phy ) and please do a proper iteration - actually, why don't we transfer _all_ keys @andreab1997 ?

Yes indeed that would be the proper fix. I can take care of it, here.

andreab1997:
> > > I think it is fine to have it here, pineko has the k-factor option - of course the other option is to push it back to yadism
> > 
> > 
> > Pineko is applying k-factors, but not computing them. We always kept Pineko process agnostic, and I believe that's a good thing (since all the complexity of the processes is delegated to the respective generators, and we avoid coupling to a specific one - in particular DIS, as in APFEL).
> 
> okay, I agree - so @toonhasenack
> 
> * please move the scripts to `yadism/extras`
> * keep the fixes for SV and add (at least) `timelike` (cc @t7phy ) and please do a proper iteration - actually, why don't we transfer _all_ keys @andreab1997 ?

What do you mean with all keys?

felixhekhorn:
> > What do you mean with all keys?
> 
> all metadata present in original grid like `result`, generator info, etc. (pinefarm can add an arbitrary amount of metadata, which is also unpredictable) - I think @giacomomagni understood ... in practice only a `for k,v in ori.key_values()` is needed

this is done in #161 - so please just close @toonhasenack when you have moved the scripts

felixhekhorn:
@toonhasenack please don't push here. Close this one an reopen in yadism.

data/kfactors/generate_ALL.py
```diff
@@ -0,0 +1,112 @@
...
+        data = yaml.safe_load(file)
+
+    bins = data["bins"]
+    prediction = np.zeros(len(bins))
+    for i, bin in enumerate(bins):
+        prediction[i] = (
+            bin["y"]["mid"]
+            * (2 - bin["y"]["mid"])
+            / (bin["y"]["mid"] ** 2 + 2 * (1 - bin["y"]["mid"]))
+        )
```
Radonirinaunimi, line 39: @toonhasenack  Here we are missing the $1/(2xF_1)$ factor.

felixhekhorn:
moved to https://github.com/NNPDF/yadism/pull/270

PR #134, opened 2023-10-27
account for k{fl}Thr in damping step function fonll

RoyStegeman:
@niclaurenti this should correctly account for the shifted threshold in damping

### DIFF
--- a/src/pineko/fonll.py
+++ b/src/pineko/fonll.py
@@ -122,9 +122,11 @@ def produce_dampings(theorycard_constituent_fks, fonll_info, damp):
     """Return the damping factors for each of the relevant masses."""
     mc = theorycard_constituent_fks["mc"]
     mb = theorycard_constituent_fks["mb"]
+    kcthr = theorycard_constituent_fks["kcThr"]
+    kbthr = theorycard_constituent_fks["kbThr"]
     q2grid = fonll_info.Q2grid
-    step_function_charm = mc**2 < q2grid
-    step_function_bottom = mb**2 < q2grid
+    step_function_charm = (kcthr * mc) ** 2 < q2grid
+    step_function_bottom = (kbthr * mb) ** 2 < q2grid
     damping_factor_charm = (1 - mc / q2grid) ** damp[1]
     damping_factor_bottom = (1 - mb / q2grid) ** damp[1]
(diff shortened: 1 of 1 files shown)


src/pineko/fonll.py
```diff
@@ -122,9 +122,11 @@ def produce_dampings(theorycard_constituent_fks, fonll_info, damp):
...
     mb = theorycard_constituent_fks["mb"]
+    kcthr = theorycard_constituent_fks["kcThr"]
+    kbthr = theorycard_constituent_fks["kbThr"]
     q2grid = fonll_info.Q2grid
-    step_function_charm = mc**2 < q2grid
-    step_function_bottom = mb**2 < q2grid
+    step_function_charm = (kcthr * mc) ** 2 < q2grid
+    step_function_bottom = (kbthr * mb) ** 2 < q2grid
     damping_factor_charm = (1 - mc / q2grid) ** damp[1]
     damping_factor_bottom = (1 - mb / q2grid) ** damp[1]
```
AleCandido, line 131: Shouldn't be the matching even in the *damping factor*?

andreab1997 (approved):
Thanks for this, it looks good. Sorry if I did not realize this was missing.

src/pineko/fonll.py
RoyStegeman: Ah right, for the function to be smooth yes

RoyStegeman: Not sure what is done in aFONLL?

andreab1997: I was asking myself the same but I believe we should not, at least according to the paper (if I remember correctly)

AleCandido: The FONLL paper? In that case, I believe the matching is taken to be the mass.

The damp is there to make things continuous, so it should smooth at matching scale, no way out...

RoyStegeman: Looking at yadism, aFONLL does what this PR does now. Though I agree that since my understanding is that the damping factor is there to ensure a smooth transition, it should vanish at the matching scale and not at `mb`.

RoyStegeman: Sorry didn't see your message before mine, but fine, looks like we agree :smile:

PR #140, opened 2023-12-01
Correct c and bmatching to be squared

andreab1997:


andreab1997:
If someone like @felixhekhorn or @giacomomagni can please run pre-commit it would be good because my pre-commit seems to be broken now

RoyStegeman:
I didn't look, but does `tcards_fonll` require any changes?

andreab1997:
> I didn't look, but does `tcards_fonll` require any changes?

I don't believe so

felixhekhorn (approved):
1. this is indeed a bug - good you found it!
2. we will need to adjust all theory cards, no? adding the two new fields

andreab1997:
> 1. this is indeed a bug - good you found it!
> 2. we will need to adjust all theory cards, no? adding the two new fields

Yes, can some of you take care of this in `theories_slim` while I produce the grids and fktables? Otherwise I will do it later

RoyStegeman:
> Yes, can some of you take care of this in theories_slim while I produce the grids and fktables? Otherwise I will do it later

I can, but we need to discuss for which theories we want to.

RoyStegeman (approved):

andreab1997:
> > Yes, can some of you take care of this in theories_slim while I produce the grids and fktables? Otherwise I will do it later
> 
> I can, but we need to discuss if we want to have 0 or 2 `DAMPPOWERb` at NNLO and N3LO.

I would say we should have DAMP: 1 at NLO, NNLO and N3LO because at least the theta we want. Also DAMPPOWERc: 0 at NLO, NNLO and N3LO. Then at NLO I would do DAMPPOWERb: 2, while at NNLO and N3LO I would do DAMPPOWERb: 0

RoyStegeman:
Okay I'll update them

RoyStegeman:
Actually, should we do damping in the same way as the old theories. Meaning that `DAMP: 0` does only the step function and `DAMP: 1` requires also `DAMPPOWERc` and `DAMPPOWERb`? This reduces flexibility but keeps the meaning of `DAMP` more consistent among theories. 

I'm not saying this is what we should do was just thinking about it when thinking about what to do with the theorycard for theory 400 in https://github.com/NNPDF/theories_slim/pull/18

andreab1997:
> Actually, should we do damping in the same way as the old theories. Meaning that `DAMP: 0` does only the step function and `DAMP: 1` requires also `DAMPPOWERc` and `DAMPPOWERb`? This reduces flexibility but keeps the meaning of `DAMP` more consistent among theories (if I understand the old situation correctly).
> 
> I'm not saying this is what we should do was just wondering about it when thinking about what to do with the theorycard for theory 400 in [NNPDF/theories_slim#18](https://github.com/NNPDF/theories_slim/pull/18)

Yes maybe we should. However in this way it would not be possible anymore to avoid damping completely

felixhekhorn:
> However in this way it would not be possible anymore to avoid damping completely

we could abuse the fact that it is an int and say 1: damping with function, 0: damping with theta, -1: all FKs always everywhere

or do we consider -1 to be unphysical? (I can't see an obvious reason why we should ...)

RoyStegeman:
Since the -1 would probably only be a pineko feature for internal/development use and never end up in a final theory that could work, it would just need to be documented somewhere.

andreab1997:
> > However in this way it would not be possible anymore to avoid damping completely
> 
> we could abuse the fact that it is an int and say 1: damping with function, 0: damping with theta, -1: all FKs always everywhere
> 
> or do we consider -1 to be unphysical? (I can't see an obvious reason why we should ...)

Yes I agree with this!

src/pineko/cli/fonll.py
```diff
@@ -119,10 +119,14 @@ def subcommand(
 
     # Get theory info
     tcard = theory_card.load(theoryid)
-    if not "DAMPPOWER" in tcard:
-        if tcard["DAMP"] != 0:
-            raise InconsistentInputsError("If DAMP is set, set also DAMPPOWER")
-        tcard["DAMPPOWER"] = None
+    if tcard["DAMP"] not in [-1,0]:
```
RoyStegeman, line 122:
```suggestion
    if tcard["DAMP"] == 1:
```

RoyStegeman:
> (and also please run pre-commit)

pretty sure these last changes are fine

RoyStegeman:
Okay I think this is ready to be merged

PR #142, opened 2023-12-04
Fix condition to new damping convention

### LINKED ISSUE #141: Fix new `DAMP` convention
In #140, we have changed the convention for `DAMP: 0`, but this line has not been fixed accordingly. 

https://github.com/NNPDF/pineko/blob/68c41e774c34952542441b8f5241297f7ed3775b/src/pineko/fonll.py#L238

It should be replaced with: 

```
 is_damped = damp != -1
 ```

### PR DESCRIPTION
**Pull request by felixhekhorn**
Closes #141

### DIFF
--- a/src/pineko/fonll.py
+++ b/src/pineko/fonll.py
@@ -235,7 +235,7 @@ def collect_updates(fonll_fns, damp):
     """Produce the different theory cards according to which FONLL is asked for."""
     updates = []
     is_mixed = fonll_fns in MIXED_ORDER_FNS
-    is_damped = damp != 0
+    is_damped = damp >= 0
     base_pto = FNS_BASE_PTO[fonll_fns]
     cfgs = MIXED_FNS_CONFIG if is_mixed or is_damped else FNS_CONFIG
     for cfg in cfgs:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,7 +20,7 @@ repos:
     hooks:
       - id: pycln
         args: [--config=pyproject.toml]
-  - repo: https://github.com/psf/black
+  - repo: https://github.com/psf/black-pre-commit-mirror
     rev: 23.11.0
     hooks:
       - id: black

### CODE REVIEW

File: src/pineko/fonll.py [hunk 1]
```diff
@@ -235,7 +239,7 @@ def collect_updates(fonll_fns, damp):
     """Produce the different theory cards according to which FONLL is asked for."""
     updates = []
     is_mixed = fonll_fns in MIXED_ORDER_FNS
-    is_damped = damp != 0
+    is_damped = damp >= 0
```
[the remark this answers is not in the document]
felixhekhorn: done in [643313b](https://github.com/NNPDF/pineko/pull/142/commits/643313bc1707e4f3f6f6349b5a49bfaea49ecc23) - I also had to adjust the original fix since I broke the benchmarks .... (I guess I'm still confused by the fact that mixing and damping are the same ...)

src/pineko/fonll.py
```diff
@@ -235,7 +239,7 @@ def collect_updates(fonll_fns, damp):
     """Produce the different theory cards according to which FONLL is asked for."""
     updates = []
     is_mixed = fonll_fns in MIXED_ORDER_FNS
-    is_damped = damp != 0
+    is_damped = damp > 0
```
AleCandido, line 242: Just to remind/clarify: in
https://github.com/NNPDF/pineko/issues/141#issuecomment-1837481519
I suggested not doing this...

I don't have such a strong opinion, so in the end just do whatever you wish.

src/pineko/fonll.py
```diff
@@ -235,7 +239,7 @@ def collect_updates(fonll_fns, damp):
     """Produce the different theory cards according to which FONLL is asked for."""
     updates = []
     is_mixed = fonll_fns in MIXED_ORDER_FNS
-    is_damped = damp != 0
+    is_damped = damp == 1
```
giacomomagni, line 242:
```suggestion
    is_damped = damp != -1
```

AleCandido: It doesn't look correct to me, since we agreed that `0` should be *damped* (in the sense of the `theta` function, and thus the extra sub-tables).

> [cheezburger.com/6060857344/it-hurt-itself-in-its-confusion](https://cheezburger.com/6060857344/it-hurt-itself-in-its-confusion)

I perfectly agree with @giacomomagni (and if it's a benchmark problem, most likely the benchmarks will be outdated)

felixhekhorn: I suspect the variable should be called `is_non_trivially_damped`, but better @andreab1997 should comment

andreab1997: @giacomomagni are you doing that here? Otherwise I'll do it as soon as I can

andreab1997:
@felixhekhorn and @giacomomagni please have a look when you have time. Note that I also reworked a bit the checks that we do at the beginning.

src/pineko/fonll.py
```diff
@@ -224,9 +222,11 @@ def scheme(self):
 FNS_CONFIG = [
```
giacomomagni, line 222: Now `MIXED_FNS_CONFIG` and `FNS_CONFIG` should coincide no?

andreab1997: The only difference between the two is the `delta_pto`. Maybe we can get rid of on of the two

src/pineko/fonll.py
```diff
@@ -260,7 +248,8 @@ def collect_updates(fonll_fns, damp):
 def dump_tcards(tcard, tcard_parent_path, theoryid):
     """Produce the seven FONLL theory cards from the original one.
 
-    The produced theory cards are dumped in `tcard_parent_path` with names from '{theoryid}00.yaml' to '{theoryid}06.yaml'.
+    The produced theory cards are dumped in `tcard_parent_path` with names
+    from '{theoryid}00.yaml' to '{theoryid}06.yaml'.
     """
     updates = collect_updates(tcard["FNS"], tcard["DAMP"])
```
giacomomagni, line 254:
```suggestion
    updates = collect_updates(tcard["FNS"])
```
this seems correct as right now there will be always 7 theories

src/pineko/fonll.py
giacomomagni, line 1: To be really picky also `FNS_BASE_PTO` is now redundant as the `pto` can be found in the base theory card. 
Maybe you can use it for a check, something like: you do `FONLL-B` with `PTO: 2` you crash ...

AleCandido:
Just a small note: I like a lot the idea of simplifying the FONLL handling. But even simplifications are changes, and changes are the opposite of stability.

Make sure to make careful checks before merging.

andreab1997:
> Just a small note: I like a lot the idea of simplifying the FONLL handling. But even simplifications are changes, and changes are the opposite of stability.
> 
> Make sure to make careful checks before merging.

Yes sure, you are right, I am perfectly aware of the risks of reworking a code that works

andreab1997:
@felixhekhorn, @giacomomagni Can you please review this when you have time?

giacomomagni (approved):

andreab1997:
I tried to reproduce the fktables with this branch and everything seem to work fine. @felixhekhorn when you are happy we can merge

felixhekhorn:
I trust you tested it in real life :upside_down_face:

PR #145, opened 2023-12-14
Add N3LO ren SV

### LINKED ISSUE #144: Generated renormalization SV does not work at N3LO (and beyond)
[`ren_sv_coeffs`](https://github.com/NNPDF/pineko/blob/e2cb2415ff4c1a952956a5386dbee129b66217b5/src/pineko/scale_variations.py#L33) does not work at N3LO, as it seems hardcoded to NNLO:
https://github.com/NNPDF/pineko/blob/e2cb2415ff4c1a952956a5386dbee129b66217b5/src/pineko/scale_variations.py#L56-L63
If you don't want to do the fully general math (which I guess should be [Bell polynomials](https://en.wikipedia.org/wiki/Bell_polynomials) - look at the second formula image under properties), you could still do an explicit spell out (as N4LO is far future, hopefully) ...

The relevant coeffs can be found in the MHOU paper:
grafik

cc @giacomomagni

felixhekhorn:
Closes #144

I hope I did the correct thing :see_no_evil: ... (unfortunately the module is not so clearly written ...) if you agree I can add a N3LO unit test

### DIFF
--- a/src/pineko/scale_variations.py
+++ b/src/pineko/scale_variations.py
@@ -51,16 +51,32 @@ def ren_sv_coeffs(m, max_as, logpart, which_part, nf):
     float
         renormalization scale variation contribution
     """
-    bcoeff = beta.beta_qcd((max_as - logpart - which_part + 2, 0), nf)
-    as_normalization = AS_NORM ** (max_as - which_part)
+    # nothing to do?
     if max_as == 0:
         return 0.0
-    if max_as == 2:
-        if which_part > 0:
-            m += 1
-        elif logpart > 1:
-            m = 0.5 * m * (m + 1)
-    return m * as_normalization * bcoeff
+    # eko uses as = alpha_s/(4pi), but pineappl just alpha_s
+    as_normalization = AS_NORM ** (max_as - which_part)
+    # the coefficients can be found in the NNLO MHOU paper
+    # (which also contains a generating MMa script)
+    beta0 = beta.beta_qcd_as2(nf)
+    beta1 = beta.beta_qcd_as3(nf)
+    beta2 = beta.beta_qcd_as4(nf)
+    ren_coeffs = {
+        # NLO
+        (1, 1, 0): m * beta0,
+        # NNLO
+        (2, 1, 1): (m + 1) * beta0,
+        (2, 1, 0): (m + 0) * beta1,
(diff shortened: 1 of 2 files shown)


andreab1997 (commented):
Thanks for this. It looks ok (even if I need to have a more careful look). My only comment is that this function is only called to get one contribution, while in this way every time it needs to compute all the terms, while returning only one. This seems a waste to me.

felixhekhorn:
> Thanks for this. It looks ok (even if I need to have a more careful look). My only comment is that this function is only called to get one contribution, while in this way every time it needs to compute all the terms, while returning only one. This seems a waste to me.

True - but I'm sure we can afford the calculation of O(10) polynomials for the sake of clarity

andreab1997 (approved):

src/pineko/scale_variations.py
```diff
@@ -51,16 +51,32 @@ def ren_sv_coeffs(m, max_as, logpart, which_part, nf):
...
+        (2, 1, 1): (m + 1) * beta0,
+        (2, 1, 0): (m + 0) * beta1,
+        (2, 2, 0): m * (m + 1) / 2.0 * beta0**2,
+        # N3LO
+        (3, 1, 2): (m + 2) * beta0,
+        (3, 1, 1): (m + 1) * beta1,
+        (3, 1, 0): (m + 0) * beta2,
+        (3, 2, 1): (m + 1) * (m + 2) / 2.0 * beta0**2,
+        (3, 2, 0): m * (2 * m + 3) / 2.0 * beta0 * beta1,
+        (3, 3, 0): m * (m + 1) * (m + 2) / 6 * beta0**2,
```
giacomomagni, line 77:
```suggestion
        (3, 3, 0): m * (m + 1) * (m + 2) / 6 * beta0**3,
```

giacomomagni (changes requested):

src/pineko/scale_variations.py
andreab1997: Well spotted! Sorry I did not notice

felixhekhorn: Conclusion: triple check is not enough, always ask someone else :innocent:

giacomomagni (approved):

Issue #130, opened 2023-10-18
Reuse ekos for nFONLL


felixhekhorn:
FFNS and FFN0 are the same from the evolution point of view so there is no need to compute the same eko twice, we can just inherit

PR #153, opened 2024-02-15
Automatise numerical FONLL

giacomomagni:
In this PR we try to automatise the procedure to generate numerical FONLL fk tables. 

- [x] We introduce 2 additional commands that will do the job, reusing the ekos where possible #130.
- [x] We add a documentation how to run numerical fonll. 
- [x] Fix `FONLL-B` bug.
- [x] Test and review.
- [ ] Expand documentation

### DIFF
--- a/docs/source/overview/running.rst
+++ b/docs/source/overview/running.rst
@@ -68,7 +68,7 @@ in the final step.
 There are 2 workflows possible, one in which all the steps are performed individually
 and one that is more automatized.
 
-The automatize workflows always assumes that |FONLL| is performed both for
+The automatized workflows always assumes that |FONLL| is performed both for
 charm and bottom effects.
 
 The manual procedure works as follow:
--- a/docs/source/overview/running.rst
+++ b/docs/source/overview/running.rst
@@ -122,7 +122,7 @@ In this case the workflows is:
       pineko fonll_ekos THEORY_ID DATASET1 DATASET2 ...
 
     Note: this is usually an expensive operation as multiple ekos are run sequentially.
-    Depending on the resources that you have available it might be more convinient
+    Depending on the resources that you have available it might be more convenient
     to call the command separately for each DATASET.
 
   4. Generate the final |FONLL| FK table directly running::
--- a/docs/source/overview/running.rst
+++ b/docs/source/overview/running.rst
@@ -1,4 +1,3 @@
-################
 Running `pineko`
 ################
 
@@ -27,8 +26,6 @@ This is a two step process:
 
     pineko theory ekos THEORY_ID DATASET1 DATASET2 ...
 
-
-
 Inherit |EKO| or grids from Existing Theory
 """""""""""""""""""""""""""""""""""""""""""
 
@@ -40,7 +37,8 @@ You can reuse the grid from a different theory by running::
 
   pineko theory inherit-grids SOURCE_THEORY_ID TARGET_THEORY_ID DATASET1 DATASET2 ...
 
-The relation between the source theory and the target theory is non-trivial [5]_.
+The relation between the source theory and the target theory is non-trivial
+(e.g. they may differ by scale variations, different DIS settings, etc).
 
 Generating the FK Table
 -----------------------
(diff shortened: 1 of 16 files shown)


andreab1997:
Hi @giacomomagni, thanks for this. Are you planning to solve here the issue of FONLL-B? Otherwise I can do it myself. 
Also, should I review this now or is it better to wait?

giacomomagni:
> Hi @giacomomagni, thanks for this. Are you planning to solve here the issue of FONLL-B? 

If you can solve it, I guess it's faster. 
If you can solve it here, so much the better.
Keep in mind that this PR might take a bit longer (as we need to write docs), 
so if it's just a commit better do it on master.

> Also, should I review this now or is it better to wait?

Yes, it works right now. The new commands are:

```
pineko fonll_ekos -c config.toml 800 INTEGXT3 INTEGXT8
pineko fonll_fks -c config.toml 800 INTEGXT3 INTEGXT8
```

Although multiple datasets are supported the ekos are running sequentially, 
so it might be faster to call the command twice instead than appending datasets in `fonll_ekos`. 
But this is up to the user.

andreab1997:
> > Hi @giacomomagni, thanks for this. Are you planning to solve here the issue of FONLL-B?
> 
> If you can solve it, I guess it's faster. If you can solve it here, so much the better. Keep in mind that this PR might take a bit longer (as we need to write docs), so if it's just a commit better do it on master.
> 
> > Also, should I review this now or is it better to wait?
> 
> Yes, it works right now. The new commands are:
> 
> ```
> pineko fonll_ekos -c config.toml 800 INTEGXT3 INTEGXT8
> pineko fonll_fks -c config.toml 800 INTEGXT3 INTEGXT8
> ```
> 
> Although multiple datasets are supported the ekos are running sequentially, so it might be faster to call the command twice instead than appending datasets in `fonll_ekos`. But this is up to the user.

Ok then let me take care of the `PTO` part and then I will review your part, after also the docs have been added.

docs/source/overview/running.rst
```diff
@@ -54,6 +57,79 @@ Note that you can also convolute a single grid with a single eko (obtaining a si
...
+
+Generating FONLL FK Tables
+--------------------------
+
+In order to generate FK tables with (numerical) |FONLL| the different flavor schemes
+need to be evolved separately and joined together in a single FK table only
+in the final step.
+
+There are 2 workflows possible, one in which all the steps are performed individually
+and one that is more automatize.
```
andreab1997, line 69:
```suggestion
and one that is more automatized.
```

docs/source/overview/running.rst
```diff
@@ -54,6 +57,79 @@ Note that you can also convolute a single grid with a single eko (obtaining a si
...
+  5. Generate the FK tables each of the 7 theories with the usual commands.
+
+  6. Combine the various FK tables into a single file, using the command::
+
+      pineko combine_fonll THEORY_ID DATASET --FFNS3 THEORY_ID00 --FFN03 THEORY_ID01 --FFNS4til THEORY_ID02 --FFNS4bar THEORY_ID03 --FFN04 THEORY_ID04 --FFNS5til THEORY_ID05 --FFNS5bar THEORY_ID06
+
+    where the first 3 theories are needed to perform |FONLL| on charm effects,
+    while the last 4 are needed to include also bottom effects.
+
+The automatize procedure can be faster, but it might be less flexible.
```
andreab1997, line 109:
```suggestion
The automatized procedure can be faster, but it might be less flexible.
```

andreab1997:
Please @giacomomagni when you can have a look at [28b8fe8](https://github.com/NNPDF/pineko/pull/153/commits/28b8fe8f3c3831ca5fe7cd68789e91113212cb64) and maybe try to compute an fktable to see that we get the correct result. 

Note that in [28b8fe8](https://github.com/NNPDF/pineko/pull/153/commits/28b8fe8f3c3831ca5fe7cd68789e91113212cb64) I was forced to add an entry to the tcard (`PTOEKO`) because otherwise there is no way to distinguish between, say, the 00 contribution of FONLL-B and of FONLL-C as their theory cards are exactly the same.

giacomomagni:
> Please @giacomomagni when you can have a look at [28b8fe8](https://github.com/NNPDF/pineko/pull/153/commits/28b8fe8f3c3831ca5fe7cd68789e91113212cb64) and maybe try to compute an fktable to see that we get the correct result.
> 

I checked that the automatized FONLL reproduces an older FKTable. 


> Note that in [28b8fe8](https://github.com/NNPDF/pineko/pull/153/commits/28b8fe8f3c3831ca5fe7cd68789e91113212cb64) I was forced to add an entry to the tcard (`PTOEKO`) because otherwise there is no way to distinguish between, say, the 00 contribution of FONLL-B and of FONLL-C as their theory cards are exactly the same.

Regarding your fix,  the only way to avoid this additional key would be to act at the level of the `opcard` creation, but it's a bit more convoluted, so as you prefer...

andreab1997:
> > Please @giacomomagni when you can have a look at [28b8fe8](https://github.com/NNPDF/pineko/pull/153/commits/28b8fe8f3c3831ca5fe7cd68789e91113212cb64) and maybe try to compute an fktable to see that we get the correct result.
> 
> I checked that the automatized FONLL reproduces an older FKTable.
> 
Ok good!
> > Note that in [28b8fe8](https://github.com/NNPDF/pineko/pull/153/commits/28b8fe8f3c3831ca5fe7cd68789e91113212cb64) I was forced to add an entry to the tcard (`PTOEKO`) because otherwise there is no way to distinguish between, say, the 00 contribution of FONLL-B and of FONLL-C as their theory cards are exactly the same.
> 
> Regarding your fix, the only way to avoid this additional key would be to act at the level of the `opcard` creation, but it's a bit more convoluted, as you need to detect the FONLL-B/D and the correct scheme to switch, so as you prefer...

How do you mean? Also, I am not sure why you are talking about the opcard that should be unchanged, right?

giacomomagni:
> How do you mean? Also, I am not sure why you are talking about the opcard that should be unchanged, right?

Not really, right now you are consuming `PTOEKO` exactly when computing the eko... 
So this operation can be done when writing the opcard.

andreab1997:
> > How do you mean? Also, I am not sure why you are talking about the opcard that should be unchanged, right?
> 
> Not really, if I understand correctly, right now you are consuming `PTOEKO` exactly when computing the eko... So this operation can be done when writing the opcard.

But I don't understand what you would put in the opcard exactly

andreab1997:
Can someone review this? I can review @giacomomagni's part but it does not make sense reviewing my own part. @felixhekhorn @RoyStegeman

felixhekhorn:
yes, sorry - I promised to take a look, but haven't had the time so far :see_no_evil: I'll try again tomorrow

docs/source/overview/running.rst
```diff
@@ -54,6 +57,79 @@ Note that you can also convolute a single grid with a single eko (obtaining a si
...
+      pineko fonll_tcards THEORY_ID
+
+    The different flavor patches are named following the convention:
+
+      * ``THEORY_ID00`` : |FFNS| :math:`n_f=3`
+      * ``THEORY_ID01`` : |FFN0| :math:`n_f=3`
+      * ``THEORY_ID02`` : massless component of |FFNS| :math:`n_f=4`
+      * ``THEORY_ID03`` : massive component of |FFNS| :math:`n_f=4`
+      * ``THEORY_ID05`` : massless component of |FFNS| :math:`n_f=5`
+      * ``THEORY_ID06`` : massive component of |FFNS| :math:`n_f=5`
```
RoyStegeman, line 87:
```suggestion
      * ``<THEORY_ID>00`` : |FFNS| :math:`n_f=3`
      * ``<THEORY_ID>01`` : |FFN0| :math:`n_f=3`
      * ``<THEORY_ID>02`` : massless component of |FFNS| :math:`n_f=4`
      * ``<THEORY_ID>03`` : massive component of |FFNS| :math:`n_f=4`
      * ``<THEORY_ID>04`` : |FFN0| :math:`n_f=4`
      * ``<THEORY_ID>05`` : massless component of |FFNS| :math:`n_f=5`
      * ``<THEORY_ID>06`` : massive component of |FFNS| :math:`n_f=5`
```

docs/source/overview/running.rst
```diff
@@ -54,6 +57,79 @@ Note that you can also convolute a single grid with a single eko (obtaining a si
...
+
+    where for |FFNS| :math:`n_f=4,5` massive and massless parts are splitted to
```
RoyStegeman, line 89:
```suggestion
    where for |FFNS| :math:`n_f=4,5` massive and massless parts are split to
```

docs/source/overview/running.rst
```diff
@@ -54,6 +57,79 @@ Note that you can also convolute a single grid with a single eko (obtaining a si
...
+    See above for the intermediate theories naming convention.
+
+  2. Generate the grids corrisponding to all the 7 theories with the external program.
+
+  3. Generate the ekos for only 3 theories and inherit the others running::
+
+      pineko fonll_ekos THEORY_ID DATASET1 DATASET2 ...
+
+    Note: this is usually an expensive operation as multiple ekos are run sequentially.
+    Depending on the resources that you have available it might be more convinient
```
RoyStegeman, line 125:
```suggestion
    Depending on the resources that you have available it might be more convenient
```

docs/source/overview/running.rst
```diff
@@ -54,6 +57,79 @@ Note that you can also convolute a single grid with a single eko (obtaining a si
...
+    where for |FFNS| :math:`n_f=4,5` massive and massless parts are splitted to
+    allow for a damping option.
+
+  2. Generate the grids corrisponding to all the 7 theories with the external program.
+
+  3. Generate the operator cards for each theory with the usual command.
+     Note that, in principle, only 3 ekos are needed, as there are only 3 different :math:`n_f` patches.
+     So you might speed up the procedure taking advantage of inherit-ekos.
+
+  4. Generate the ekos for each theory with the usual commands.
```
RoyStegeman, line 98: People who don't know the "usual commands" should also be able to follow this guide. Same goes for the point below

docs/source/overview/running.rst
```diff
@@ -54,6 +57,79 @@ Note that you can also convolute a single grid with a single eko (obtaining a si
...
+      * ``THEORY_ID01`` : |FFN0| :math:`n_f=3`
+      * ``THEORY_ID02`` : massless component of |FFNS| :math:`n_f=4`
+      * ``THEORY_ID03`` : massive component of |FFNS| :math:`n_f=4`
+      * ``THEORY_ID05`` : massless component of |FFNS| :math:`n_f=5`
+      * ``THEORY_ID06`` : massive component of |FFNS| :math:`n_f=5`
+
+    where for |FFNS| :math:`n_f=4,5` massive and massless parts are splitted to
+    allow for a damping option.
+
+  2. Generate the grids corrisponding to all the 7 theories with the external program.
```
RoyStegeman, line 92: Hmm this is a bit ugly. I understand wanting to keep things modular, but I think we should really have a step-by-step guide somewhere that explains how to use pinefarm/yadism together with pineko to create nFONLL FKtables. We will put it in the paper, but it would also be good to have it in a living document, i.e. docs, somewhere.

docs/source/overview/running.rst
```diff
@@ -54,6 +57,79 @@ Note that you can also convolute a single grid with a single eko (obtaining a si
...
+The automatized procedure can be faster, but it might be less flexible.
+In this case the workflows is:
+
+  1. Generate 7 theories for all the different flavor patches with the command::
+
+      pineko fonll_tcards THEORY_ID
+
+    See above for the intermediate theories naming convention.
+
+  2. Generate the grids corrisponding to all the 7 theories with the external program.
```
RoyStegeman, line 118:
```suggestion
  2. Generate the grids corresponding to all the 7 theories with the external program.
```

docs/source/overview/running.rst
```diff
@@ -54,6 +57,79 @@ Note that you can also convolute a single grid with a single eko (obtaining a si
...
+
+  3. Generate the ekos for only 3 theories and inherit the others running::
```
RoyStegeman, line 120:
```suggestion
  3. Generate the three ekos, one for each :math:`n_f`, and inherit the others running:
```

RoyStegeman (commented):
Some comments on the docs only. 

I think we should add docs under THEORY as well. To introduce the FONLL scheme

docs/source/overview/running.rst
giacomomagni: I meant the one listed above. Let me clarify

docs/source/overview/running.rst
giacomomagni: True, but this is why I wanted a single place where you can run everything...
As far as `pineko` is concerned we can add a link to `pinefarm` of `pineline` (although the page would be non existing yet.)

RoyStegeman:
I was indeed aware of the docs in yadism. I would say the should be moved here since yadism no longer does FONLL. However, since FONLL is precisely the motivation for having FFN0, it may be nice to then link in yadism to these pineko docs when the FONLL scheme is mentioned under FFN0.

felixhekhorn:
> I was indeed aware of the docs in yadism. I would say the should be moved here since yadism no longer does FONLL. However, since FONLL is precisely the motivation for having FFN0, it may be nice to then link in yadism to these pineko docs when the FONLL scheme is mentioned under FFN0.

I started to shuffle the docs around, but yes this needs more work. We need to move the whole [yadism stuff](https://yadism.readthedocs.io/en/latest/theory/fns.html#fonll) here to explain what we are actually talking about; we need to massage it such that we talk about an abstract observable, which for illustrative purpose is a DIS structure function.

We also need some business about the 2 mass problem ...

src/pineko/cli/fonll.py
```diff
@@ -69,83 +54,122 @@ def subcommand(
...
+    load_config(cfg)
+
     tcard = theory_card.load(theoryid)
     tcard_parent_path = theory_card.path(theoryid).parent
     if "FONLL" not in tcard["FNS"]:
         raise TheoryCardError("The theorycard does not correspond to an FONLL scheme.")
     fonll.dump_tcards(tcard, tcard_parent_path, theoryid)
+
+
+@command.command("fonll_ekos")
```
felixhekhorn, line 87: I'd say at this point we should definitely make a group (as for `theory`) and bind the subcommands there (thus also inverting `combine_fonll` to `fonll combine`)

andreab1997: Yes, I agree that at this point it is worth doing this.

giacomomagni: okay done, although this will break everyone scripts.

andreab1997:
> @felixhekhorn , @RoyStegeman and @andreab1997 maybe we can try to merge this PR and the improve the docs in a subsequent branch. Would that work for you?

I agree

felixhekhorn:
> @felixhekhorn , @RoyStegeman and @andreab1997 maybe we can try to merge this PR and the improve the docs in a subsequent branch. Would that work for you?

fine - let's just not delay it too long, in any case we need it for our paper

felixhekhorn (approved):

PR #159, opened 2024-03-14
Use the NNPDF data instead of yamldb

scarlehoff:
With this `pineko` will use `nnpdf` whenever the `nnpdf=True` key is set in the `[general]` part of the config.

Since when using `nnpdf`, the `yamldb` is not needed, I've dropped it from required keys.

At the moment this is not the default. I think it _should_ become default because the NNPDF no longer looks at the yamldb files, the data is the only source of truth, so the fktables should have the same names. But this is something that can always be done manually afterwards.


andreab1997 (commented):
Ok I see and it is fine for me. 

I see that this is a proof of concept, but are you planning to do the full job here? In other words, should I take care of this?

scarlehoff:
Nono, I will. I just wanted to know whether you were ok with the approach.

I hope to merge tomorrow this https://github.com/NNPDF/nnpdf/pull/1965 so I will finish it when that's in.

giacomomagni:
This would have the advantage of being able to read the kinematics from there, 
instead from the grid.

scarlehoff:
This is ready for review (and modifications) @andreab1997 

At the moment is using the whole of validphys since it uses the reader. It could only use the data through https://github.com/NNPDF/nnpdf/pull/1965 but since at the moment it is optional I think it is ok.
If you want to make it default you need to load the theory part of the metadata yourself from here (or move the parser to the data package in validphys).

I have not updated the dependencies because at the moment they are quire restrictive and don't want to go around updating eko/banana/python/etc. I've tried putting 3.9-3.13 and that broke banana.

I've updated only pineappl (which should be updated also in master).

alecandido:
> I have not updated the dependencies because at the moment they are quire restrictive and don't want to go around updating eko/banana/python/etc. I've tried putting 3.9-3.13 and that broke banana.

3.13 is definitely unsupported :P

scarlehoff:
It's <3.13

andreab1997:
> Details

Ok I will review this ASAP

giacomomagni (commented):
@scarlehoff  Right now the `yamldb` is used also in the `kfactor` module, can you change that one as well ?

scarlehoff:
Do you want me to do it before #161 ?

The changes needed for this are *very minimal* so it might be better to wait?

scarlehoff:
on second thoughts, should this even touch the kfactors part? We should _not_ need kfactors going forward, specially for the new data?

felixhekhorn:
> Do you want me to do it before #161 ?
> 
> The changes needed for this are _very minimal_ so it might be better to wait?

I think whoever is ready to be merged first can go ... we need to sort out the dependency problem here first: at the moment Github is just complaining that the lock file is outdated

> on second thoughts, should this even touch the kfactors part? We should _not_ need kfactors going forward, specially for the new data?

in the long run maybe yes, but surely we still need support for a while (e.g. for EW)

src/pineko/theory.py
```diff
@@ -22,6 +22,21 @@
 logger = logging.getLogger(__name__)
 
 
+def _load_grids_from_nnpdf(dataset_name):
+    """Load the list of fktables/grids necessary from a NNPDF dataset."""
+    # Hide the validphys import to avoid unnecessary failures
+    from validphys.commondataparser import EXT
+    from validphys.datafiles import legacy_to_new_map
+    from validphys.loader import Loader
```
felixhekhorn, line 30: will they eventually move to `nnpdf_data`? (since they did not yet in https://github.com/NNPDF/nnpdf/pull/2009 ? )

scarlehoff:
> in the long run maybe yes, but surely we still need support for a while (e.g. for EW)

Not from pineko

src/pineko/theory.py
scarlehoff: It's already merged to master.
However, at the moment this will rely on vp since I'd rather use the `Loader` that does various checks.

alecandido:
> in the long run maybe yes, but surely we still need support for a while (e.g. for EW)

I believe there are two point of views:
- Pineko (or whoever else) will burn kfactors in the grids, so downstream you shouldn't need kfactors any longer
- the kfactors themselves you will need forever: there will always be a calculation for which you don't have a generator, or you only have results that are not sufficiently exclusive - and there will be someone willing to include them anyhow, because a kfactor is a better approximation than nothing

scarlehoff:
Why is pylint being applied to the benchmark as well?

scarlehoff:
Ok, this is done as far as I'm concerned. Hopefully it even works. I've been using it for the perturbative charm theories.

The kfactor at the moment (current `main`) has the yamldb_path as a command line argument which is passed through (while this PR instead overrides the yamldb from the configuration).

I can modify if you want but that is just going to create unnecessary conflicts with #161.

src/pineko/theory.py
```diff
@@ -22,6 +22,32 @@
 logger = logging.getLogger(__name__)
 
 
+def read_grids_from_nnpdf(dataset_name):
```
andreab1997, line 25: Maybe this function can be moved outside theory. I would create an `utils.py` file and put it there

felixhekhorn:
> Why is pylint being applied to the benchmark as well?

mh? it is not: https://github.com/NNPDF/pineko/blob/812893e63d4d9762ca8b4fbd3b8ba5023ce74589/pyproject.toml#L71

problem is pylint needs to be update to be compatible with 3.12 - see also https://github.com/NNPDF/eko/pull/349

andreab1997 (approved):
Other than the trivial comment about the function location, this is ok for me

scarlehoff:
> > Why is pylint being applied to the benchmark as well?
> 
> mh? it is not: https://github.com/NNPDF/pineko/blob/812893e63d4d9762ca8b4fbd3b8ba5023ce74589/pyproject.toml#L71
> 
> problem is pylint needs to be update to be compatible with 3.12 - see also https://github.com/NNPDF/eko/pull/349

Not the benchmark folder but rather in the benchmark task.

src/pineko/theory.py
felixhekhorn: agreed (in case you foresee a growth there also `nnpdf_.py` could be a good name)

scarlehoff: I haven't done it here because it would be a mess with more stuff, but actually this should be `read_grids` and used everywhere, and the function internally should choose (according to the config) where to read from and how to create the grids.

At the moment fonll and theory do it in a different way (and kfactor in yet another way) so that would be a bigger change, but something to keep in mind.

scarlehoff:
But repeating the same check in several different workflows just reduces the amount of information you get out of the failures.

giacomomagni (commented):
Some minor comments: 

* Please update the docs [here](https://github.com/NNPDF/pineko/blob/main/docs/source/overview/prerequisites.rst), so we keep track of the changes.
* This [line](https://github.com/NNPDF/pineko/blob/d75ebe11e50c941f9c2221940787c74a055d4820/src/pineko/fonll.py#L125) has to be removed.

src/pineko/configs.py
```diff
@@ -59,8 +59,22 @@ def enhance_paths(configs_):
     configs_ : dict
         configuration
     """
+    required_keys = list(NEEDED_KEYS)
```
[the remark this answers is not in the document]
scarlehoff: I'm modifying the list so I need the copy. I can explicitly use `copy` if you prefer.

scarlehoff:
> This [line](https://github.com/NNPDF/pineko/blob/d75ebe11e50c941f9c2221940787c74a055d4820/src/pineko/fonll.py#L125) has to be removed.

Why? This is not using the theories from NNPDF.

giacomomagni:
> Why? This is not using the theories from NNPDF.

Sure, but it doesn't use the yamldb as well, it only looks at theory cards.

src/pineko/configs.py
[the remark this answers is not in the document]
giacomomagni: ah ok true, fine then

docs/source/overview/prerequisites.rst
```diff
@@ -30,10 +32,23 @@ This is a standard example:
...
+If a custom database of mappings is to be used, the path to the folder containing
+this files needs to be explicitly provided:
+
+::
+
+  [paths]
+  ymldb = "data/yamldb"
 
-You need all files of the *ymldb* [2]_ which define the mapping from datasets to FK tables.
+This *ymldb* [2]_ defines the mapping from datasets to FK tables.
```
felixhekhorn, line 51: can you please drop the footnote? as it is resolved by this PR

Issue #160, opened 2024-03-15
`kfactor` module maintenance 


giacomomagni:
we should maintain the `kfactor` module, right now it isn't documented, variables names are not consistent (`max_as` should look somenting as `order_to_update`) , and there are some hardcoded pieces which most likely will not work anymore with the new commondata: 

https://github.com/NNPDF/pineko/blob/06458035efaa06e5236acdd07cffde12afa2bfe6/src/pineko/kfactor.py#L401-L403

alecandido:
Btw: a huge rework, that could simplify it a lot, is now possible because of `merge` support in PineAPPL:
https://github.com/NNPDF/pineappl/pull/247

This was not available at the time this module was written (same for scale variations), and because of this many files were generated and later merged.
Ok, PineAPPL subgrids are not NumPy arrays, so not all operations are available. But indexing, scalar multiplications and a combination of sum/concatenation (through merge) is. This should be enough to reimplement it as:
- collect weights and paths (from here to there times this)
- apply the operation in memory

giacomomagni:
Yes indeed that we can also benefit. 
But also other pieces can be simplified:

https://github.com/NNPDF/pineko/blob/06458035efaa06e5236acdd07cffde12afa2bfe6/src/pineko/kfactor.py#L115-L121

for instace here `m` is subtracted and added back, so can simply be dropped by the function.

alecandido:
There are also entire functions to:
- extract the subgrid to a NumPy array
- scale the array
- set the scaled array as a new subgrid

https://github.com/NNPDF/pineko/blob/06458035efaa06e5236acdd07cffde12afa2bfe6/src/pineko/kfactor.py#L124-L145

While instead you could directly scale the subgrid itself
https://github.com/NNPDF/pineappl/blob/e0a45f6107befb63e25164c91244eb669b1bcfe1/pineappl_py/src/subgrid.rs#L202-L204

giacomomagni:
@alecandido do you remember what happen in case of merging with an overlapping order ?
It's simply overwritten?

alecandido:
Not really, the two subgrids will get merged.

The `Grid.merge()` algorithm is simple: if there is only one, pick that, if there are two, call `Subgrid.merge()`.

Merging subgrids has a different meaning for each subgrid kind, but the overall idea is an addition (e.g. for the n-tuple means concatenating the tuples' lists, and for Lagrange should be actual addition of the weights).

PR #161, opened 2024-03-15
`kfactor` maintenance and documentation 

giacomomagni:
Address #160. 

I'd like to: 

* [x] document the code
* [x] allow for nnpdf settings, ie drop the explicit yaml_db. 
* [ ] <strike>simplify the code, also making use of the new pineappl functions<strike>


andreab1997:
Should I review this immediately or wait until at least the tests pass?

giacomomagni:
> Should I review this immediately or wait until at least the tests pass?

Feel free to comment, I'll try to fix tests in the meantime

src/pineko/kfactor.py
```diff
@@ -112,11 +107,12 @@ def compute_scale_factor(
...
     """
-    max_as = to_construct_order[0] - m
+    max_as = to_construct_order[0]
     alpha_val = alphas.alphasQ2(Q2)
-    alpha_term = 1.0 / pow(alpha_val, max_as - (nec_order[0] - m))
+    alpha_term = 1.0 / pow(alpha_val, max_as - (nec_order[0]))
     k_term = central_k_factor[bin_index] - 1.0
-    if order_exists and (max_as - (nec_order[0] - m)) == 0:
+    # TODO: why do we need a double condition here ???
+    if order_exists and (max_as - (nec_order[0])) == 0:
```
giacomomagni, line 115: @andreab1997 can you remind me why we need both conditions?

andreab1997: You need to think at the equation. However the idea is that we always need the -1.0, but for the case in which the order we are constructing already exists (so we are replacing it) and, even in this case, only if nec_order is exactly this order (while nec_orders loop over all the orders in the grid anyway)

andreab1997: This is the reason we need both. You clearly see what is happening with a pdb

giacomomagni: I think I manage to solve it a bit easily. because whenever you will reach the max order you can get $1/as^0$, which in the end is 1. I need to look at it once more though

giacomomagni:
@andreab1997 the benchmark is not passing because `(3,0,0,0)` is already in the grid `ATLAS_TTB_FAKE`, 
is this correct?

andreab1997:
> @andreab1997 the benchmark is not passing because `(3,0,0,0)` is already in the grid `ATLAS_TTB_FAKE`, is this correct? How come this was not there before ?? Or maybe I'm messing up with the syntax and I should ask for `(4,0,0,0)` ?

I do not remember by hearth. Did you change the grids being tested?

giacomomagni:
> I do not remember by hearth. Did you change the grids being tested?

No I did not...
So the grids contains: 
```
[array([2, 0, 0, 0]), array([3, 0, 0, 0]), array([3, 0, 1, 0]), array([3, 0, 0, 1])]
```

andreab1997:
> > I do not remember by hearth. Did you change the grids being tested?
> 
> No I did not... So the grids contains:
> 
> ```
> [array([2, 0, 0, 0]), array([3, 0, 0, 0]), array([3, 0, 1, 0]), array([3, 0, 0, 1])]
> ```

Okay then yes, of course in this case the benchmark was meant to compute an additional order, the (4,000). Instead you are using it to update (3,000) which however is already there, so it fails. So yes, you should update the (4,0,0,0). I think now that the name `order_to_update` is not that good, because in this case you are actually adding an order, not updating it. The update part is only when you use `order_exist`, but in general you want to create new orders, not updating them. I guess this was the reason of the confusion

giacomomagni:
I left 2 todos which I don't know if or how we want to solve:

https://github.com/NNPDF/pineko/blob/64e252d751439e86408e3ce70718c2075499b159/src/pineko/scale_variations.py#L180

https://github.com/NNPDF/pineko/blob/64e252d751439e86408e3ce70718c2075499b159/src/pineko/kfactor.py#L346-L349

But apart from that the rest should be a bit nicer now, if you want please review.

andreab1997:
> I left 2 todos which I don't know if or how we want to solve:
> 
> https://github.com/NNPDF/pineko/blob/64e252d751439e86408e3ce70718c2075499b159/src/pineko/scale_variations.py#L180
> 
> https://github.com/NNPDF/pineko/blob/64e252d751439e86408e3ce70718c2075499b159/src/pineko/kfactor.py#L346-L349
> 
> But apart from that the rest should be a bit nicer now, if you want please review.

Ok thanks for this, I will review this ASAP. 

Regarding the two TODOs:

> https://github.com/NNPDF/pineko/blob/64e252d751439e86408e3ce70718c2075499b159/src/pineko/scale_variations.py#L180
> 

This has already been simplified a lot in the past (it was even worst). Personally I would not know how to further simplify it, other than dividing it in unit functions (which is still not trivial).

> 
> https://github.com/NNPDF/pineko/blob/64e252d751439e86408e3ce70718c2075499b159/src/pineko/kfactor.py#L346-L349
> 

This might be a good idea but the problem is that, while for QCD kfactor it is rather trivial to know what they do, other kfactors can be defined in an arbitrary way which makes particulary difficult to generalize the equations that we implement here. We can try to assume a general behaviour and put some options for the rest but I believe that it easily become ugly

alecandido:
This part is just creating an empty grid like the other, replacing orders:
https://github.com/NNPDF/pineko/blob/64e252d751439e86408e3ce70718c2075499b159/src/pineko/scale_variations.py#L181-L192

and here you're iterating to filter the subgrids to export:
https://github.com/NNPDF/pineko/blob/64e252d751439e86408e3ce70718c2075499b159/src/pineko/scale_variations.py#L193-L203
This could be possible with a single iteration, by making a clever usage of the `IndexedIter`.

Moreover, constructing grids from filtering existing ones could be a feature for PineAPPL itself (if you filter with indices, instead of callables, there is no problem of exposing in the non-Rust APIs).

This is just a function of `ori_grid` and nothing else:
https://github.com/NNPDF/pineko/blob/64e252d751439e86408e3ce70718c2075499b159/src/pineko/scale_variations.py#L204-L215

This is also a separate piece:
https://github.com/NNPDF/pineko/blob/64e252d751439e86408e3ce70718c2075499b159/src/pineko/scale_variations.py#L218-L220

In any case, this function altogether is just the filtering one that I mentioned, and then taking care of propagating additional pieces of information. If you could delete subgrids, just a full copy and then drop would be simpler.
The nicest way I see to achieve the goal would be to have a `GridView` object, such that you could obtain a consistent filtered view without copying subgrids, and then copy-on-write (or just provide a way to dump the view).

However, most of these improvements would require to do something inside PineAPPL itself.

giacomomagni:
> This has already been simplified a lot in the past (it was even worst). Personally I would not know how to further simplify it, other than dividing it in unit functions (which is still not trivial).
>
Okay if you can take a look, please let me know otherwise we leave it as it is.

> This might be a good idea but the problem is that, while for QCD kfactor it is rather trivial to know what they do, other kfactors can be defined in an arbitrary way which makes particulary difficult to generalize the equations that we implement here. We can try to assume a general behaviour and put some options for the rest but I believe that it easily become ugly

We might want to allow for `NRM` (ie. rescale all the orders) and `QED` (same as we do now for the other coupling) with are not so difficult, but this we can do it in another PR. Here I'd prefer just to drop the hardcoded `QCD` if you agree.

docs/source/theory/kfactors.rst
```diff
@@ -4,8 +4,16 @@ K-Factors
 Another useful tool that `pineko` includes is ``pineko kfactor`` which allows the embedding of a kfactor
 as a proper order in a grid. The usage is the following::
 
-  pineko kfactor GRIDS_FOLDER KFACTOR_FOLDER YAMLDB_PATH TARGET_FOLDER MAX_AS ORDER_EXISTS
+  pineko kfactor GRIDS_FOLDER KFACTOR_FOLDER YAMLDB_FILE TARGET_FOLDER PTO_TO_UPDATE [--order_exists]
 
 where ``GRIDS_FOLDER`` is the folder containing the grids to update, ``KFACTOR_FOLDER`` is the folder
```
felixhekhorn, line 9: Consider turning this into a proper list - at the moment it is not really clear

docs/source/theory/kfactors.rst
```diff
@@ -4,8 +4,16 @@ K-Factors
...
-containing the kfactor files and ``YAMLDB_PATH`` is the path to the yamldb file of the requested dataset.
+containing the kfactor files and ``YAMLDB_FILE`` is the path to the yamldb file of the requested dataset.
 The other inputs have already been described in the previous section.
```
felixhekhorn, line 11: I don't think we use `TARGET_FOLDER` any where else, do we? and even if we do - let's repeat it here for the sake of having a single place

src/pineko/kfactor.py
```diff
@@ -145,52 +127,51 @@ def scale_subgrid(extracted_subgrid, scales_array):
     return scaled_subgrid
 
 
-def compute_orders_map(m, max_as, min_al, order_exists):
-    """Compute a dictionary with all the necessary orders to compute the requested order.
+def is_already_in(to_check, list_orders):
```
felixhekhorn, line 130: maybe we can add a "no_logs" or similar? because it is not just plain search

src/pineko/cli/kfactor.py
```diff
@@ -11,28 +11,28 @@
 @command.command("kfactor")
 @click.argument("grids_folder", type=click.Path(exists=True))
 @click.argument("kfactor_folder", type=click.Path(exists=True))
-@click.argument("yamldb_path", type=click.Path(exists=True))
+@click.argument("yamldb_file", type=click.Path(exists=True))
 @click.argument("target_folder", type=click.Path(exists=True))
-@click.argument("max_as", type=int)
-@click.argument("order_exists", type=bool)
+@click.argument("pto_to_update", type=int)
+@click.option("--order_exists", is_flag=True, help="Owerwrite an existing order.")
```
felixhekhorn, line 17:
```suggestion
@click.option("--overwrite", is_flag=True, help="Owerwrite an existing order.")
```
maybe? just as we do for `theory`

src/pineko/kfactor.py
```diff
@@ -9,15 +9,11 @@
...
 
 
-def factgrid(subgrid):
-    """Return the array of the factorization scales squared from a subgrid."""
-    return np.array([mu2.fac for mu2 in subgrid.mu2_grid()])
-
-
 def rengrid(subgrid):
     """Return the array of the renormalization scales squared from a subgrid."""
     return np.array([mu2.ren for mu2 in subgrid.mu2_grid()])
```
felixhekhorn, line 19: I would inline this - there is a single use and the function is almost trivial

src/pineko/kfactor.py
```diff
@@ -52,72 +48,58 @@ def read_kfactor(kfactor_path):
 
 def construct_scales_array(
```
felixhekhorn, line 49: again I would inline - using list comprehension this is a one-liner again

src/pineko/kfactor.py
```diff
@@ -52,72 +48,58 @@ def read_kfactor(kfactor_path):
...
-    Q2,
+    order,
+    order_to_update,
+    mu2,
     central_k_factor,
     bin_index,
     alphas,
-    order_exists,
 ):
     """Compute the factor to be multiplied to the given nec_order.
```
felixhekhorn, line 81: Please update docs (here and below)

src/pineko/kfactor.py
```diff
@@ -52,145 +48,133 @@ def read_kfactor(kfactor_path):
...
-    if order_exists and (max_as - (nec_order[0] - m)) == 0:
-        k_term = central_k_factor[bin_index]
     return k_term * alpha_term
 
 
 def scale_subgrid(extracted_subgrid, scales_array):
-    """Rescales the array contained in the subgrid using scales_array and returns a new subgrid constructed with the scaled array."""
+    """Rescales the array contained in the subgrid using scales_array."""
     original_array = extracted_subgrid.to_array3()
     if len(original_array) != len(scales_array):
```
felixhekhorn, line 109: shouldn't we check on `.shape` here?

src/pineko/kfactor.py
```diff
@@ -209,148 +193,108 @@ def create_singlegridonly(
...
+    Parameters
+    ----------
+    central_kfactor : np.ndarray
+        kfactors to apply
+    alphas : lhapdf.AlphaS
+        alphas
+    grid : pineappl.grid
+        loaded grid
+    pto_to_update : int
+        perturbative order to update: 1 = LO, 2 = NLO ...
```
felixhekhorn, line 215: Wait a second this is not NNPDF default: LO <=> PTO=0

src/pineko/kfactor.py
```diff
@@ -209,148 +193,108 @@ def create_singlegridonly(
...
+    for i, as_order in enumerate(orders_list):
+        order_grid = construct_new_order(
+            grid, as_order, order_to_update, central_kfactor, alphas
+        )
+        if i == 0:
+            new_order_grid = order_grid
+        else:
+            new_order_grid.merge(order_grid)
+
+    new_grid = grid
```
felixhekhorn, line 262: This statement seems useless ...

src/pineko/kfactor.py
```diff
@@ -209,148 +193,108 @@ def create_singlegridonly(
...
-    # Merging all together
-    scale_variations.merge_grids(
-        gridpath=grid_path,
-        grid_list_path=grids_paths,
-        target_path=target_folder,
-        nec_orders=nec_orders,
-        order_exists=order_exists,
-    )
+    # the actual alpha_s order to update
+    order_to_update = pto_to_update + min_as - 1
```
felixhekhorn, line 234: - Maybe split this function in 2: one where you do the order algebra and one where you do the grid rescaling;
- you also note that later you redefine `order_to_update = (order_to_update, min_al, 0, 0)`, which instead should be done here as the next line is already using that one

src/pineko/kfactor.py
```diff
@@ -52,145 +48,133 @@ def read_kfactor(kfactor_path):
...
 
 def scale_subgrid(extracted_subgrid, scales_array):
-    """Rescales the array contained in the subgrid using scales_array and returns a new subgrid constructed with the scaled array."""
+    """Rescales the array contained in the subgrid using scales_array."""
     original_array = extracted_subgrid.to_array3()
     if len(original_array) != len(scales_array):
         raise ValueError("The original and the scales arrays have different shapes.")
+    # construct subgrid
     scaled_array = []
     for scale_value, arr_to_scale in zip(scales_array, original_array):
```
felixhekhorn, line 113: actually, can't we collapse all those loops with a single [`np.multiply`](https://numpy.org/doc/stable/reference/generated/numpy.multiply.html#numpy.multiply)

src/pineko/kfactor.py
```diff
@@ -209,148 +193,108 @@ def create_singlegridonly(
...
+    if is_in:
+        new_grid = scale_variations.construct_and_dump_order_exists_grid(
+            grid, order_to_update
+        )
+    # merge the updated order with the original one.
+    new_grid.merge(new_order_grid)
+    new_grid.write_lz4(target_grid_path)
 
 
 def filter_k_factors(pigrid, centrals_kfactor):
```
felixhekhorn, line 273:
```suggestion
def filter_k_factors(grid, central_kfactors):
```
? (+ all the implications ...)

src/pineko/kfactor.py
```diff
@@ -209,148 +193,108 @@ def create_singlegridonly(
...
-    """Filter the centrals k-factors according to their lenght compared to the number of bins of the grid."""
+    """Filter the centrals k-factors according to their length compared to the number of bins of the grid."""
     centrals_kfactor_filtered = np.array([])
     if pigrid.bins() == len(centrals_kfactor):
-        rich.print("[orange] The number of bins match the lenght of the k-factor.")
+        rich.print("[orange] The number of bins match the length of the k-factor.")
```
felixhekhorn, line 277:
```suggestion
        rich.print("[green] The number of bins match the length of the k-factor.")
```
this is good, right?

src/pineko/kfactor.py
```diff
@@ -209,148 +193,108 @@ def create_singlegridonly(
...
         centrals_kfactor_filtered = centrals_kfactor
```
felixhekhorn, line 278:
```suggestion
        return centrals_kfactor
```

src/pineko/kfactor.py
```diff
@@ -209,148 +193,108 @@ def create_singlegridonly(
...
+        )
+    # merge the updated order with the original one.
+    new_grid.merge(new_order_grid)
+    new_grid.write_lz4(target_grid_path)
 
 
 def filter_k_factors(pigrid, centrals_kfactor):
-    """Filter the centrals k-factors according to their lenght compared to the number of bins of the grid."""
+    """Filter the centrals k-factors according to their length compared to the number of bins of the grid."""
     centrals_kfactor_filtered = np.array([])
```
felixhekhorn, line 275: This variable is not necessary: use early exit or plain python + `np` (e.g. `np.array([100]*7)`)

alecandido: Or just NumPy, e.g. `np.ones(7) * 100` :)

docs/source/theory/kfactors.rst
giacomomagni: Mmm do you mean iterating over datasets?

src/pineko/cli/kfactor.py
giacomomagni: I though the same, but it would be more confusing. `overwrite` is done to overwrite an operation that has been done with the same pineko command, while here the situation is slightly different, you want to overwrite and order that is already present.

src/pineko/kfactor.py
giacomomagni: Sure this is possible.

src/pineko/kfactor.py
giacomomagni: Indeed that is why I started this PR 🙃

src/pineko/kfactor.py
giacomomagni: you mean using  `new_grid` instead of `grid` ? I think it's a bit clearer this way.

src/pineko/kfactor.py
giacomomagni: done in [3163be4](https://github.com/NNPDF/pineko/pull/161/commits/3163be4955b25bc48118002614d3f799284fa3f1)

src/pineko/kfactor.py
giacomomagni: don in [3163be4](https://github.com/NNPDF/pineko/pull/161/commits/3163be4955b25bc48118002614d3f799284fa3f1)

src/pineko/kfactor.py
giacomomagni: done in [3163be4](https://github.com/NNPDF/pineko/pull/161/commits/3163be4955b25bc48118002614d3f799284fa3f1)

src/pineko/kfactor.py
giacomomagni: I should have not changed anything here....

docs/source/theory/kfactors.rst
giacomomagni: done in [3163be4](https://github.com/NNPDF/pineko/pull/161/commits/3163be4955b25bc48118002614d3f799284fa3f1)

src/pineko/kfactor.py
giacomomagni: I actually don't understand how this case can work:

https://github.com/NNPDF/pineko/blob/06458035efaa06e5236acdd07cffde12afa2bfe6/src/pineko/kfactor.py#L336-L342

but fine...

src/pineko/kfactor.py
giacomomagni: not sure about it.

src/pineko/cli/kfactor.py
andreab1997, line 17:
```suggestion
@click.option("--order_exists", is_flag=True, help="Overwrite an existing order.")
```

andreab1997:
I still have to have a look at the two most important files, but I'll wait until @felixhekhorn is happy with them

docs/source/theory/kfactors.rst
felixhekhorn: nono, like
```rst
where

- ``GRIDS_FOLDER`` is the folder
- bla
- blub
```

src/pineko/kfactor.py
andreab1997: you mean, instead of using `len` in line 109?

src/pineko/kfactor.py
felixhekhorn:
```suggestion
def is_already_in_no_logs(to_check, list_orders):
```
or something like - just to reflect more the actual content

src/pineko/kfactor.py
andreab1997: Indeed this is not NNPDF conventions for which PTO=2 is NNLO, for example. However this is EKO convention and thus, in general, pineline convention. Since `pineko` is part of the pineline it seemed sensible to me to use the `pineline` convention, even if here we are using NNPDF kfactors (note in fact that we are not limited to NNPDF kfactor, but we can use any kfactor written in the same format).

src/pineko/kfactor.py
andreab1997: What do you mean?

src/pineko/kfactor.py
felixhekhorn: > you mean, instead of using `len` in line 109?

yes - and [as said below](https://github.com/NNPDF/pineko/pull/161#discussion_r1528606957) we should collapse this with just `np.multiply`

src/pineko/kfactor.py
andreab1997: In fact you can see that in line 339 I am checking that indeed the kfactor is a constant

src/pineko/kfactor.py
felixhekhorn: :scream: I mean a name is just a name, but then let's NOT use the letter combination `pto` and let's call it `order_to_update` e.g. else it is really confusing

src/pineko/kfactor.py
felixhekhorn: > Yes but the green is for when the job is done,

fine, then let's drop the color - orange sounds too much like a warning to me

src/pineko/kfactor.py
felixhekhorn: see [55e015d](https://github.com/NNPDF/pineko/pull/161/commits/55e015d7f60753c0982d059cbfc6fb4c32c37aab)

src/pineko/kfactor.py
felixhekhorn: see [55e015d](https://github.com/NNPDF/pineko/pull/161/commits/55e015d7f60753c0982d059cbfc6fb4c32c37aab)

src/pineko/kfactor.py
felixhekhorn: done in [55e015d](https://github.com/NNPDF/pineko/pull/161/commits/55e015d7f60753c0982d059cbfc6fb4c32c37aab)

src/pineko/kfactor.py
felixhekhorn: see [55e015d](https://github.com/NNPDF/pineko/pull/161/commits/55e015d7f60753c0982d059cbfc6fb4c32c37aab)

felixhekhorn:
https://github.com/NNPDF/pineko/pull/161/commits/82e4657f552d832b66e4ae391c24634e4382b54b was necessary because here we still have py3.8 in game (will be dropped by #159 )

src/pineko/kfactor.py
```diff
@@ -209,168 +168,129 @@ def create_singlegridonly(
...
+    # this is not a problem because in those cases either all kfactor values are the
+    # same (thus there is no doubt about whether we have the correct one) or the
+    # non-exisiting kfactor would be multiplied by bins corresponding to all '0' in the
+    # grid.
+    # Let's check if we are in the first or second case
+    if len(np.unique(central_kfactors)) == 1:
+        # In this case I just need to add more elements to the kfactor
+        return np.full(grid.bins(), central_kfactors[0])
+    # In this case this means that the missing entries will
+    # multiply zero subgrids so we can just add 0s
```
felixhekhorn, line 285: This is not doing what is saying: it is not *adding* - it is just 0s ... please make it consistent (one way or another)

src/pineko/kfactor.py
```diff
@@ -9,20 +9,11 @@
...
-    """Return the array of the factorization scales squared from a subgrid."""
-    return np.array([mu2.fac for mu2 in subgrid.mu2_grid()])
-
-
-def rengrid(subgrid):
-    """Return the array of the renormalization scales squared from a subgrid."""
-    return np.array([mu2.ren for mu2 in subgrid.mu2_grid()])
-
-
 def read_kfactor(kfactor_path):
```
felixhekhorn, line 17:
```suggestion
def read_from_file(path):
```

src/pineko/kfactor.py
```diff
@@ -209,168 +168,129 @@ def create_singlegridonly(
...
+    # Let's check if we are in the first or second case
+    if len(np.unique(central_kfactors)) == 1:
+        # In this case I just need to add more elements to the kfactor
+        return np.full(grid.bins(), central_kfactors[0])
+    # In this case this means that the missing entries will
+    # multiply zero subgrids so we can just add 0s
+    return np.full(grid.bins(), 0)
 
 
 def compute_k_factor_grid(
```
felixhekhorn, line 289:
```suggestion
def apply_to_dataset(
```

src/pineko/kfactor.py
```diff
@@ -384,35 +304,41 @@ def compute_k_factor_grid(
...
+            current_grid = pineappl.grid.Grid.read(grids_folder / grid_name)
+
+            central_kfactor, pdf_set = read_kfactor(cfac_path)
+            central_kfactor_filtered = to_list(current_grid, central_kfactor)
             alphas = lhapdf.mkAlphaS(pdf_set)
-            grid_path = grids_folder / (f"{grid}.pineappl.lz4")
-            pigrid = pineappl.grid.Grid.read(grid_path)
-            centrals_kfactor_filtered = filter_k_factors(pigrid, centrals_kfactor)
+
             do_it(
```
felixhekhorn, line 337:
```suggestion
            apply_to_grid(
```

felixhekhorn (commented):
no, need to tell we're dealing with k-factors in the `kfactor` module

src/pineko/kfactor.py
felixhekhorn: well, you're more experienced than myself here, so do as you feel more comfortable. However, please add a fool-proof documentation somewhere! e.g. add this discussion together with a big warning in the docstring

docs/source/theory/kfactors.rst
giacomomagni: done in 5a31fc0674d7fbc004a851f5635af0432bd0ab86

src/pineko/kfactor.py
giacomomagni: done in 5a31fc0674d7fbc004a851f5635af0432bd0ab86

src/pineko/kfactor.py
felixhekhorn, line 1: Please choose your favourite spelling of "[K|k][-]factor[s]" and then let's stick to that in all documentation

docs/source/theory/kfactors.rst
```diff
@@ -4,8 +4,21 @@ K-Factors
...
 
-  pineko kfactor GRIDS_FOLDER KFACTOR_FOLDER YAMLDB_PATH TARGET_FOLDER MAX_AS ORDER_EXISTS
+  pineko kfactor GRIDS_FOLDER KFACTOR_FOLDER YAMLDB_FILE TARGET_FOLDER PTO_TO_UPDATE [--order_exists]
 
-where ``GRIDS_FOLDER`` is the folder containing the grids to update, ``KFACTOR_FOLDER`` is the folder
-containing the kfactor files and ``YAMLDB_PATH`` is the path to the yamldb file of the requested dataset.
-The other inputs have already been described in the previous section.
+where
+
+where
```
felixhekhorn, line 11:
```suggestion
where
```

src/pineko/kfactor.py
giacomomagni: Indeed it was not consistent before, I'd stick to `kfactors`.

docs/source/theory/kfactors.rst
```diff
@@ -4,8 +4,21 @@ K-Factors
...
+- ``YAMLDB_FILE`` is the path to the yamldb file of the requested dataset,
+- ``TARGET_FOLDER`` is the folder where the new updated grids is going to be saved,
+- ``PTO_TO_UPDATE`` is the :math:`\alpha_s` perturbative order to update or create,
+  with the convention that ``LO=1``, ``NLO=2`` and so on, irrespectively to the powers of :math:`\alpha_s`.
+- ``--order_exists`` is a flag used to apply the kfactor to the specified perturbative order, insead of the next.
+
+Note that only pure QCD kfactors are taken into account.
+For example to add the NNLO in a grid containing at most NLO one has to select ``PTO_TO_UPDATE=2``;
+nn the other hand to reweight the NNLO present in a grid with a kfactor,
+one should do ``PTO_TO_UPDATE=2 --order_exists``.
```
felixhekhorn, line 22:
```suggestion
one should do ``PTO_TO_UPDATE=2 --order_exists``.
Note also that this differs from the conventions by the NNPDF collaboration, but is consistent with the pineappl conventions.
```
or similar

src/pineko/kfactor.py
giacomomagni: Idk anything of this function, you have just modify it...

felixhekhorn: fine fine, but let's do what this PR is promising - i.e. improve the `kfactor` module; and the main author of that  is @andreab1997 so you can delegate and he should take actively part in this PR :innocent:

src/pineko/kfactor.py
```diff
@@ -209,172 +168,133 @@ def create_singlegridonly(
...
+    return np.full(grid.bins(), 0)
+
+
+def apply_to_dataset(
     grids_folder,
     kfactor_folder,
     yamldb_path,
-    max_as,
+    pto_to_update,
     target_folder=None,
```
felixhekhorn, line 294:
```suggestion
    target_folder,
```
I don't think `None` works - specifically `target_folder / grid_name` does not

giacomomagni: Yes, let's drop it. I believe it' better to have it always specified and not overwrite the previous grid.

src/pineko/kfactor.py
giacomomagni: done in 4665cadc6dfd9f262b5b490504159becbe3656fa

src/pineko/kfactor.py
giacomomagni: done in b62249b9271626b2ab78f71235e3939676758df5

src/pineko/kfactor.py
giacomomagni: done it b62249b9271626b2ab78f71235e3939676758df5

src/pineko/kfactor.py
giacomomagni: done in [25be7c4](https://github.com/NNPDF/pineko/pull/161/commits/25be7c4c1df7fab88905c5c2fc788acb39dd3d80)

docs/source/theory/kfactors.rst
giacomomagni: done in [21572f8](https://github.com/NNPDF/pineko/pull/161/commits/21572f8492360fe951eb0920810fb1a218430d41)

src/pineko/kfactor.py
giacomomagni: done in [21572f8](https://github.com/NNPDF/pineko/pull/161/commits/21572f8492360fe951eb0920810fb1a218430d41)

src/pineko/kfactor.py
felixhekhorn: actually this is dangerous since it may lead to unpredictable behaviour: in python arguments are always passed by reference and hence depending on `is_in`, which is a complicated expression, you may or may not modify the argument you passed in. If `is_in`  is true you are modifying in-place, if not you're not (since `construct_and_dump_order_exists_grid`* returns a new grid). I think grids are now cloneable, so we should do that here - but please double check this (cc @alecandido ).

PS: * that name is awful, please find a better one

giacomomagni:
Give #159 I'd remove the `yamldb` also from here and just pass the single grids and kfactor files one by one with the CLI.

giacomomagni:
At this point also this PR is waiting for #181 to be closed and pineappl dependency to be updated.

giacomomagni:
Can we close this, before we forget what we have done ?

src/pineko/kfactor.py
felixhekhorn: postponed to #191

src/pineko/kfactor.py
felixhekhorn: postponed to #191

felixhekhorn:
I opened #191 and let's not close the discussion here.

Then let's merge (now) - this is getting really old

giacomomagni:
@andreab1997  are u okay?

PR #163, opened 2024-03-25
Update fonll docstr and json loading

scarlehoff:
While updating this line https://github.com/NNPDF/pineko/blob/d75ebe11e50c941f9c2221940787c74a055d4820/src/pineko/fonll.py#L125

I tried to run it just because and noticed that not all fktables have the theories saved as valid json. In particular some of the HERA fktables in 701XX sometimes have single quotes instead of double so I've added a "bypass" for that as well (and a few boolean values are also wrong). Not sure why it happened.

### DIFF
--- a/src/pineko/fonll.py
+++ b/src/pineko/fonll.py
@@ -45,6 +45,20 @@ class InconsistentInputsError(Exception):
     """Raised if the inputs are not consistent with FONLL."""
 
 
+def _json_theory_read(theory_info):
+    """Read the theory info from a pineappl grid.
+
+    Theory information within pineappl grids might not be saved in a consistent
+    json-compatible way. This functions corrects some found problems.
+    """
+    try:
+        ret = json.loads(theory_info)
+    except json.decoder.JSONDecodeError:
+        # Some grids have the theories saved with incompatible formats
+        ret = json.loads(theory_info.replace("'", '"').replace("True", "true"))
+    return ret
+
+
 class FONLLInfo:
     """Class containing all the information for FONLL predictions."""
 
@@ -98,7 +112,11 @@ class FONLLInfo:
     @property
     def theorycard_no_fns_pto(self):
         """Return the common theory info between the different FONLL FK tables."""
-        theorycards = [json.loads(self.fks[p].key_values()["theory"]) for p in self.fks]
+        theorycards = []
+        for pinefk in self.fks.values():
+            thinfo = pinefk.key_values()["theory"]
+            theorycards.append(_json_theory_read(thinfo))
+
         # Only these should differ
         for card in theorycards:
             del card["FNS"]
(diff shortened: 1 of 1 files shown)


giacomomagni:
Thanks for spotting it. Which are the keys with the wrong syntax? 
Maybe we should correct this also in `pinefarm` or `yadism`

src/pineko/fonll.py
```diff
@@ -45,6 +45,20 @@ class InconsistentInputsError(Exception):
...
+
+    Theory information within pineappl grids might not be saved in a consistent
+    json-compatible way. This functions corrects some found problems.
+    """
+    try:
+        ret = json.loads(theory_info)
+    except json.decoder.JSONDecodeError:
+        # Some grids have the theories saved with incompatible formats
+        ret = json.loads(theory_info.replace("'", '"').replace("True", "true"))
+    return ret
```
alecandido, line 59:
```suggestion
        return json.loads(theory_info)
    except json.decoder.JSONDecodeError:
        # Some grids have the theories saved with incompatible formats
        return json.loads(theory_info.replace("'", '"').replace("True", "true"))
```

scarlehoff: why?

(in this cases I prefer to leave the return outside so that it is clear that they should be the same object, but I don't know whether there's a reason to have it separated)

scarlehoff:
> Which are the keys with the wrong syntax?

There's a few grids where the theory key uses `'` instead of `"` (which is invalid json)

For the `True` instead of `true`, I don't remember.

felixhekhorn:
> Thanks for spotting it. Which are the keys with the wrong syntax? Maybe we should correct this also in `pinefarm` or `yadism`

I was sure that `pinefarm` writes the theory as meta data, but I can't find it https://github.com/NNPDF/pinefarm/blob/39abf325d3ab8ec227f7c296935e55f966184434/src/pinefarm/external/interface.py#L120 what am I missing? (and it should be fixed in `pinefarm`)

src/pineko/fonll.py
alecandido: Early return: as soon as you finish, return :)

It just makes the code more readable (slightly, in this case).

The output of a function is only one, though conditional. So, whatever you return is the same object (i.e. `ret = ...; return ret` is as good as `return ...` at telling that they are the same).

scarlehoff:
> I was sure that pinefarm writes the theory as meta data,

@felixhekhorn I was curious and checked that yadism actually writes this using `json.dumps` so the error was, in theory, impossible.

Indeed, the error is generated by pineko but I don't understand how and why it happens only for some datasets and not others.
If you are curious, HERACOMB_SIGMARED_B in theory 701 has this problem.

I'll open an issue about this.

Issue #164, opened 2024-03-25
Issue in the generation of the metadata of the fktables


scarlehoff:
When pineko generates an fktable coming from the copnvolution of a grid and an eko, it can happen that the theory metadata gets saved in a json-incompatible format.

This is generally not a problem (since this metadata is never read in practice), however, the `fonll` module now looks at this key which raises an error at `load`
This PR fixes the error when reading: https://github.com/NNPDF/pineko/pull/163 the json data.
However we should ideally fix the problem at creation time, but I'm not sure where it is (honestly, I'm not even sure it is pineko and not something happening when the string goes through pineappl)

alecandido:
I'm not sure if this is a problem only with FONLL combined FK tables, but this seems fishy:
https://github.com/NNPDF/pineko/blob/d75ebe11e50c941f9c2221940787c74a055d4820/src/pineko/fonll.py#L135
I would have expected `json.dumps()`, not `str()`.

scarlehoff:
Yes, I'm pretty sure this one is wrong.

```
>>> str({"test": "oh, no"})
"{'test': 'oh, no'}"
>>> json.dumps({"test": "oh, no"})
'{"test": "oh, no"}'
```

 However it must happen somewhere else (searching for either `str(` or `set_key_value` I could not find it) since this one happens after the load (so the bug has already happened). That's why I'm not sure it is pineko itself.

alecandido:
In pinefarm there is this:
https://github.com/NNPDF/pinefarm/blob/39abf325d3ab8ec227f7c296935e55f966184434/src/pinefarm/tools.py#L212-L214
But it propagates whatever you already have (starting from JSON, you will obtain JSON...).

I also audited the rest of pinefarm, and it doesn't seem to have any chance of introducing something like that.

And yadism is correctly using `json.dumps()`:
https://github.com/NNPDF/yadism/blob/bde1c764669f23e78898da06e11a6185ad098736/src/yadbox/export.py#L90

However, since you applied it only to the FONLL part, could it be that this has been done in some of the untracked scripts?
Not sure, it would be helpful to know which are the corrupted grids...

alecandido:
You could run a scan, with a simple Python or shell script (just use `jq '.'` and check for the error code).

However, while fixing the error is for sure relevant, I'm not so sure if we could still attribute it in any way to Pineko...
(that is, this could now be an issue of some untracked script, and not really a Pineko one)

scarlehoff:
[...]

PR #166, opened 2024-03-26
Use json for theory card

felixhekhorn:
As @alecandido [said](https://github.com/NNPDF/pineko/issues/164#issuecomment-2018766175) in #164, we should not use `str` to write the theory back to the grid but `json.dumps`. The present fix may or may not be the whole problem discussed in #164

### DIFF
--- a/src/pineko/fonll.py
+++ b/src/pineko/fonll.py
@@ -122,7 +122,7 @@ class FONLLInfo:
 def update_fk_theorycard(combined_fk, input_theorycard_path):
     """Update theorycard entries for the combined FK table.
 
-    Update by reading the yamldb of the original theory.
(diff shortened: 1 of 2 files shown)


scarlehoff (approved):
With this PR as far as I can see the problem is solved and all fktables produced by pineko have correct theories.
All (dis) fktables that we have in the server probably have invalid json in them but if nobody noticed them until now we can probably live with it.

As to why I was not able to figure out where it was being written, I happened to have a few dangling fktables from before the PTO issue which get read and trigger the error. Most likely those were generated by the bug this PR solves

alecandido:
Ok, then maybe that one was actually the only error

alecandido (approved):

Issue #146, opened 2023-12-15
`mu0` scale is not updated correctly in the operator cards.


giacomomagni:
We have a potential problem in the function that writes the operator cards, if the `Q0` scale provided in the theory card is not the same as the one defined as `mu0` in the template. 

In particular here: 

https://github.com/NNPDF/pineko/blob/f653ee289edcb98222b7e76572dac2d3465c2e08/src/pineko/evolve.py#L84

we should add something as:

```
operators_card["mu0"] = tcard["Q0"]
```

Thanks @toonhasenack  and @Radonirinaunimi

felixhekhorn:
I agree, this is an oversight.

PS: remember that in https://github.com/NNPDF/eko/pull/295 we also shift $n_{f,0}$ from theory to operator

giacomomagni:
> PS: remember that in https://github.com/NNPDF/eko/pull/295 we also shift nf,0 from theory to operator

okay thanks for the reminder. That one do not require an hot fix now, right?

giacomomagni:
I feel we have been lucky here...

felixhekhorn:
> okay thanks for the reminder. That one do not require an hot fix now, right?

no that is still for the now v0.15 series and unmerged

> I feel we have been lucky here...

I agree :see_no_evil:

PR #169, opened 2024-04-11
Use nnpdf single-truth theory

scarlehoff:
Changes to use https://github.com/NNPDF/nnpdf/pull/2028 with pineko.

At the moment it doesn't do anything clever, it just loads the yaml from there. I've tested the eko and fktable generation.
(the tests and so won't pass until the PR is merged in nnpdf)


scarlehoff:
I've made a change (wrt #146). Now if the template and the theory have a different `mu0` this raises an Exception instead of overwriting the value silently.
I can revert the change back of course, but I think this is something important enough to be fail as early as possible (and the solution is "simply" to change the template or theory if one really wants to play god)

src/pineko/evolve.py
```diff
@@ -122,8 +122,12 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
     xif = 1.0 if sv_method is not None else tcard["XIF"]
     # update scale variation method
     operators_card["configs"]["scvar_method"] = sv_method
-    # update initial scale mu0
+
+    # Make sure that we are using the theory Q0 and fail if the template has a different oen
```
felixhekhorn, line 126:
```suggestion
    # Make sure that we are using the theory Q0 and fail if the template has a different one
```

scarlehoff:
One problem here is the nfonnl part.
However, I believe that, since now one can run nfonll by providing only the main theory id, the only logical thing is that the theory cards can also be autogenerated in the go (and so tcards should not be necessary).

This helps reducing the extra amount of theory cards and having to tell NNPDF about things like `PTOEKO` or `FONLLParts`.

scarlehoff:
With the last commit from here https://github.com/NNPDF/nnpdf/pull/2067
this works with pineko to create theories without having to provide a theory or yamldb folder.

docs/source/overview/prerequisites.rst
```diff
@@ -18,35 +18,47 @@ This is a standard example:
...
 
 
 All the relevant inputs are described below. The command ``pineko scaffold new`` will generate all necessary folders.
 
+
+Theory Runcards
+---------------
+
+You need to provide the necessary theory runcards named with their respective theory ID inside the *paths.theory_cards* folder [1]_ unless you are using the ``nnpdf`` key (see below).
+For more details about theory runcards you can look at https://eko.readthedocs.io/en/latest/code/IO.html under **Theory Runcards**.
```
felixhekhorn, line 43: I'm not sure how much that link is up-to-date (the answer is it isn't) and it is surely not the official source of truth of which we are speaking here - instead it is NNPDF, i.e. https://docs.nnpdf.science/theory/theoryparamsdefinitions.html should be linked instead

scarlehoff, line 43:
```suggestion
For more details about theory runcards you can look at https://docs.nnpdf.science/theory/theoryparamsdefinitions.html
```

docs/source/overview/running.rst
```diff
@@ -95,3 +95,37 @@ With the command ``pineko compare`` it is possible to compare the predictions as
 
 again eventually specifying the values of *renormalization* and *factorization* scales with the
 appropriate options.
+
+Using pineko with NNPDF
+"""""""""""""""""""""""
+
+It is possible to use ``pineko`` without providing a mapping between data and grids
```
giacomomagni, line 102: well this mapping is just hidden in the nnpdf metadata...

src/pineko/theory.py
```diff
@@ -330,8 +330,11 @@ def eko(self, name, _grid, tcard):
         # from the PTO needed for the PDF evolution (and so by EKO). Here we
         # ensure that the PTO used in the EKO calculation reflects the real
         # perturbative order of the prescription.
-        if "PTOEKO" in tcard:
+        if tcard.get("PTOEKO") is not None:
             tcard["PTO"] = tcard["PTOEKO"]
+        # Keys still needed by eko below
```
giacomomagni, line 335:
```suggestion
        # Deprecated keys still needed by eko below. TODO: remove them asap.
```

giacomomagni (approved):
I left two two useless comments.

felixhekhorn (approved):
I dropped most of the yamldb stuff, because it is not used here and nowhere else

docs/source/overview/running.rst
```diff
@@ -95,3 +95,37 @@ With the command ``pineko compare`` it is possible to compare the predictions as
...
+(i.e., without a ``yamldb`` database) or theory cards, by using the data declared in the NNPDF
+repository.
```
scarlehoff, line 104:
```suggestion
It is possible to use ``pineko`` without providing an explicit mapping between data and grids
(i.e., without a ``yamldb`` database) or theory cards, by using the data declared in the NNPDF
repository (which includes a data-theory mapping).
```

docs/source/overview/running.rst
```diff
@@ -95,3 +95,37 @@ With the command ``pineko compare`` it is possible to compare the predictions as
...
 appropriate options.
+
+Using pineko with NNPDF
+"""""""""""""""""""""""
+
+It is possible to use ``pineko`` without providing an explicit mapping between data and grids
+(i.e., without a ``yamldb`` database) or theory cards, by using the data declared in the NNPDF
+repository (which includes a data-theory mapping).
+
+In order to do this you need to install ``pineko`` with the ``nnpdf`` extra, which will install
```
felixhekhorn, line 106:
```suggestion
In order to do this you need to install ``pineko`` with the ``nnpdf`` extra, which installs
```
no need for future here, I would say - it is doing it right now

PR #171, opened 2024-04-16
Expose eko info

### LINKED ISSUE #170: Save eko operator card and theory card in the fktable
The theory and operator information is very cheap compared with the weight of the fktable so we should save it.

felixhekhorn:
Closes #170

since I haven't run pineko in a while myself, I only have a very limited testing ground - please some one of you have a more thorough check

### DIFF
--- a/src/pineko/evolve.py
+++ b/src/pineko/evolve.py
@@ -1,6 +1,7 @@
 """Tools related to evolution/eko."""
 
 import copy
+import json
 import logging
 import os
 import pathlib
@@ -269,6 +270,8 @@ def evolve_grid(
     rich.print(f"Optimizing for {assumptions}")
     fktable.optimize(assumptions)
     fktable.set_key_value("eko_version", operators.metadata.version)
+    fktable.set_key_value("eko_theory", json.dumps(operators.theory_card.raw))
+    fktable.set_key_value("eko_operator", json.dumps(operators.operator_card.raw))
     fktable.set_key_value("pineko_version", version.__version__)
     # compare before/after
     comparison = None
--- a/src/pineko/evolve.py
+++ b/src/pineko/evolve.py
@@ -270,8 +270,8 @@ def evolve_grid(
     rich.print(f"Optimizing for {assumptions}")
     fktable.optimize(assumptions)
     fktable.set_key_value("eko_version", operators.metadata.version)
-    fktable.set_key_value("eko_theory", json.dumps(operators.theory_card.raw))
-    fktable.set_key_value("eko_operator", json.dumps(operators.operator_card.raw))
+    fktable.set_key_value("eko_theory_card", json.dumps(operators.theory_card.raw))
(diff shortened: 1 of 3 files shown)


scarlehoff:
Thanks.
I've tested this and it does the right thing.

However, maybe it makes more sense to have this in `theory.py` after the call to `evolve.evolve_grid` has already happened?

Otherwise, we could add extra keys to `evolve_grid` to be burned into the evolved fktable. In particular I'm thinking about the nnpdf theory card, which would be interesting to have as well (ideally all fktables would've have been done with the same... but that's not always the case...)

src/pineko/evolve.py
```diff
@@ -269,6 +270,8 @@ def evolve_grid(
     rich.print(f"Optimizing for {assumptions}")
     fktable.optimize(assumptions)
     fktable.set_key_value("eko_version", operators.metadata.version)
+    fktable.set_key_value("eko_theory", json.dumps(operators.theory_card.raw))
+    fktable.set_key_value("eko_operator", json.dumps(operators.operator_card.raw))
```
andreab1997, line 274: Is there a particular reason to call it `eko_operator`? Because I find rather confusing given that we are actually talking about the operator card. I would rather call it `eko_operator_card` or `eko_opcard`.

src/pineko/evolve.py
```diff
@@ -1,6 +1,7 @@
 """Tools related to evolution/eko."""
 
 import copy
+import json
```
andreab1997, line 4: Is this really needed?

src/pineko/evolve.py
felixhekhorn: I was first thinking the same, but then I thought "its obvious" - turns out it is not, we can just change -done in [697af5f](https://github.com/NNPDF/pineko/pull/171/commits/697af5f40ab3eee4e93d83abcb9c5058258e0986)

src/pineko/evolve.py
felixhekhorn: after #166 I would say yes, do you agree?

andreab1997 (approved):

scarlehoff (changes requested):
Pleasee add also the nnpdf theory card so we have the full set of information :__

felixhekhorn:
> plz add also the nnpdf theory card so we have the full set of information :__

done in https://github.com/NNPDF/pineko/pull/171/commits/84d2ea819ac9c7a07cbff2ba471ee80d95b4c5fe

andreab1997:
@scarlehoff I believe this is ready to go now (benchmarks were failing only for codecov)

scarlehoff:
Maybe we should remove the upload of the data to codecov. I honestly don't really see the value on having it uploaded to a 3rd party server anyway.

felixhekhorn:
Please merge if your happy with our code changes!

Re codecov: it's becoming more and more unreliable ... I would like to have a reliable badge, but it isn't ... e.g. I think the eko 47% is a lie ...

PR #178, opened 2024-05-08
Add --min_as option to convolute

scarlehoff:
Although maybe the way I'm doing it is a bit... convoluted (canned laughter)

### DIFF
--- a/src/pineko/cli/convolute.py
+++ b/src/pineko/cli/convolute.py
@@ -17,6 +17,7 @@ from ._base import command
 @click.argument("max_al", type=int)
 @click.option("--xir", default=1.0, help="renormalization scale variation")
 @click.option("--xif", default=1.0, help="factorization scale variation")
+@click.option("--min_as", type=int, help="Minimum exponent of as")
 @click.option(
     "--pdf", default=None, help="if given, print comparison table", show_default=True
 )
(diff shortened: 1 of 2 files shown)


src/pineko/cli/convolute.py
```diff
@@ -26,14 +27,18 @@
...
     """Convolute PineAPPL grid and EKO into an FK table.
 
     GRID_PATH and OP_PATH are the path to the respective elements to convolute, and
     FKTABLE is the path where to dump the output.
 
     MAX_AS and MAX_AL are used to specify the order in QCD and QED
-    couplings (i.e. the maximum power allowed for each correction).
+    couplings (i.e. the maximum power allowed for each correction)
+    e.g., max_as = 1, max_al = 0 would select LO QCD only
+    max_as = 3 instead would select LO, NLO, NNLO QCD
```
andreab1997, line 41: Can you a comment also about `min_as`?

andreab1997 (changes requested):
In general, I was not aware that this feature could be needed. Is there a specific application for which we would need to select only orders beyond, for example, NLO?

scarlehoff:
Yes. If you want to create an fktable with only the NNLO contribution for instance. It is not usual (for instance when doing a theory you won't) but we (w/ @cschwan) are using it now to study the NNLO of DY as a function of q.

This is equivalent to doing in pineappl:

```
pineappl evolve ${grid} ${eko} ${fktable} NNPDF40_nnlo_pch_as_01180 --orders "a2as2"
```

andreab1997:
> Yes. If you want to create an fktable with only the NNLO contribution for instance. It is not usual (for instance when doing a theory you won't) but we (w/ @cschwan) are using it now to study the NNLO of DY as a function of q.
> 
> This is equivalent to doing in pineappl:
> 
> ```
> pineappl evolve ${grid} ${eko} ${fktable} NNPDF40_nnlo_pch_as_01180 --orders "a2as2"
> ```
> 
> in pineko it would become
> 
> ```
> pineko convolute ${grid} ${eko} ${fktable} 3 1 --min_as 3
> ```

Yes, I guess that in `pineko` this is surely not the main feature so it makes sense to implement it as an optional argument. It is fine for me to merge this as soon as the docstrings are updated.

andreab1997 (approved):

PR #192, opened 2024-08-23
Support Timelike with up to 2 ekos

giacomomagni:
Support time-like evolution with up to 2 ekos.

### DIFF
--- a/src/pineko/evolve.py
+++ b/src/pineko/evolve.py
@@ -245,10 +245,6 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
     kv = pineappl_grid.key_values()
     conv_type_1, conv_type_2 = get_grid_convolution_type(kv)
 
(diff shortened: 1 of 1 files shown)


giacomomagni:
@t7phy I'll leave to you the pleasure to test and check if it works... 
Most likely there is some corner case left 😅

src/pineko/evolve.py
```diff
@@ -245,10 +245,6 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
     kv = pineappl_grid.key_values()
     conv_type_1, conv_type_2 = get_grid_convolution_type(kv)
 
-    # fragmentation function grid?
-    if "timelike" in kv:
```
giacomomagni, line 241: Actually the previous implementation was not even working cause the "True" business

scarlehoff:
Deprecated by #206

PR #193, opened 2024-09-02
Fix bug in sv operator cards

andreab1997:
I think that line is wrong at the moment. Can you please check? @felixhekhorn @giacomomagni

### DIFF
--- a/src/pineko/evolve.py
+++ b/src/pineko/evolve.py
@@ -197,7 +197,7 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
     muf2_grid = evol_info.fac1
     operators_card = copy.deepcopy(default_card)
     sv_method = sv_scheme(tcard)
-    xif = 1.0 if sv_method is not None else tcard["XIF"]
+    xif = 1.0 if sv_method is None else tcard["XIF"]
     # update scale variation method
     operators_card["configs"]["scvar_method"] = sv_method
 
--- a/src/pineko/evolve.py
+++ b/src/pineko/evolve.py
@@ -397,13 +397,14 @@ def evolve_grid(
     )
     # To compute the alphas values we are first reverting the factorization scale shift
     # and then obtaining the renormalization scale using xir.
+    ren_grid2 = xir * xir * mur2_grid
     alphas_values = [
         4.0
         * np.pi
         * sc.a_s(
-            xir * xir * mur2,
+            mur2,
         )
-        for mur2 in mur2_grid
+        for mur2 in ren_grid2
     ]
 
     def prepare(operator):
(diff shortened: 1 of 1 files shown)


src/pineko/evolve.py
```diff
@@ -426,15 +426,15 @@ def prepare(operator):
         fktable = grid.evolve_with_slice_iter2(
             prepare(operators1),
             prepare(operators2),
-            ren1=mur2_grid,
+            ren1=xir*xir*mur2_grid,
```
giacomomagni, line 429: This looks correct, I was introduced in #181, my bad

src/pineko/evolve.py
```diff
@@ -354,7 +354,7 @@ def evolve_grid(
     evol_info = grid.evolve_info(order_mask)
     x_grid = evol_info.x1
     mur2_grid = evol_info.ren1
-    xif = 1.0 if operators1.operator_card.configs.scvar_method is not None else xif
+    xif = 1.0 if operators1.operator_card.configs.scvar_method is None else xif
```
giacomomagni, line 357: About this I'm not entirely sure I understand the logic... it has been there since a while (18 months)

felixhekhorn: to be precise https://github.com/NNPDF/pineko/commit/f18c623496cb4fc9fa914d0b36960a3e65a49518 which is part of https://github.com/NNPDF/pineko/pull/82 - and I think this makes sense, because for scheme A and B eko does on it's own the necessary integral adjustments. Instead for scheme C pineko has to take care (as eko can not know); so this new changes seems to be wrong (same arguments apply of course above)

src/pineko/evolve.py
felixhekhorn: agreed:
https://github.com/NNPDF/pineko/blob/29a02b95774cd3b39af1334ca2e3b30dbd83b473/src/pineko/evolve.py#L278

src/pineko/evolve.py
andreab1997: About this I remember now the logic and I agree with @felixhekhorn . I started playing with this because I was getting an error from pineappl when producing the fktables but turns out that the other fix in this PR is enough to solve everything. So let me revert this change here

src/pineko/evolve.py
```diff
@@ -419,22 +419,22 @@ def prepare(operator):
                 pid_basis=PyPidBasis.Evol,
             )
             yield (info, op.operator)
-
+    ren_grid2 = xir*xir*mur2_grid
```
giacomomagni, line 422: you can even define it before `alphas_values` so it's more clear for eventual upcoming changes

giacomomagni (approved):

felixhekhorn (approved):

andreab1997:
Ok thanks, I am merging this

PR #200, opened 2024-10-20
add nnpdf dependency to 'test' group

RoyStegeman:
To avoid pytest from failing we need nnpdf to be downloaded as part of the pypi upload workflow: https://github.com/NNPDF/workflows/blob/v2/.github/workflows/python-poetry-pypi.yml

Changing that workflow makes it less general, while adding nnpdf to 'test' dependencies doesn't look very nice, but making non-optional is perhaps also overkill (or we can skip the failing test_utils.py test of the nnpdf_data package). What is the preferred solution?

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -24,11 +24,11 @@ jobs:
           virtualenvs-create: false
           installer-parallel: true
       - name: Install dependencies
-        run: poetry install --no-interaction --no-root --with test -E nnpdf
+        run: poetry install --no-interaction --no-root --with test
       - name: Install project
         # it is required to repeat extras, otherwise they will be removed from
         # the environment
-        run: poetry install --no-interaction -E nnpdf
+        run: poetry install --no-interaction
       - name: Install task runner
         run: pip install poethepoet
       - name: Lint with pylint
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,6 @@ pandas = "^2.1"
 rich = "^12.5.1"
 click = "^8.0.4"
 tomli = "^2.0.1"
-nnpdf = { git = "https://github.com/NNPDF/nnpdf", optional = true}
 lhapdf-management = { version = "^0.5", optional = true }
 
 [tool.poetry.group.docs]
@@ -51,6 +50,7 @@ setuptools = "^69.0"
 optional = true
 
 [tool.poetry.group.test.dependencies]
+nnpdf = { git = "https://github.com/NNPDF/nnpdf" } # needed for test_utils
 pytest = "^7.1.3"
 pytest-cov = "^4.0.0"
 pytest-env = "^0.6.2"


alecandido:
@RoyStegeman you could have it both optional (for the package, such that users can specify the extra) and as a test dependency, because it's true that to run (some) tests you'll need it (DD:).

Do you see any problem with that?

RoyStegeman:
What is the benefit of having it twice? It seems version restrictions are still resolved even if the dependency is in a group: https://python-poetry.org/docs/managing-dependencies

 so I think this means that if we do `poetry add nnpdf` it's the same whether we put it twice or not.

alecandido:
They are fundamentally two different things.

Dependencies outside the groups are those entering the package, so they will be resolved also by the users installing from the wheel (e.g. from PyPI) and not from source. For them, `pyproject.toml` will not be available, nor `poetry.lock`, but only the `*.dist-info/METADATA` file generated by the build process. The extra is needed to enter that file, and be distributed to the user.

Instead, the group is a feature required to manage the environment. You decide for what you need that, but my personal advice would be that if you have a group named "test(s)" it should contain all the dependencies required to run the tests. Thus, it is worth to repeat the optional dependencies with the implied meaning "if I'm running the tests, this is not optional any longer".

This is the rationale, and to me ugly it's not ugly to have either, and advisable to have both. But of course you could handle it also in different ways :)

scarlehoff:
I agree, if it is an optional in one case and necessary in the other so be it, it needs to appear in both.

felixhekhorn:
Mmm I don't like this change, because it makes nnpdf mandatory

counter proposal: we add a (pytest) marker and if necessary we can also add a shortcut for selecting (similar to eko)

scarlehoff:
> Mmm I don't like this change, because it makes nnpdf mandatory

Only for tests though. pypi should not see nnpdf right? (only via `pip install pineko[nnpdf]`)

felixhekhorn:
> > Mmm I don't like this change, because it makes nnpdf mandatory
> 
> Only for tests though. pypi should not see nnpdf right? (only via `pip install pineko[nnpdf]`)

yes - let me be clearer: "it makes nnpdf mandatory for me, which I try to fight" :upside_down_face: (and actually this fight could even be wise for both nnpdf and pineko in terms of dependency management

alecandido:
Personally, I would have liked that Pineko would have depended on `nnpdf_data` alone, rather than `nnpdf`, like the following:
```mermaid
flowchart RL
    nnpdf --> nnpdf_data
    pineko --> nnpdf_data
```
instead of
```mermaid
flowchart RL
    pineko --> nnpdf --> nnpdf_data
```
since `nnpdf_data` is truly the only thing required by Pineko (unless I'm still missing something...).

However, for as long as things stay as it is, I guess it shouldn't be a big deal, if only TensorFlow was left as an optional dependency in `nnpdf` (instead of entering all the isolated environments or forcing everyone to use `--system-site-packages`, which is renown as [bad practice](https://xkcd.com/1987/) for development).

scarlehoff:
It only depends on nnpdf_data, technically... but someone needs to create the nnpdf_data package :__ (and I don't want to distract human resources from the data-implementation tasks in which everyone should be concentrated now! But I'll make it a priority afterwards!)

felixhekhorn:
I reread the head of this PR more carefully and the problem @RoyStegeman tries to solve is the PyPI workflow. However, the workflow allows to specify extras which are installed along the way:
https://github.com/NNPDF/workflows/blob/e681e4d636ae8000bd806e8124119c180dd7c6b7/.github/workflows/python-poetry-pypi.yml#L16
and in the workflow this is not active yet:
https://github.com/NNPDF/pineko/blob/main/.github/workflows/pypi.yml
so why don't we just add it there?

RoyStegeman:
Ah you're right @felixhekhorn! I hadn't appreciated that option. That is indeed the ideal solution for this issue (and a separate one from the packaging of nndpf_data)

RoyStegeman:
In the end I agree with @alecandido's reasoning - if we want to do test we need download also nnpdf so this should be part of the test group. 

If you are happy with the current solution, can someone approve this PR?

PR #205, opened 2024-10-30
Support eko `0.15`

giacomomagni:
This PR is to support the new version of eko.

### DIFF
--- a/src/pineko/evolve.py
+++ b/src/pineko/evolve.py
@@ -16,6 +16,8 @@ import rich.box
 import rich.panel
 import yaml
 from eko import basis_rotation
+from eko.interpolation import XGrid
+from eko.io import manipulate
 from eko.io.types import ScaleVariationsMethod
 from eko.matchings import Atlas, nf_default
 from eko.quantities import heavy_quarks
@@ -201,10 +203,12 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
     # update scale variation method
     operators_card["configs"]["scvar_method"] = sv_method
 
-    # Make sure that we are using the theory Q0 and fail if the template has a different one
-    operators_card["mu0"] = tcard["Q0"]
-    if default_card.get("mu0") is not None and default_card["mu0"] != tcard["Q0"]:
-        raise ValueError("Template declares a value of Q0 different from theory")
+    operators_card["init"] = (tcard["Q0"], tcard["nf0"])
+    if default_card.get("init") is not None and default_card["init"] != [
+        tcard["Q0"],
+        tcard["nf0"],
+    ]:
+        raise ValueError("Template declares a value of Q0, nf0 different from theory")
 
     q2_grid = (xif * xif * muf2_grid).tolist()
     masses = np.array([tcard["mc"], tcard["mb"], tcard["mt"]]) ** 2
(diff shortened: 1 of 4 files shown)


Radonirinaunimi:
I believe the easiest solution here is to reshape the `data_files/data/ekos/400/HERA_NC_225GEV_EP_SIGMARED.tar` EKO to be squared. I attempted to do so by using older version of EKO, fixing some errors here and there, but didn't fully manage. @felixhekhorn would you be perhaps able to do so? I think for you it'd be much easier/faster instead of me banging my head against the wall...

felixhekhorn:
@Radonirinaunimi  Sorry, I had not time so far to look into this (and I will be on holidays next week) so I wanted to see if I can do something quickly now. Do you still need my help? (we can also chat elsewhere :upside_down_face: )

As for `mu0`: I don't understand - look here:
https://github.com/NNPDF/eko/blob/c69a7b964a68342cfac728627be1a2c1084287ed/src/eko/io/v1.py#L81
and
https://github.com/NNPDF/eko/blob/c69a7b964a68342cfac728627be1a2c1084287ed/src/eko/io/v2.py#L50

Radonirinaunimi:
I was thinking that one should do:
```Python
new_ocard = legacy_class.new_operator
```
after:
https://github.com/NNPDF/pineko/blob/f4097f29dec964e8b2c17818b502575ffa3444f0/src/pineko/theory.py#L372-L373

But that doesn't work because of various errors (again, using the OP cards above).

src/pineko/evolve.py
```diff
@@ -371,28 +357,34 @@ def xgrid_reshape(full_operator):
     # To compute the alphas values we are first reverting the factorization scale shift
     # and then obtaining the renormalization scale using xir.
     ren_grid2 = xir * xir * mur2_grid
+    # TODO: double-check that getting `nf` from the first Operator is always correct
```
felixhekhorn, line 360: I think "by chance" this is correct, because we have only two kind of datasets in game: LHC data, which is always in nf=5 (and so is the first EKO), OR nFONLL grids, which are at at a fixed nf by construction (and so is the first EKO). This would not work for aFONLL since the nf in the observable and alpha_s would need to be in sync (alphas in different flavors is an explicit player in FONLL). The true story is as usual more complicated: actually the grid = PineAPPL should support evolution points, since it is the observable which dictates the evolution point (both for PDF=eko and alpha_s) - we just implicitly assume the above categorization.

felixhekhorn (commented):
some minor comments

benchmarks/bench_regression.py
```diff
@@ -72,9 +72,10 @@ def xfxQ2(self, pid, x, q2):
 
 def _trim_template(template_card, take_points=10):
     """Trim the template card so that the number of x-values to compute is much smaller"""
-    card_info = OperatorCard.from_dict(
-        safe_load(template_card.read_text(encoding="utf-8"))
-    )
+    raw_card = safe_load(template_card.read_text(encoding="utf-8"))
+    raw_card["init"] = (raw_card["mu0"], 4)
+    del raw_card["mu0"]
```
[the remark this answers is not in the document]
Radonirinaunimi: Actually this still has to be there because the operator cards are downloaded from https://github.com/NNPDF/theories; don't ask me why ` ¯\_(ツ)_/¯`

src/pineko/evolve.py
```diff
@@ -103,9 +105,11 @@ def write_operator_card_from_file(
     if not pathlib.Path(pineappl_path).exists():
         raise FileNotFoundError(pineappl_path)
     pineappl_grid = pineappl.grid.Grid.read(pineappl_path)
+    pineappl_grid.optimize()
```
scarlehoff, line 108: Why is there an optimize here? I think this might be the cause of #228

giacomomagni: Yes, the problem was that another optimize was already present in pineko evolve (https://github.com/NNPDF/pineko/blob/0c01d8d480a16821bf1e24e3884420bb9d72f146/src/pineko/theory.py#L435-L436) , so I added one before the `opcard` generation because otherwise there were some mismatches. 
But I'm not sure why the grid in #228 can't be optimised.

scarlehoff: I'm not entirely sure how I feel about that. I would remove both optimize because ptherwise you cannot use directly pineappl evolve since the original grid and the ekos don't match.

Radonirinaunimi: On a second thought (again), I agree with @felixhekhorn. In my comment above, I was still thinking of scenarios in which optimise could "modify" the grids (scenarios such as the initial states are different but optimised in the same way). But from v1 that will never happen anymore. So one **should** always optimise by default.

felixhekhorn: > That said, practically I would either remove them or wrap them with try-except so we can move forward with the current version of the codes.

This is the short-term solution, I agree. However, mid-term we definitely want to fix the PineAPPL bug instead ...

> But wouldn't that mean that ekos created with pineko cannot be used with `pineappl evolve`? (as you would need to optimize to make sure the grids and channels match)

This would only happen in a scenario, which I consider unlikely: one has a grid, calls optimize without saving the grid, then extract the necessary ekos, and then use the original (unoptimized) grid with the ("optimized") EKO. I would not support such a thing - or if you want to, force directly an optimize inside `pineappl evolve`.

felixhekhorn: > edit: We can enforce an o `optimize`

I was about to suggest the edit, i.e. I'm much in favour of it :innocent: 

Also as eko developer I would say, that computing an eko is expensive and worth to optimize :see_no_evil:

felixhekhorn: > I don't think this would be a problem though! The EKO grids are Q 2 slices of a 4D array of ( f out , x out , f in , x in ) and PineAPPL doesn't require f in to match the grid. So, as long as x in matches, everything should work fine.

`optimize` can remove $Q^2$ which would be a problem for the CLI (but is the feature we=eko wants)

Issue #183, opened 2024-06-30
Rotate FK-tables instead of EKOs


cschwan:
There's another big optimization potential to speed up the generation of FK-tables. Currently, I believe, pineko works the following way (correct me if I'm wrong):
1. we store flavour-basis to flavour-basis EKOs for all Q2 slices, but always for the standard 50 x-node points. The biggest EKO that we have, `ATLAS_1JET_8TEV_R06`, is 2.4 GB.
2. Next, we reinterpolate the EKO to match the x-grid points actually used in the grid. For `ATLAS_1JET_8TEV_R06` this blows up the size of the EKO to 22 GB.
3. Finally, we rotate the EKO, such that the resulting FK-tables are in the evolution basis. This further increases the size, resulting in whopping 45 GB.

But thanks to some [developments on PineAPPL's side](https://github.com/NNPDF/pineappl/commit/ba0bbf92ab912100941755b07e30f8bb07e5795c), we can now perform a rotation at the level of grids (and therefore also FK-tables). So I suggest that
- we perform the evolution using the unrotated EKOs, which has the advantage that the EKOs are much smaller, and finally
- rotate the evolved FK-tables from the flavour-basis to the evolution basis.

The evolution in the flavour basis is much faster, it's roughly the quotient of the size of the EKOs. For `ATLAS_1JET_8TEV_R06` it's roughly twice as fast (`45 GB / 22 GB`). With the added benefit that we need less disk space, of course.

Practically speaking, the changes on pineko's side should be minimal. We probably have to delete the following lines:
https://github.com/NNPDF/pineko/blob/9f8b03ce9132c91866a160e4396998abaca843c2/src/pineko/evolve.py#L282-L286
and instead add the rotation after the evolution:
https://github.com/NNPDF/pineko/blob/9f8b03ce9132c91866a160e4396998abaca843c2/src/pineko/evolve.py#L313-L320

alecandido:
Yes, I definitely agree about this, and we already discussed many times in the past with @felixhekhorn.

The only limitation in this respect was never PineAPPL itself, but just person power. In principle, we didn't even need to be able to rotate grids in order to implement this optimization: it would have been sufficient for EKO to expose the rotation, and for PineAPPL to consume during evolution, i.e. at the end of the evolution.
[...]

PR #206, opened 2024-11-13
Upgrade to PineAPPL `v1`

Radonirinaunimi:
This PR tests the new Python interface of the upcoming PineAPPL `v1`. There are a few notable changes affecting `pineko` directly:
- The types of convolutions (be it for grids or FK tables) are accessed through the `convolutions` attributes. These information are no longer determined from the metadata (address the shortcomings discussed in https://github.com/NNPDF/nnpdf/pull/2129#discussion_r1688126656 and https://github.com/NNPDF/pineko/pull/181#discussion_r1689312568).
- Grids can be evolved with an arbitrary number of EKOs (no longer limited to two operators). This removes the need for `operator1` and `operator2` in favor of just `operators`. 
- Grids and FK tables can be convolved with an arbitrary number of PDFs (no longer limited to two PDFs). Similarly, this removes the need for `pdf1` and `pdf2` in favor of `pdfs`.
- Add a new variation of scale - which is the fragmentation scale `xia`.

The only remaining thing for which we will still require an ad-hoc hack/tweak is the checks that the PDFs are ordered according to the type of convolutions. This is an LHAPDF problem so there isn't much that we can do.

This will deprecate #192.

Remaining to do:
- [x] profile the computation of the Jet EKOs


cschwan:
When this is working I suggest you also try to implement #183 - this should decrease the time and space needed to perform an evolution. Also the required changes should be minimal.

Radonirinaunimi:
> When this is working I suggest you also try to implement #183 - this should decrease the time and space needed to perform an evolution. Also the required changes should be minimal.

Thanks for letting me know about this! I will surely do once this is working indeed.

src/pineko/check.py
```diff
@@ -117,26 +111,25 @@ def check_grid_and_eko_compatible(pineappl_grid, operators, xif, max_as, max_al)
...
     ----------
-    lumi : list(list(tuple))
-           luminosity info
+    channels : list(list(tuple(list, float)))
+        object containing the information on the channels. From PineAPPL v1,
+        the tuple contains two elements. The first is a list containing the
+        PIDs of the partons. The length of the list coincides with the number
+        of partons involved, ie for DIS there is only one element, and for a
+        hadron production in pp collision there are three elements. The 2nd
+        element is the weight/multiplicative factor as usual.
```
giacomomagni, line 125:
```suggestion
        element is the weight/multiplicative factor.
```

benchmarks/bench_comparator.py
```diff
@@ -8,9 +9,10 @@ def benchmark_compare(lhapdf_path, test_files, test_pdf):
     grid = pineappl.grid.Grid.read(pine_path)
     fk_path = test_files / "data/fktables/208/LHCB_DY_13TEV_DIMUON.pineappl.lz4"
     fk = pineappl.fk_table.FkTable.read(fk_path)
-    pdf = "NNPDF40_nlo_as_01180"
+    pdfs = ["NNPDF40_nlo_as_01180"]
+    xi = (1.0, 1.0, 1.0)
     with lhapdf_path(test_pdf):
-        comp_table = pineko.comparator.compare(grid, fk, 2, 0, pdf, 1.0, 1.0)
+        comp_table = pineko.comparator.compare(grid, fk, 2, 0, pdfs, xi)
+    # TODO: Check if there is NOW a slight difference in PineAPPL convolve
```
giacomomagni, line 16: I guess this is now passing...

```suggestion
```

src/pineko/cli/convolve.py
```diff
@@ -58,48 +57,55 @@ def subcommand(
...
+    of the PDF sets, ordered in the same as the convolution types in the GRID/FK table.
+    The PDFs are passed as strings with the names separated by commas.
     """
     grid = pineappl.grid.Grid.read(grid_path)
     n_ekos = len(op_paths)
-    with eko.EKO.edit(op_paths[0]) as operators1:
+    with eko.EKO.edit(op_paths[0]) as first_operator:
+        operators = [first_operator]
+        path_operators = f"[+] {op_paths[0]}\n"
+        # If there are more than ONE operators, then account for all of them.
```
giacomomagni, line 76:
```suggestion
        # If there are more than ONE operator, then account for all of them.
```

src/pineko/cli/fonll.py
```diff
@@ -122,26 +122,30 @@ def ekos(theoryid, datasets, overwrite):
...
     """Command to generate numerical FONLL FK tables.
 
     1. Produce the 7 FK tables needed for numerical FONLL.
     2. Combine the FKtables into a single one.
     """
     # create the 7 FK tables
+    pdfs = pdfs.split(",") if pdfs is not None else pdfs
     for th_suffix in range(0, 7):
-        theory.TheoryBuilder(
+        theory.TheoryBuilder(  # [too-many-function-args]
```
giacomomagni, line 142: Is this really needed ? 
```suggestion
        theory.TheoryBuilder(
```

src/pineko/comparator.py
```diff
@@ -19,14 +18,11 @@ def compare(pine, fktable, max_as, max_al, pdf1, xir, xif, pdf2=None):
...
         maximum power of electro-weak coupling
-    pdf1 : str
-        PDF set name
-    xir : float
-        renormalization scale variation
-    xif : float
-        factorization scale variation
-    pdf2: str or None
-        PDF set for the second convolution, if different from the first
+    pdf1 : list(str)
```
giacomomagni, line 21:
```suggestion
    pdfs : list(str)
```

src/pineko/evolve.py
```diff
@@ -284,30 +266,27 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
...
-        dump_card(card_path, operators_card, conv_type_1)
-    else:
-        # dump card_a
-        dump_card(card_path, operators_card, conv_type_1, suffix=True)
-        # dump card_b
-        dump_card(card_path, operators_card, conv_type_2, suffix=True)
+    # Get the types of convolutions required for this Grid
+    convolutions = pineappl_grid.convolutions
+
+    # TODO: Find a clever way to recombine the same convolutions
```
giacomomagni, line 272: What do you mean ?

src/pineko/evolve.py
```diff
@@ -329,36 +308,35 @@ def evolve_grid(
...
     xif : float
         factorization scale variation
-    operators2: eko.EKO
-        additonal evolution operator if different from operators1
+    xia : float
+        fragmentation scale variation
     assumptions : str
         assumptions on the flavor dimension
-    comparison_pdf1 : None or str
+    comparison_pdf1 : list(str) or None
```
giacomomagni, line 315:
```suggestion
    comparison_pdfs : list(str) or None
```

giacomomagni (commented):
Looks great to me. Some really minor comments: 
 
* Just a small question: why `key_values` are still around and now we don't have just metadata ? Why the theory card is still saved to `key_values` ? 

https://github.com/NNPDF/pineko/blob/e8d72825d6d56be906fb28b9f9186f612944eb41/src/pineko/scale_variations.py#L238-L240

* Will the old grid still be usable / convertible ?
* I did not check, I'm able to produce a real FKtable myself.

cschwan:
> Looks great to me. Some really minor comments:
> 
>     * Just a small question: why `key_values` are still around and now we don't have just metadata ? Why the theory card is still saved to `key_values` ?

Indeed, it should be `{set_,}metadata` instead of `{set_,}key_value`. I'm not why it still works the old way.

src/pineko/evolve.py
Radonirinaunimi: As it is currently implemented, if we have three convolutions `unpolarized PDF⊗unpolarized PDF⊗ unpolarized FF` then the unpolarized PDF operator will be dumped twice while it should only done **once** and called by the two objects.

src/pineko/cli/fonll.py
Radonirinaunimi: I believe so, otherwise one of the linting in the hooks will fail.

src/pineko/evolve.py
giacomomagni: I see but then the `pineko theory ekos` command will compute the ekos twice ?

giacomomagni: Okay good then it's fine!

Radonirinaunimi: After double-checking, this was actually not the case. I pushed a fix in [c792ace](https://github.com/NNPDF/pineko/pull/206/commits/c792ace2dc7ab18a43df50e5ac160a78caa6f65d).

As for the TODO comment, we can simply remove it now.

giacomomagni: Thanks perfect!

src/pineko/evolve.py
```diff
@@ -284,30 +264,27 @@ def write_operator_card(pineappl_grid, default_card, card_path, tcard):
...
-        dump_card(card_path, operators_card, conv_type_1)
-    else:
-        # dump card_a
-        dump_card(card_path, operators_card, conv_type_1, suffix=True)
-        # dump card_b
-        dump_card(card_path, operators_card, conv_type_2, suffix=True)
+    # Get the types of convolutions required for this Grid
+    convolutions = pineappl_grid.convolutions
+
+    # TODO: Find a clever way to recombine the same convolutions
```
giacomomagni, line 270:
```suggestion
```

giacomomagni (approved):
As long as it works it's good for me!

Radonirinaunimi:
> As long as it works it's good for me!

Perfect, thanks! Let's wait for #215 and we can test it here.

Radonirinaunimi:
So luckily I started with #215 first because there were some issues in the FONLL part that were not covered by the tests and benchmarks.  

The main issue, however, is that for some reasons the `FFNS{3,6}-massive` positivity grids are different from the others. Specifically, `pineappl-v1` cannot read the grids because the length of the interpolation and the kinematics are inconsistent, [here](https://github.com/NNPDF/pineappl/blob/279ea0fe7a6636b99bcc00c4c611758634e94259/pineappl/src/grid.rs#L113-L118) is the where the error is raised.

scarlehoff:
>  the length of the interpolation and the kinematics are inconsistent

Is the pineappl CLI able to convolute them though?

(great job with the regression tests btw, already saving our lifes :D )

Radonirinaunimi:
I should note that for HERA, everything is fine. So the difference might be in the way how `pinefarm` produces these grids.

scarlehoff:
I would've assumed that there's no difference for yadism between one and the other?
If you upgrade the grid version or recompute it do you have the same problem?

Radonirinaunimi:
> I would've assumed that there's no difference for yadism between one and the other?

This I am not completely sure because `pinefarm` might be doing some extra-stuff on top.

> If you upgrade the grid version or recompute it do you have the same problem?

I cannot overwrite the version as any `pineappl` operation is not working. That's why I am relying on @comane to provide me with the grids he just generated 😅

Radonirinaunimi:
So I re-computed `NNPDF_POS_F2D_40` for `FFNS{3,6}-massive` with `pineappl-v0.8.6` and the grids can be parsed properly with `v1`. These old POS grids were generated with `pineappl-v0.6.2` and I am not sure at which stage things differ -  but I guess that's to be investigated.

felixhekhorn:
> So I re-computed `NNPDF_POS_F2D_40` for `FFNS{3,6}-massive` with `pineappl-v0.8.6` and the grids can be parsed properly with `v1`. These old POS grids were generated with `pineappl-v0.6.2` and I am not sure at which stage things differ - but I guess that's to be investigated.

but this is a problem, right? we (and others) have a number of old grids around, e.g. for mg5 etc. (and at the very minimum I personally do for MaunaKea, since I haven't updated yet to > v0.8)

> for the structure functions there is no difference

so, just to confirm explicitly: this is not a yadism bug, but a PineAPPL one, right?

scarlehoff:
> but this is a problem, right? we (and others) have a number of old grids around, e.g. for mg5 etc. (and at the very minimum I personally do for MaunaKea, since I haven't updated yet to > v0.8)

Not if the `pineappl write <old grid> --upgrade <new grid>` works with 0.8.6 which is both able to read the old grids and its grids can be used with 1.0. That's why I'm insisting on checking whether the upgrade (not to 1.0 but to 0.8.7) fixes the problem.

Radonirinaunimi:
> but this is a problem, right? we (and others) have a number of old grids around, e.g. for mg5 etc. (and at the very minimum I personally do for MaunaKea, since I haven't updated yet to > v0.8)

But the problem is not the old grid themselves . As I mentioned above, the HERA grids work perfectly fine while they were also generated with `v0.6.2`. The problem only appears when it is a POS grid **and** generated with `~v0.6.2`.

> Not if the pineappl write <old grid> --upgrade <new grid> works with 0.8.6 which is both able to read the old grids and its grids can be used with 1.0. That's why I'm insisting on checking whether the upgrade (not to 1.0 but to 0.8.7) fixes the problem.

Unfortunately, this doesn't solve the problem. I will investigate what exactly the issue and will (potentially) add some compatibility fix in PineAPPL.

Radonirinaunimi:
Ok, so after digging into PineAPPL, I found out the issue. The problem is that positivity grids from `FFNS{3,6}-massive` do not contain any channel (ie `grid.channels().is_empty()==true`). 

This is problematic because PineAPPL uses the information on the channel to infer the type of convolutions from the old grids. See [here](https://github.com/NNPDF/pineappl/blob/279ea0fe7a6636b99bcc00c4c611758634e94259/pineappl/src/v0.rs#L294-L304):

```Rust
let condition = !grid.channels().iter().all(|entry| {
    entry.entry().iter().all(|&(a, b, _)|
        match index {
            1 => a,
            2 => b,
            _ => unreachable!()
        } == pid
    )
});

condition.then_some(Conv::new(ConvType::UnpolPDF, pid))
```
This means that the returned convolution vector is `[None, None]` instead of being:
```Rust
[Conv { conv_type: UnpolPDF, pid: 2212 }, None]
```
And because the number of elements in the interpolation is deduced from the non-`None` elements of the convolution vector (see this [function](https://github.com/NNPDF/pineappl/blob/279ea0fe7a6636b99bcc00c4c611758634e94259/pineappl/src/v0.rs#L13)), it ensues that the size of the interpolation is not the same as the size of the kinematics, and therefore the error in the assertion above.

This is not a problem with grids after `0.8.6` because either the convolution types are read from the metadata or, since `v1`, defined as intrinsic attributes of the grid.

So, we basically have two solutions: (a) re-generate the POS grids with `v0.8.6`, or (b) account for this in PineAPPL.

scarlehoff:
> (b) account for this in PineAPPL.

I believe the right thing to do is for `--upgrade` to take care of it.

However, if it doesn't contain any channels... doesn't that simply mean the grid is empty?

Radonirinaunimi:
> However, if it doesn't contain any channels... doesn't that simply mean the grid is empty?

Yes, they are indeed empty.

> I believe the right thing to do is for --upgrade to take care of it.

`--upgrade` still passes through those same functions for legacy reading, so fixing that will also fix the parsing of the grids (and therefore upgrading them will not be needed). ON how to best implement this, I still need to think a bit.

cschwan:
That's very strange. How can a grid be empty but give a non-zero convolution?

Radonirinaunimi:
> That's very strange. How can a grid be empty but give a non-zero convolution?

Hi @cschwan, the convolutions are actually zero:
```
-------------------------------------------------------------------------
bin   PineAPPL         MC        sigma      central         min      max
                                 1/100   sigma   1/1000   1/1000   1/1000
-------------------------------------------------------------------------
0   0.000000e+00  0.000000e+00 NaN NaN NaN NaN NaN
1   0.000000e+00  0.000000e+00 NaN NaN NaN NaN NaN
2   0.000000e+00  0.000000e+00 NaN NaN NaN NaN NaN
3   0.000000e+00  0.000000e+00 NaN NaN NaN NaN NaN
4   0.000000e+00  0.000000e+00 NaN NaN NaN NaN NaN
5   0.000000e+00  0.000000e+00 NaN NaN NaN NaN NaN
6   0.000000e+00  0.000000e+00 NaN NaN NaN NaN NaN
7   0.000000e+00  0.000000e+00 NaN NaN NaN NaN NaN
8   0.000000e+00  0.000000e+00 NaN NaN NaN NaN NaN
9   0.000000e+00  0.000000e+00 NaN NaN NaN NaN NaN
10  0.000000e+00  0.000000e+00 NaN NaN NaN NaN NaN
11  0.000000e+00  0.000000e+00 NaN NaN NaN NaN NaN
12  0.000000e+00  0.000000e+00 NaN NaN NaN NaN NaN
13  0.000000e+00  0.000000e+00 NaN NaN NaN NaN NaN
14  0.000000e+00  0.000000e+00 NaN NaN NaN NaN NaN
15  0.000000e+00  0.000000e+00 NaN NaN NaN NaN NaN
16  0.000000e+00  0.000000e+00 NaN NaN NaN NaN NaN
17  0.000000e+00  0.000000e+00 NaN NaN NaN NaN NaN
18  0.000000e+00  0.000000e+00 NaN NaN NaN NaN NaN
19  0.000000e+00  0.000000e+00 NaN NaN NaN NaN NaN
```

Just for the record, I added an issue in PineAPPL https://github.com/NNPDF/pineappl/issues/334.

cschwan:
That's even stranger, what's the purpose of an empty grid? Judging from <https://github.com/NNPDF/pineappl/blob/279ea0fe7a6636b99bcc00c4c611758634e94259/examples/python/positivity.py#L69> a positivity grid shouldn't actually be empty. What am I missing here?

Radonirinaunimi:
> That's even stranger, what's the purpose of an empty grid? Judging from https://github.com/NNPDF/pineappl/blob/279ea0fe7a6636b99bcc00c4c611758634e94259/examples/python/positivity.py#L69 a positivity grid shouldn't actually be empty. What am I missing here?

Yes, the final positivity grid must not be empty but some of the intermediate grids in numerical FONLL can be because some contributions may vanish (in this case, it is the massive ones). And then you sum everything over and this will give you final positivity grid.

scarlehoff:
If you plan to merge this before fixing the problem (because it is too difficult in pineappl or whatever), could you add a try-except block to explain a bit the problem? (so that someone finding the problem can just regenerate that grid or remove it)

Radonirinaunimi:
> If you plan to merge this before fixing the problem (because it is too difficult in pineappl or whatever), could you add a try-except block to explain a bit the problem? (so that someone finding the problem can just regenerate that grid or remove it)

This will actually be fixed by https://github.com/NNPDF/pineappl/pull/335. But we'll need to wait for a new release, so to test that everything is working fine (relevant for the next point below), I temporarily commented out the dataset. But this will definitely be re-included as soon as we have a new version.

However, I've found another issue 🤯

As you can see, the regression test is fully working with `v1` (modulo the empty grid issue) but one benchmark is failling because `pineappl` might be parsing an **old** EKO wrongly.

Using `v1`, you can see below that the resulting FK table is completely off:
```sh
pineappl evolve benchmarks/data_files/data/grids/400/HERA_NC_225GEV_EP_SIGMARED.pineappl.lz4 benchmarks/data_files/data/ekos/400/HERA_NC_225GEV_EP_SIGMARED.tar check.pineappl.lz4 NNPDF40_nnlo_as_01180
...
200 6.5651061e-1  -3.2408490e1  -5.0364761e1
201 5.9886400e-1  -3.2060513e1  -5.4535549e1
202 5.1115508e-1  -2.4311589e1  -4.8562061e1
203 4.3146785e-1  -7.7409428e0  -1.8940949e1
204 3.6205190e-1   1.4088141e1   3.7911939e1
205 3.1306435e-1   2.6728773e1   8.4377887e1
206 2.5017238e-1   3.2783871e1   1.3004512e2
207 1.3061592e-1   2.0587443e1   1.5661817e2
208 1.9842599e-2  -2.0272451e1  -1.0226631e3
```
while using `v0.8.6`:
```sh
pineappl evolve benchmarks/data_files/data/grids/400/HERA_NC_225GEV_EP_SIGMARED.pineappl.lz4 benchmarks/data_files/data/ekos/400/HERA_NC_225GEV_EP_SIGMARED.tar check.pineappl.lz4 NNPDF40_nnlo_as_01180
...
200 6.5651061e-1 6.5640136e-1 -1.6640789e-4
201 5.9886400e-1 5.9875473e-1 -1.8246832e-4
202 5.1115508e-1 5.1110183e-1 -1.0416535e-4
203 4.3146785e-1 4.3143906e-1 -6.6714382e-5
204 3.6205190e-1 3.6204251e-1 -2.5927215e-5
205 3.1306435e-1 3.1305374e-1 -3.3876609e-5
206 2.5017238e-1 2.5016857e-1 -1.5247784e-5
207 1.3061592e-1 1.3060681e-1 -6.9766074e-5
208 1.9842599e-2 1.9840153e-2 -1.2325215e-4
```
Any lead? @felixhekhorn, @giacomomagni  Is there an obvious difference between that version of EKO and say the one used here?

cschwan:
@Radonirinaunimi are you using `master` or the prerelease? If you're using `master` please test with the pre-release.

Radonirinaunimi:
> @Radonirinaunimi are you using `master` or the prerelease? If you're using `master` please test with the pre-release.

Everything related to `pineko` is done with the pre-release. I should have used `v1.0.0a2` above to make it explicit xd (fixed now).

cschwan:
> However, I've found another issue 🤯
> 
> As you can see, the regression test is fully working with `v1` (modulo the empty grid issue) but one benchmark is failling because `pineappl` might be parsing an **old** EKO wrongly.

Do you have any evidence that it's the parsing that's wrong?

> 
> Using `v1.0.0a2`, you can see below that the resulting FK table is completely off:

That looks like a definite regression and we should fix it ASAP. I'm traveling right now, so if you want to go ahead, one potential avenue - if the evolution itself is broken and it's not the EKO parsing - is to do printf (or should I say `dbg!`) debugging, by printing the result of the convolution of each evolved slice with a PDF set for both good and bad versions. You probably only need the first bin. In `master`, you'd insert the statements here:

https://github.com/NNPDF/pineappl/blob/18f2b004577204fb59e6e16225e39433377b5a5d/pineappl/src/grid.rs#L1364

and print the convolution of `evolved_slice`. A quick check is counting whether the number of convolutions is the same between the good and bad version. If they're not, the problem is probably a reincarnation of <https://github.com/NNPDF/pineappl/issues/304>.

Radonirinaunimi:
> Do you have any evidence that it's the parsing that's wrong?

Nope. And I do no longer believe that this is necessarily the issue.

> if the evolution itself is broken

If it is broken, then it is **only** broken for very old (not sure from which version exactly) EKOs. Just to summarize the status:

|  EKO version | PineAPPL version  | Status  |
|---|---|---|
| old (not sure from which version) | pre-`v1.0.0a2`  | ✅  |
| old  |  `v1.0.0a2`  |  ❌  |
| recent  |  `v1.0.0a2` |  ✅  (checked by `pineko` regression) |

> I'm traveling right now, so if you want to go ahead, one potential avenue

Yes, if I have some time today I will try to debug what's going on.

felixhekhorn:
You should be able to check the EKO version from `metadata.yaml` (inside the tar file). However, since it is working with pre-1.0-alpha the eko seems sensible. I guess that something in reading the ekos has changed in v1.0-alpha - this is again a hint to support https://github.com/NNPDF/eko/issues/97, i.e. specifically delegate the parsing to EKO itself rather then here

cschwan:
@Radonirinaunimi: using the files provided by you,
- https://data.nnpdf.science/pineko/data_files/data/grids/400/HERA_NC_225GEV_EP_SIGMARED.pineappl.lz4 and
- https://data.nnpdf.science/pineko/data_files/data/ekos/400/HERA_NC_225GEV_EP_SIGMARED.tar,

and `pineappl evolve ...` I (fortunately/unfortunately?) can't reproduce the wrong numbers, neither with `1.0.0-alpha2` nor with `master`. The file `operator.yaml` from the EKO above says `eko_version: 0.13.1`, which I assume is the 'old' EKO version?

regression_check.sh
```diff
@@ -11,7 +11,7 @@ POLARIZED_UNPOLPDF_NAME="NNPDF40_nlo_pch_as_01180"
 
 LIST_DIS_DATASETS=(
   "HERA_CC_318GEV_EP-SIGMARED"
-  "NNPDF_POS_2P24GEV_F2D"
+  # "NNPDF_POS_2P24GEV_F2D"
```
Radonirinaunimi, line 14: @cschwan, you're right. The results are the same in both versions using the grid and EKO I provided you before. Somehow, that EKO and the ones I have locally and the in the benchmarks were not the same. But this is now fixed.

Now, we can make a new release that includes the fixes to empty grids, add back the POS dataset in the regression, and this PR should be done.
```suggestion
  "NNPDF_POS_2P24GEV_F2D"
```

Radonirinaunimi: Addressed in [69268eb](https://github.com/NNPDF/pineko/pull/206/commits/69268ebe5309aae91afb5abc1c276be6393dc93d).

Radonirinaunimi:
Everything here is now done! The only thing missing here is the stable version of PineAPPL; as soon as that is available we can bump the version here and merge.

src/pineko/check.py
```diff
@@ -111,43 +111,38 @@ def check_grid_and_eko_compatible(pineappl_grid, operators, xif, max_as, max_al)
...
+    convolutions: pineappl.grid.convolutions
+           an object containing the information on the convolution
 
     Returns
     -------
     bool
         true if the fktable is a DIS fktable
     """
-    return True if len(channels[0][0][0]) == 1 else False
+    return True if len(convolutions) == 1 else False
```
felixhekhorn, line 127: you also need to check if it is space-like - else it might be also SIA ...

src/pineko/comparator.py
```diff
@@ -89,8 +89,10 @@ def compare(pine, fktable, max_as, max_al, pdfs, scales, threshold=5.0):
     df["FkTable"] = after
     df["permille_error"] = (after / before - 1.0) * 1000.0
 
-    if (df["permille_error"].abs() >= threshold).any():
-        print(df)
+    # Remove Q2 points that are below 1 GeV2
+    check_df = df[df["O2 left"] >= 1] if "O2 left" in df.columns else df
```
felixhekhorn, line 93: maybe take the "1" as input (from arguments and thus from CLI)?

src/pineko/check.py
Radonirinaunimi: But SIA always involves two convolution though.

Radonirinaunimi: Ah, you are thinking of these processes. Because we won't have measurements and predictions for these in the foreseeable future, I didn't think about them. But you're right, we should also account for these cases. I'll add such modifications.

scarlehoff (commented):
I'm currently using this branch and seems to work well :). The only problem I've found up to now is what I already reported when there are empty bins.

src/pineko/check.py
felixhekhorn: hihi, nono - SIA = semi-inclusive annihilation and this is the baseline for all current FF extractions. We have lot's of data: look e.g. to Table 1 of [2503.21311](https://arxiv.org/pdf/2503.21311) . Also Belle is still actively taking data I think

Radonirinaunimi: hahaha, of course. I was being very stupid. I still did have in mind SIDI while still writing SIA. My brain is not functioning well because of exhaustion 🙈

src/pineko/cli/compare.py
```diff
@@ -19,7 +19,8 @@
 @click.option(
     "--threshold", default=5.0, help="threshold in permille to accept Grid -> FK"
 )
-def subcommand(fktable_path, grid_path, max_as, max_al, pdfs, xir, xif, threshold):
+@click.option("--xia", default=1.0, help="fragmentation scale variation")
+def subcommand(fktable_path, grid_path, max_as, max_al, pdfs, xir, xif, threshold, xia):
```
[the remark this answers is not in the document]
Radonirinaunimi: Thanks a lot @scarlehoff for testing this in real-case scenario!! That helps quite a lot!

The issue with the zero bin should be fixed soon.

scarlehoff:
Hi @Radonirinaunimi one request before merging this (and I think it is better to do it here since it changes in 1.0), could we ensure that all metadata from the original grid is propagated to the fktable?

Radonirinaunimi:
> Hi @Radonirinaunimi one request before merging this (and I think it is better to do it here since it changes in 1.0), could we ensure that all metadata from the original grid is propagated to the fktable?

Yes! In principle this should be the case but I will explicitly double-check.

scarlehoff:
Great! 

(just to say, I didn't check it wasn't the case, I was looking at the changes and realized that in the scale_variations.py the propagation is explicit but I couldn't find in the evolve, but perhaps I just missed it, too many changes)

Radonirinaunimi:
> Hi @Radonirinaunimi one request before merging this (and I think it is better to do it here since it changes in 1.0), could we ensure that all metadata from the original grid is propagated to the fktable?

Sorry, I missed to reply to this earlier. The metadata are always automatically propagated into the FK tables. Scale varied FK grids/FK tables are created from new instances so that's why it is needed to manually copy the metadata.

src/pineko/cli/convolve.py
```diff
@@ -58,48 +57,55 @@ def subcommand(
...
     n_ekos = len(op_paths)
-    with eko.EKO.edit(op_paths[0]) as operators1:
+    with eko.EKO.edit(op_paths[0]) as first_operator:
+        operators = [first_operator]
+        path_operators = f"[+] {op_paths[0]}\n"
+        # If there are more than ONE operator, then account for all of them.
+        if len(operators) > 1:
+            for op_idx in range(1, n_ekos):
+                operators.append(eko.EKO.edit(op_paths[op_idx]))
+                path_operators += f"[+] {op_paths[1]}\n"
```
felixhekhorn, line 80: `{op_paths[1]}` -> `{op_paths[op_idx]}`

felixhekhorn (commented):
I didn't read the big `evolve.py` and `theory.py` (yet), but I found a stupid typo

PR #209, opened 2024-11-23
Use just the nnpdf package

scarlehoff:
This is the moment that @felixhekhorn had been waiting for so long!

At long last, a package containing the data and the theories without any of the tensorflow / keras / vp bullshit! https://pypi.org/project/nnpdf-data/

It also doesn't install the `rawdata` of course (which will be removed from the main repository anyway). Currently I've uploaded the package by hand to pypi since until the commondata project is not finished I cannot realistically have this PR https://github.com/NNPDF/nnpdf/pull/2225 in master since it would create too many conflicts.

I was hoping the benchmarks would finish before boarding the plane but alas, I won't know whether this worked ootb until tomorrow. It was fine locally though.


scarlehoff:
Let me know if you have any specific objections to this, otherwise we could even merge and release.

pyproject.toml
```diff
@@ -35,8 +35,7 @@ pandas = "^2.1"
 rich = "^12.5.1"
 click = "^8.0.4"
 tomli = "^2.0.1"
-nnpdf = { git = "https://github.com/NNPDF/nnpdf", optional = true}
-lhapdf-management = { version = "^0.5", optional = true }
```
RoyStegeman, line 39: Why remove the `lhapdf-management` dependency? Doesn't pineko itself also use it

src/pineko/utils.py
```diff
@@ -29,14 +29,17 @@ def read_grids_from_nnpdf(dataset_name, configs=None):
...
 
     # Import NNPDF only if we really want it!
-    from nnpdf_data import legacy_to_new_map
-    from validphys.commondataparser import EXT
-    from validphys.loader import Loader
+    from nnpdf_data import legacy_to_new_map, path_commondata
+    from nnpdf_data.commondataparser import EXT, parse_new_metadata
 
     # We only need the metadata, so this should be enough
+    # pass it through the legacy_to_new in case this is an old name
```
RoyStegeman, line 36:
```suggestion
    # We only need the metadata, so this should be enough.
    # Pass it through the legacy_to_new in case this is an old name.
```

pyproject.toml
scarlehoff: No, it was only nnpdf and only through imports.

RoyStegeman: Ah so whenever pineko uses lhapdf, we need genuine lhapdf and can't use `lhapdf-management`?

scarlehoff: Yes.But I think it is only used if you choose to compare fktable and grid at the end (which maybe we should do by default but that's another topic)

RoyStegeman: okay anyway replacing lhapdf with lhapdf-management is not the point of this PR

giacomomagni (approved):

src/pineko/utils.py
```diff
@@ -29,14 +29,17 @@ def read_grids_from_nnpdf(dataset_name, configs=None):
...
+    # We only need the metadata, so this should be enough.
+    # Pass it through the legacy_to_new in case this is an old name.
     dataset_name, variant = legacy_to_new_map(dataset_name)
-    cd = Loader().check_commondata(dataset_name, variant=variant)
-    fks = cd.metadata.theory.FK_tables
+
+    setname, observable = dataset_name.rsplit("_", 1)
+    metadata_file = path_commondata / setname / "metadata.yaml"
+    metadata = parse_new_metadata(metadata_file, observable, variant=variant)
+    fks = metadata.theory.FK_tables
```
felixhekhorn, line 42: it's fine to have this stuff here for the moment, but shouldn't that all be moved eventually up into `nnpdf-data`? all that information is very specific to that package

felixhekhorn: actually, the last line (`fks = ...`) will be left here

RoyStegeman: You're probably right, but putting it in nnpdf-data would result in some double counting with the regular nnpdf package which is also not ideal

scarlehoff: To be looked at after this is actually used.

Issue #194, opened 2024-09-12
Full regression test


scarlehoff:
`pineko` is a software central to any PDF fit, as it is used to create the fktables by evolving grids down to the fitting scale.

As such, is very important that:

1) Any changes keep results unchanged or, if there are changes, they are understood.
2) Updates in `nnpdf`, `eko` and `pineappl` are kept track of (this is important since `pineko` is mostly feature complete, and we should try to make sure it doesn't lag to far behind).

The best way to accomplish both points is with regression tests that can be run with PRs or whenever there are changes to the dependencies like in https://github.com/NNPDF/pineko/pull/190

The tests should then:

- [ ] Use the latest nnpdf release or repository (since we are still not putting the releases in pip... soon). This will ensure that `pineko` is able to read the data and theory from the NNPDF data files (crucial importance for the fit!!)
- [ ] Create the operator cards: 
- [ ] Create the ekos (this ensures that a) `pineko` is still compatible with eko b) the theory information from nnpdf is still compatible with pineko -> eko)
- [ ] Create the fktables by convoluting the ekos with the grids (ensures that the interface `pineappl`-`eko` used by `pineko` is still working as expected)
- [ ] Check (by convoluting with the right PDF) that the fktable and the grid are compatible
- [ ] Check (by comparing to previous results) that the results haven't changed to a certain precision

This should be done for a bunch of DIS grids, and a bunch of double-hadronic grids.
And it should be done for a bunch of theories, it should include `41_000_000` and, hopefully, also a polarized theory.
Take into account the two limitations of github workflows in terms of runtime and memory (so we cannot do it with the full dataset).

Some extra (bonus) points:
- In order to start, add a few grids to the repository in your branch so you don't have to deal with authentication to download them or anything for the time being. We can work on that after the regression tests are ready.
- You probably want to not run this in every commit. Maybe only for PR that are "ready for review" or using a specific label for it. You can leave this point for the end.
- Github allows to have "self hosted runners". We might want to explore that option in order to do a more comprehensive test.

You can look at these workflows for inspiration:

https://github.com/NNPDF/pineko/blob/main/.github/workflows/bench.yml (note that you don't need pylint, coverage or anything like that, it should focus on the regression and the regression only)
https://github.com/NNPDF/nnpdf/blob/master/.github/workflows/regression_tests.yml

This will teach you:
1) How to generate fktables starting from scratch
2) How to use all the moving parts that create a theory.

PR #215, opened 2025-02-21
Towards a full regression test

Radonirinaunimi:
In order to quickly wrap up https://github.com/NNPDF/pineko/pull/206, I realize I better as well address this now. This will not only check that https://github.com/NNPDF/pineko/pull/206 is properly working but will also address https://github.com/NNPDF/pineko/issues/194.

Here are the various checks that need to be done:
- [x] clarify which theories/grids should be used and check that the **template operator cards** are the correct ones (currently using `40008005`)
- [x] `pineko` is able to generate EKOs
- [x] `pineko` is able to produce FK tables by convolving the grids with the EKOs
- [x] check that the produced FK tables are compatible with the prior grids by convolving them with PDFs (differences are checked to be below **2 permille**, `ATLAS_SINGLETOP_8TEV_T-RAP-NORM` is slightly above 1 permille)
- [x] check that the resulting FK tables are consistent with some references by convolving them with PDFs (differences are checked to be below **1 permille**)
- [x] include varieties of NNPDF datasets - this should be representative of the processes included in the fit
  - [x] `HERA_CC_318GEV_EP-SIGMARED`
  - [x] `ATLAS_Z0_7TEV_36PB_ETA` 
  - [x] `LHCB_WPWM_8TEV_MUON_Y`
  - [x] `ATLAS_SINGLETOP_8TEV_T-RAP-NORM`
  - [x] `NNPDF_POS_2P24GEV_F2D`
- [x] check cases in which multiple EKOs, polarized⊗space-like⊗time-like, are needed for a single grid
  - [x] `STAR_WMWP_510GEV_WP-AL` 

I will at some point write some documentations (while also updating the deprecated information) but for the time I am putting the steps here:
- The information regarding the experimental data and the theory cards are fetched from `nnpdf_data` which is installed with `pineko`.
- The data (such as grids and template operator cards) are stored in https://data.nnpdf.science/pineko/theory_productions/. `theory_productions` also contains the correct folder structure to produce the theory predictions as described in `pineko.ci.toml`.
- `pineko.ci.toml` contains the `pineko` configuration to run the regression test
- The `theory_productions` data is downloaded once for a given version and then cached in order to not download it every time the workflow is running. If new data are introduced in the `theory_productions` server, the version at the following line needs to be incremented:
https://github.com/NNPDF/pineko/blob/b357030535cf8372a74ca0c154bd2ac92c684da0/.github/workflows/regression.yml#L44
- Whenever new data are added to the `theory_productions` server, they should be put in the correct folders according to the configurations defined in `pineko.ci.toml`.

> [!NOTE] 
> Currently, this is using the a runner `pineko-stbc3` hosted in the Nikhef stoomboot cluster under my user-space. The status of the runner can be tracked in `settings > actions > runners`.


Radonirinaunimi:
So it looks like we need a self-hosted runner after all - the github one cannot even deal with a simple grid that only has 3 $Q^2$ points. I will set up one on Stoomboot first and check that it runs properly and we can decide afterwards where to host it permanently.

.github/workflows/regression.yml
```diff
@@ -0,0 +1,60 @@
+name: regression
+
+# start job only for PRs when a label is added.
+on: push
+  # pull_request:
+  #   types: [labeled]
+
+jobs:
+  regresstion:
+    # if: contains(github.event.pull_request.labels.*.name, 'run-regression')
```
scarlehoff, line 10: Maybe given the fact that uses a runner under your account it would be better to run it only with the labels to avoid hammering the runner on every commit?

pineko.cli.toml
```diff
@@ -0,0 +1,15 @@
+[general]
+nnpdf=true
+
+[paths]
+# inputs
+grids = "./theory_productions/data/grids"
+operator_card_template_name = "_template.cli.toml"
```
scarlehoff, line 7: I thought the templaets were .yaml file? Has this changed in eko?

scarlehoff (commented):
Thank you so much for this. I guess in Amsterdam I'll ask you how to set up the runner in the cluster so we can have more than one.

Let me know if it is finished and I'll try running it myself and review it asap.

(thanks again ¡!)

pineko.cli.toml
felixhekhorn: they are yaml files

Radonirinaunimi: Fixed in [5e89cb9](https://github.com/NNPDF/pineko/pull/215/commits/5e89cb9964aefbae2cef1a9457a09c60b32bb99d).

.github/workflows/regression.yml
Radonirinaunimi: Reverted back in [04e91d6](https://github.com/NNPDF/pineko/pull/215/commits/04e91d6a2abc0577ab8c5c6310b417f0e3a2676f), now given that there are more datasets.

Radonirinaunimi:
With a hadronic dataset that contains ~30 $Q^2$ points, it takes about 20 mn to compute everything. I believe we can just add one more DIS dataset with **much** less $Q^2$ points (still looking for the one).

> I guess in Amsterdam I'll ask you how to set up the runner in the cluster so we can have more than one.

Yes! If there are more computers/clusters that we can use the better.

PR #227, opened 2025-06-22
Remove rotations

Radonirinaunimi:
Remove the rotations at least for EKOs v0.15 which are always expressed in the flavour basis.


felixhekhorn:
As [argued on the otherside](https://github.com/NNPDF/pineko/pull/205#discussion_r2256939289) I would revert 0bb666a - we want as many `optimize` as possible

PR #232, opened 2025-09-24
Change check in fonll

evagroenendijk:
I changed this check in fonll because for me the XIA is not necessarily set in all theory cards but it is in the fonll cards. I don't know if this is the best solution, so please let me know!

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
 ci:
-  autofix_prs: false
+  autofix_prs: true
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v5.0.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
 ci:
-  autofix_prs: true
+  autofix_prs: false
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v5.0.0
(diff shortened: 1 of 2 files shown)


scarlehoff (approved):
I personally think we should be checking a subset of keys instead of removing keys and checking the rest?

My two cents, instead of the current code at worst I think we should have

```python
        dont_check = ["FNS", "PTO", "PTODIS", "NfFF", "ID", "FONLLParts", "Comments" "XIA"]

        differ = []
        for key in theorycards[0].keys():
            if key in dont_check:
                continue
            if len({card.get(key)) != 1:
                differ.append(key)

        if differ:
            raise ValueError(f"The following keys differ between different FONLL theory cards: {differ}")
```

That way we also get the differences instead of the current "though luck"...

But I'm ok with merging the current change and calling it a day.

Pinging also @felixhekhorn as he might have a strong opinion here.

felixhekhorn (changes requested):
I agree with the change suggested by @scarlehoff . Just fixing some typo and return value
```py
        dont_check = ["FNS", "PTO", "PTODIS", "NfFF", "ID", "FONLLParts", "Comments" "XIA"]

        differ = []
        for key in theorycards[0].keys():
            if key in dont_check:
                continue
            if len({card.get(key) for card in theorycards}) != 1:
                differ.append(key)
            del theorycards[0][key]

        if len(differ) > 0:
            raise ValueError(f"The following keys differ between different FONLL theory cards: {differ}")
```
(still coding in the void so care should be taken

felixhekhorn (approved):
I guess you tested by hand? Adding a unit test would be really great (as this is obviously a non-trivial business and even caused trouble in real life), else looks okay

felixhekhorn (changes requested):
please fix pre-commit

scarlehoff (approved):

felixhekhorn (approved):

PR #236, opened 2025-10-28
Disallow eko inheritance when the operator cards would not be compatible

scarlehoff:
When doing nfonll the inheritance of ekos doesn't check anything, it just assumes that the inherited grid will have exactly the same values of Qs that the parent.

This has triggered several time the "different values of Q" error. Now, most likely this is happening (the error) because all the grids involved should be 0 but it is still a bug nonetheless.


felixhekhorn (approved):

felixhekhorn:
> This doesn't allow the forcing though.

okay - better now? I just want to have something more concrete (and findable :innocent: )

PR #245, opened 2026-04-10
fail if there is a mismatch in the number of iterations

scarlehoff:


scarlehoff:
Funny, the bechmarks are using inconsistent iterations :)

felixhekhorn:
> Funny, the bechmarks are using inconsistent iterations :)

it just proves the point :see_no_evil: can you please fix the files on the server?

PS: @Radonirinaunimi #215 has an explicit item in the todo list :see_no_evil:

Radonirinaunimi:
> > Funny, the bechmarks are using inconsistent iterations :)
> 
> it just proves the point 🙈 can you please fix the files on the server?
> 
> PS: @Radonirinaunimi #215 has an explicit item in the todo list 🙈

Indeed - which is supposed to be ultimate benchmark/regression tests and is operational.

scarlehoff:
It's a bit more complicated than that, it is also using old datasets. The yamldb. Truncated. It was never updated to 4.1.

Let me fix the benchmark so it passes and later we redo the regression test.

PR #246, opened 2026-04-20
Completely remove `grid.optimize()`

Radonirinaunimi:
This reflects https://github.com/NNPDF/yadism/pull/365. Here, however, the changes have more significant impacts because they also affect the hadronic grids (if not pre-optimized by the MC generators) which are the main bottleneck when it comes to EKO computations.

### DIFF
--- a/src/pineko/cli/check.py
+++ b/src/pineko/cli/check.py
@@ -38,7 +38,6 @@ def sub_compatibility(grid_path, operator_path, xif, max_as, max_al):
 
     """
     pineappl_grid = pineappl.grid.Grid.read(grid_path)
-    pineappl_grid.optimize()
     with eko.EKO.read(pathlib.Path(operator_path)) as operators:
         for (q2, _), _ in operators.items():
             try:
@@ -88,7 +87,6 @@ SCVAR_MESSAGES = {
 def sub_scvar(grid_path, scale, max_as_order, max_al_order):
     """Check if PineAPPL grid contains requested scale variations for the requested order."""
     grid = pineappl.grid.Grid.read(grid_path)
-    grid.optimize()
 
     # Call the function
     scaleobj = check.Scale[scale]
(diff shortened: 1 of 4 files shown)


felixhekhorn:
1. you are the PineAPPL developer so you know best :upside_down_face:  then I still think this sounds like a bug (somewhere). If this is the quick and dirty way forward, fine let's pay the prize here
2. however, at the very minimum open an issue with title "Fix/Revert #246" (or similar). I don't think this is the _correct_ way and we should remember that (not sure we will ever change it, but we might)
3. Now, I also think (same as @scarlehoff ) that instead also the SV logic should be changed. Having an 0 coefficient function should be acceptable, I think - there is no math reason why it can't be 0 (and often it will be at a specific kinematic point). This is to say: the conjectured bug might well be in the SV logic of pineko

scarlehoff:
I agree that this is for me a bug in pineappl _and_ a bug in pineko.

If I decide to optimize all grids to save some space in the server (which I've done in the past for fktables...) suddenly the grids would all be broken. I won't do it just because I happen to be aware of this problem but I think it is not a very good situation to be in.

Radonirinaunimi:
No no, there is no bug at all in PineAPPL. PineAPPL is doing **exactly** what is supposed to do here. Not at all related to this issue, PineAPPL is just missing an interface to a function that allows one to select subsets of the optimization, which just happens can be used to address the issue.

The bug instead is in `pineko`. As I mentioned in Slack, I also agree that not having central orders is perfectly valid and `pineko` should be modified to support this. Once this is implemented, all the problems related to optimizations will be gone and one can fully optimize. It is just that doing so carefully require some time and carefully tests `pineko` and this PR (and the one in Yadism) was just meant as a quick way to not slow down things.

Radonirinaunimi:
Aborted in favor of https://github.com/NNPDF/pineko/pull/248.

Issue #225, opened 2025-05-26
Save the hash and theory directory of the grid


scarlehoff:
One common problem that we have is that from one grid folder we can generate many different theories depending on the evolution.*
Was it theory 4001? 4100000? 4005? Was it the current version of the grids or three bugs ago?

I think we should save a key when creating the fktable which saves 1) the hash of the grid 2) the theory number, something along the lines of `grid_files: 4001-<name>-<hash>` (plural grid files for DIS fktables). This key is not supposed to be used by anyone ever, hopefully with the information in the theory card we should be able to reproduce the grid (where it was coming from, which settings, etc) but if everything fails then with this information one can go to the server and browse the long list of hashes until the right grid is found.

felixhekhorn:
what does the "theory directory" in the title refer to? hash is good on it's own, no?

scarlehoff:
Theory directory in order to know which folder was used for the grids.
For instance, for theory 700/708/40_000_000 and many of the variants the grids of theory 4001 are used.

Hopefully this distinction will happen less in the future but given that is just a few characters (and we have already seen the problem) I think it is worth it to include it.

felixhekhorn:
I see - you mean the *true* underlying grid? i.e. in practice something like `Path.resolve`?

scarlehoff:
Yes! As I said, it is meant as a last resort, so we want to have as much raw information as possible.

kamillaurent:
Now the hash and path informations of the grid used to generate the FK Table are stored in the metadata of the FK Table. See PR #247.

PR #247, opened 2026-04-28
Save hash and path grid informations when generating FK Tables

kamillaurent:
When a FK Table is generated, the hash information and path of the `pineappl` grid used for it are saved in the metadata of the FK Table.

These information can be read using the `pinealppl-cli`, for example:

`pineappl read --show ATLAS_WZ_TOT_13TEV-ATLASWZTOT13TEV81PB_Z_tot.pineappl.lz4`

outputs, between the other informations, this lines:

```
grid_files: {"ATLAS_WZ_TOT_13TEV-ATLASWZTOT13TEV81PB_Z_tot.pineappl": {"hash": "f1110fd0bb8ccddf28bf722f9f9c3dd5", "path": "/data/theorie/klaurent/FKTables/pineko/data/grids/41000000/ATLAS_WZ_TOT_13TEV-ATLASWZTOT13TEV81PB_Z_tot.pineappl.lz4"}}
```
This PR is related to issue #225

### DIFF
--- a/src/pineko/evolve.py
+++ b/src/pineko/evolve.py
@@ -1,6 +1,7 @@
 """Tools related to evolution/eko."""
 
 import copy
+import hashlib
 import json
 import logging
 import os
@@ -303,6 +304,7 @@ def evolve_grid(
     assumptions="Nf6Ind",
     comparison_pdfs: Optional[list[str]] = None,
     min_as=None,
+    grid_path: Optional[os.PathLike] = None,
 ):
     """Convolute grid with EKO from file paths.
 
@@ -332,6 +334,8 @@ def evolve_grid(
         if given, a comparison table (with / without evolution) will be printed
     min_as: None or int
         minimum power of strong coupling
+    grid_path : str or os.PathLike or None
+        path to the grid file, used to store grid hash metadata
     """
     order_mask = pineappl.boc.Order.create_mask(grid.orders(), max_as, max_al, True)
     if min_as is not None and min_as > 1:
@@ -432,6 +436,13 @@ def evolve_grid(
 
     fktable.set_metadata("pineko_version", version.__version__)
     fktable.set_metadata("theory_card", json.dumps(theory_meta))
+
+    # Store grid hash and path information
+    if grid_path is not None:
+        grid_path_obj = pathlib.Path(grid_path)
+        grid_hash = hashlib.md5(grid_path_obj.read_bytes()).hexdigest()
+        grid_files = {grid_path_obj.stem: {"hash": grid_hash, "path": str(grid_path_obj.resolve())}}
+        fktable.set_metadata("grid_files", json.dumps(grid_files))
 
     # compare before/after
     comparison = None
--- a/src/pineko/evolve.py
+++ b/src/pineko/evolve.py
@@ -68,6 +68,20 @@ def construct_atlas(tcard):
     return atlas
 
 
+def set_fktable_metadata(fktable, theory_meta, grid_path=None):
+    """Store the common FK metadata and, if available, grid provenance."""
+    fktable.set_metadata("pineko_version", version.__version__)
+    fktable.set_metadata("theory_card", json.dumps(theory_meta))
+    if grid_path is None:
+        return
+
+    grid_path_obj = pathlib.Path(grid_path).resolve()
+    grid_hash = hashlib.md5(grid_path_obj.read_bytes()).hexdigest()
+    fktable.set_metadata("grid_hash", grid_hash)
+    fktable.set_metadata("grid_theory", grid_path_obj.parent.name)
+    fktable.set_metadata("grid_name", grid_path_obj.name)
+
+
 def construct_empty_fktable(grid, theory_meta, fktable_path, grid_path=None):
     """Construct and write a structurally valid but numerically empty FK table.
 
@@ -112,23 +126,7 @@ def construct_empty_fktable(grid, theory_meta, fktable_path, grid_path=None):
         scale_funcs=grid.scales,
     )
     fktable = pineappl.fk_table.FkTable(empty_grid)
-    fktable.set_metadata("pineko_version", version.__version__)
-    fktable.set_metadata("theory_card", json.dumps(theory_meta))
-    if grid_path is not None:
-        grid_path_obj = pathlib.Path(grid_path).resolve()
-        grid_hash = hashlib.md5(grid_path_obj.read_bytes()).hexdigest()
-        grid_path_parts = grid_path_obj.parts
-        if "pineko" in grid_path_parts:
-            pineko_idx = grid_path_parts.index("pineko")
-            display_path = str(pathlib.Path("/", *grid_path_parts[pineko_idx + 1 :]))
-        elif "data" in grid_path_parts:
-            data_idx = grid_path_parts.index("data")
-            display_path = str(pathlib.Path("/", *grid_path_parts[data_idx:]))
-        else:
-            display_path = grid_path_obj.name
-        fktable.set_metadata("grid_hash", grid_hash)
-        fktable.set_metadata("grid_theory", grid_path_obj.parent.name)
-        fktable.set_metadata("grid_path", display_path)
+    set_fktable_metadata(fktable, theory_meta, grid_path=grid_path)
     fktable.write_lz4(str(fktable_path))
     return fktable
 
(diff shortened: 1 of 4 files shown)


src/pineko/evolve.py
```diff
@@ -433,6 +437,13 @@ def prepare(operator, convolution_types):
     fktable.set_metadata("pineko_version", version.__version__)
     fktable.set_metadata("theory_card", json.dumps(theory_meta))
 
+    # Store grid hash and path information
+    if grid_path is not None:
+        grid_path_obj = pathlib.Path(grid_path)
+        grid_hash = hashlib.md5(grid_path_obj.read_bytes()).hexdigest()
+        grid_files = {grid_path_obj.stem: {"hash": grid_hash, "path": str(grid_path_obj.resolve())}}
```
scarlehoff:
```suggestion
        grid_files = {grid_path_obj.stem: {"hash": grid_hash, "theory_folder": grid_path_obj.parent.name}}
```

Out of privacy concerns, I would just store the path starting from the theory_slim folder.

My suggestion is, since you have the grid name as the key, save just the theory folder name.

scarlehoff (approved):
LGTM.
(haven't tested it, but I trust the OP is a real-life test)

scarlehoff:
Hi @kamillaurent please do the small change requested and fix the issues found by pre-commit (usually just means you didn't run pre-commit, if you do the changes then git add the `evolve.py`  file and run `pre-commit` it will be fixed.

src/pineko/evolve.py
felixhekhorn: agreed on the privacy, but I think I would keep the `resolve()`, because I think this way we can resolve symbolic links, right? i.e. `.resolve().parent` (or similar)

felixhekhorn (changes requested):
Remember the other things we discussed:
- split in two simple keys "grid_path" and "grid_hash"
- address FONLL magic

scarlehoff:
@kamillaurent please address the changes and run pre-commit so that we can merge this

scarlehoff:
There's a few conflicts due to having merge @Radonirinaunimi's PR but they are in the documentation of the functions so it should be quick to fix. It would be good to have and use this PR for the next batch of fk tables so we have this information in 😅

kamillaurent:
Now I made the informations in the metadata more readable. For example, using the command
`pineappl read --show ATLAS_WZ_TOT_13TEV-ATLASWZTOT13TEV81PB_Z_tot.pineappl.lz4` one can find the informations:

```
...
grid_hash: f1110fd0bb8ccddf28bf722f9f9c3dd5
grid_path: /data/grids/41000000/ATLAS_WZ_TOT_13TEV-ATLASWZTOT13TEV81PB_Z_tot.pineappl.lz4
grid_theory: 41000000
...
```
 
When we use the FONLL procedure, in the last step many FK-tables get merged into one. I am working on retaining the info about hash and theory of the original FKs into the merged one.

src/pineko/evolve.py
```diff
@@ -68,6 +68,71 @@ def construct_atlas(tcard):
...
+            pineko_idx = grid_path_parts.index("pineko")
+            display_path = str(pathlib.Path("/", *grid_path_parts[pineko_idx + 1 :]))
+        elif "data" in grid_path_parts:
+            data_idx = grid_path_parts.index("data")
+            display_path = str(pathlib.Path("/", *grid_path_parts[data_idx:]))
+        else:
+            display_path = grid_path_obj.name
+        fktable.set_metadata("grid_hash", grid_hash)
+        fktable.set_metadata("grid_theory", grid_path_obj.parent.name)
+        fktable.set_metadata("grid_path", display_path)
```
felixhekhorn: 1) This is far too much code duplication - this needs to be encapsulated
2) do we really need `grid_path`? this looks fairly complicated ...

scarlehoff: Yes, the grids path part is a bit more complicated than needed, `data/grids` are not important.

Once you have the grid path, the "path of the grid itself" is `grid_path.name` and the theory folder `grid_path.parent.name`. You don't need anything else.

kamillaurent: I tried to simplify the metadata, using a function and calling it when the FK table is constructed (so there is no code duplication):

```
def set_fktable_metadata(fktable, theory_meta, grid_path=None):
    """Store the common FK metadata and, if available, grid provenance."""
    fktable.set_metadata("pineko_version", version.__version__)
    fktable.set_metadata("theory_card", json.dumps(theory_meta))
    if grid_path is None:
        return

    grid_path_obj = pathlib.Path(grid_path).resolve()
    grid_hash = hashlib.md5(grid_path_obj.read_bytes()).hexdigest()
    fktable.set_metadata("grid_hash", grid_hash)
    fktable.set_metadata("grid_theory", grid_path_obj.parent.name)
    fktable.set_metadata("grid_name", grid_path_obj.name)
```
For consistency I moved into the function also `pineko_version` and `theory_card`. I removed the path since the relevant information is the name of the folder, which is also the theory ID.

kamillaurent:
@scarlehoff regarding the FONLL procedure, `fonll.py` merges all the FK-Tables into the `$THEORY_ID_00` and retains informations only from this one. For example, `pineappl read --show  41000000/HERA_NC_225GEV_EP_SIGMARED.pineappl.lz4` gives
```
...
eko_version: 0.15.3
fk_assumptions: Nf4Ind
fktable_id: HERACOMBNCEP460
grid_hash: 0a34218e5e6b254475243264d1ced40e
grid_name: HERA_NC_225GEV_EP_SIGMARED.pineappl.lz4
grid_theory: 4100000000
hepdata: 10.17182/hepdata.68951.v1/t4
nnpdf_id: HERACOMBNCEP460
...
```

Is this sufficient or we want informations about all the merged grids? In the case we want all the merged grids, the `grid_name` will be the same for all. The `grid_hash` and `grid_theory` will be different, should we store them into an array and store the array in the metadata of the merged FK Table?

scarlehoff:
I would say that grid theory should strip the extra 0s (so it should say just `41000000`, since that covers all).
Grid name is fine.

And my preference for the hash would be simply a string of

```
00-0a34218e5e6b254475243264d1ced40e 01-hash 02-hash 03-...
```

felixhekhorn:
Are you sure multiple plain keys wouldn't be better? Like `grid_name_00`?

In that case we should also mark the standard field `grid_name: /FONLL` or something that can never be a hash.

I'm thinking about what would be easiest to parse afterwards...

scarlehoff:
In my view, the important part about the hash is that it is unique in the comparison so making sure that all fktables have the key `grid_hash` and that the string is unique is sufficient and preferable.
The fact that one of the hashes is made by multiple hashes is irrelevant.

And, for the cases in which the hashes are not identical, then one still has the information about the partial grids but most of the time we want a yes/no question to the "are these coming from the same grids" question. It is enough for one to be different.

kamillaurent:
Done! Now for a merged FK-Table the metadata contains the main theory and the hash of all the original (merged) FKs. For example:
`pineappl read --show 41000000/HERA_NC_225GEV_EP_SIGMARED.pineappl.lz4` gives
```
...
fktable_id: HERACOMBNCEP460
grid_hash: 00-0a34218e5e6b254475243264d1ced40e 01-efd1748bb08c25c43a22971cfe8699a9 02-6d0e6e92977898c4f052c05b4880eb51 03-bf95481ca46d9859e57476a0c3ded8db 04-fe3403683c505c439ef9e0dbfbdb74a7 05-df3dc24b51036ff6f2a97d93c38f445d 06-a70df1c6de56f8f5acb0229f18faa250
grid_name: HERA_NC_225GEV_EP_SIGMARED.pineappl.lz4
grid_theory: 41000000
...
```

scarlehoff:
@kamillaurent make sure to run pre-commit on the changed files so that the CI doesn't complain!

kamillaurent:
@scarlehoff the requested changed are now implemented, let me know if there is anything else, otherwise I think we can merge.

scarlehoff:
@felixhekhorn if you confirm you are happy with the current state we can merge (there is still a few tests to do before the release so there's time to walk back some change if we find any issues)

felixhekhorn (approved):

felixhekhorn:
@kamillaurent please remember to always resolve conversation, when you believe you are done - this way it is easier to review

PR #248, opened 2026-04-29
Fix no central orders and completely empty grids in scale variations

Radonirinaunimi:
This PR fixes two issues arising from specific grids during scale variations, namely:
- grids with no central orders but the corresponding scale varied orders are available
- completely empty grids that do not contain orders

PS: It also bumps the version of `pineappl` to `v1.4.0a1` for the tests.


Radonirinaunimi:
@kamillaurent Could you please check with this branch (say for one of the datasets and sub-theories that failed before) that it is no longer an issue?

kamillaurent:
@Radonirinaunimi I checked and seems the problem
```
ValueError: Central order is not available but sv order is.
``` 
is no longer there.

However, for these datasets:

```
HERA_NC_318GEV_EAVG_BOTTOM-SIGMARED
HERA_NC_318GEV_EAVG_CHARM-SIGMARED
BCDMS_NC_NOTFIXED_P_EM-F2
BCDMS_NC_NOTFIXED_D_EM-F2
NUTEV_CC_NOTFIXED_FE_NB-SIGMARED
NUTEV_CC_NOTFIXED_FE_NU-SIGMARED
CHORUS_CC_NOTFIXED_PB_NU-SIGMARED
```

I still encounter the second error:
```
ValueError: min() iterable argument is empty
```

Radonirinaunimi:
Could you please try now @kamillaurent ? (Fingers crossed 🤞🏼 it won't break the nFONLL combinations)

kamillaurent:
@Radonirinaunimi the fix worked and I could compute the remaining ekos and FK Tables.

src/pineko/evolve.py
```diff
@@ -333,6 +333,31 @@ def evolve_grid(
...
+            convolutions=grid.convolutions,
+            interpolations=grid.interpolations,
+            kinematics=grid.kinematics,
+            scale_funcs=grid.scales,
+        )
+        fktable = pineappl.fk_table.FkTable(empty_grid)
+        fktable.set_metadata("pineko_version", version.__version__)
+        fktable.set_metadata("theory_card", json.dumps(theory_meta))
+        fktable.write_lz4(str(fktable_path))
+        return fktable
```
scarlehoff: Why is this an internal function? I'd say this should be useful enough to be outside and called taking a grid as an input. Something like `construct_empty_fktable_like(grid, **otherinfo)`, similar to `zeros_like` from numpy.

src/pineko/evolve.py
```diff
@@ -349,6 +374,14 @@ def evolve_grid(
 
     muf2_grid = evol_info.fac1
     mur2_grid = evol_info.ren1
+
+    # If the grid is effectively empty (no kinematic support), PineAPPL/eko will
+    # panic later when constructing the target interpolation grid. This situation
```
scarlehoff: Why does it panic?

pyproject.toml
```diff
@@ -28,7 +28,7 @@ packages = [{ include = "pineko", from = "src" }]
 [tool.poetry.dependencies]
 python = ">=3.10,<3.14"
 eko = "^0.15.3"
-pineappl = "^1.3.0"
+pineappl = "^1.4.0a1"
```
felixhekhorn: I guess we can move to 1.4.0 now?

scarlehoff: Doesn't the a mean alpha?

felixhekhorn: yes, the true thing was not available at the time of writing I think (but now it is)

Radonirinaunimi: Yes, we can now move to `1.4.0`. I just released it last week.

Radonirinaunimi: Done in [5445386](https://github.com/NNPDF/pineko/pull/248/commits/54453864be074be0ea4a7173f098c178d0ef59af).

src/pineko/evolve.py
Radonirinaunimi: Added a clarifying comment about why and where exactly the panic happens in [50e9b68](https://github.com/NNPDF/pineko/pull/248/commits/50e9b687954eb02cc48f9cc744ea7fe8763c6fee).

src/pineko/evolve.py
Radonirinaunimi: Simply promoted as a global function in [50e9b68](https://github.com/NNPDF/pineko/pull/248/commits/50e9b687954eb02cc48f9cc744ea7fe8763c6fee).

Radonirinaunimi:
AFAIC, this is ready to go.

src/pineko/check.py
```diff
@@ -174,7 +176,14 @@ def contains_sv(
 ) -> Tuple[AvailableAtMax, int]:
     """Check whether renormalization scale-variations are available in the pineappl grid."""
```
scarlehoff: Could you explain in the docstr of this function what are we exactly returning and why it makes sense to do `return AvailableAtMax.BOTH, 0` if the grid doesn't contain anything?

src/pineko/evolve.py
```diff
@@ -333,6 +363,12 @@ def evolve_grid(
     min_as: None or int
         minimum power of strong coupling
     """
+    # A grid with no orders is a valid input for some workflows. In that case we
```
scarlehoff, line 399: Just as before, could you add here the what are we returning and why does it make sense to return the grid, and empty fktable and None

scarlehoff (commented):
Thank you for this. I mostly get the gist of it _now_ that I have the issue in mind but I think I will have no idea why we are returning empty/none things in a year 😅

src/pineko/check.py
Radonirinaunimi: Please now have a look at [b606ad1](https://github.com/NNPDF/pineko/pull/248/commits/b606ad1553a3fdcb4c7c1cb075adb4467f433e2c). I tried to be as detailed as possible.

src/pineko/evolve.py
Radonirinaunimi: See [b606ad1](https://github.com/NNPDF/pineko/pull/248/commits/b606ad1553a3fdcb4c7c1cb075adb4467f433e2c).

scarlehoff (approved):
Thank you very much for the detailed explanations.

Radonirinaunimi:
@felixhekhorn are you happy with this?

felixhekhorn (approved):

